PR #7, opened 2021-07-25
Add eslint

joshxg:
Hm.

I just did `npx lerna add xxx --dev` at the root for these things and then it worked.. so I tried the same with typescript and it was fine?

I have read a couple places that the packages should be 'self-contained' incase they don't want to use some packages in the root- but I think for ts/linting, is that a bit silly? Maybe not so much for ts 🤔


mark-prins (approved):
I wasn't sure about whether to add to the root package or keep each one self contained. I thought making them self contained wasn't necessary and led to a lot of duplicated installs - which doesn't affect size, just seems redundant.

I've gone with root package myself. and this all looks fine to me.

PR #11, opened 2021-07-26
#1 reduce bundle size

### LINKED ISSUE #1: Reduce bundle size
Appears to be loading a lot of material-ui; reduce the size of this.

mark-prins:
Fixes #1 

Maybe worth pulling it in now? Have just noticed that there is one or two useful lines in this.

### DIFF
--- a/host/webpack.config.js
+++ b/host/webpack.config.js
@@ -75,7 +75,6 @@ module.exports = {
         },
         {
           '@openmsupply-client/common': {
-            import: path.join(__dirname, '../common'), // '@openmsupply-client/common',
             requiredVersion: require('../common/package.json').version,
           },
         },
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ lerna-debug.log*
 
 # Compiled release
 **/build/*
+**/dist/*
 
 # Dependency directories
 node_modules/
@@ -18,3 +19,6 @@ typings/
 # TypeScript cache
 *.tsbuildinfo
 .DS_Store
+
+# Package stats
+**/stats.json
--- a/common/src/components/LoadingApp.tsx
+++ b/common/src/components/LoadingApp.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
-import { makeStyles, CircularProgress, Typography } from '@material-ui/core';
+import { CircularProgress, Typography } from './index';
+import { makeStyles } from '../styles';
 
 const useStyles = makeStyles(theme => ({
   root: {
--- a/common/src/components/index.js
+++ b/common/src/components/index.js
@@ -1,5 +1,0 @@
-export * from '@material-ui/core';
-export { LoadingApp } from './LoadingApp';
-export { DataGrid } from '@material-ui/data-grid';
-export { RemoteComponent } from './RemoteComponent';
-export * from './icons';
--- a/common/src/components/index.ts
+++ b/common/src/components/index.ts
@@ -1,0 +1,10 @@
+import Typography from '@material-ui/core/Typography';
+import ListItemIcon from '@material-ui/core/ListItemIcon';
+
+export { ListItemIcon, Typography };
+export { LoadingApp } from './LoadingApp';
+export { DataGrid } from '@material-ui/data-grid';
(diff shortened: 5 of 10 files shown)


joshxg:
Nice! Missing a couple imports/exports for me to get the app to run- the error messages were a little obscure so I ended up doing the required changes to figure out what was wrong- will commit to here and hope you don't mind!

Annnd.. I did some investigating also:

What I ended up doing was `yarn build` to see actual bundles created, minified, shaken etc.. in the network tab bundles there are a bunch of imports which are annotated with `webpack_unused` something a rather and those are only removed in a production build (apparently!). 

- I was getting two bundles in each `host` and `dashboard` with ~600kb and ~550kb which seemed way OTT
- After many fails, I removed sharing of `common` in the webpack config
- After this, the biggest bundle in `dist` is ~200kb, with one other at ~100kb, the rest are <10kb

I *believe* when we share a package, the entire package is bundled and non-tree shakable.. both `host` and `dashboard` when sharing the `common` package, get the 500kb + 600kb bundles in dist after `yarn build`. If I remove the sharing of `common` from `dashboard`, there is no more 500 or 600kb bundles in `dashboard/dist` 🤔 

Some links with people talking about it:
- https://github.com/webpack/webpack/discussions/13382
- https://stackoverflow.com/questions/66254107/tree-shaking-of-shared-dependencies-in-webpack-5-module-federation
- https://github.com/webpack/webpack/discussions/12532

So.. I suppose federation reduces the *overall* bundle sizes of everything downloaded for each of the apps, but *can* make individual apps bundle sizes larger..? Need to keep a balance in mind, I suppose. It's possible that each app importing their own components from material might feel better than a single import 🤷 

UPDATE: I just removed the `import` line, and now i'm only getting a single ~550kb bundle not the ~600kb one 🤔 
UPDATE: Removing the exporting of `icons` in `common` makes the bundles crazily smaller - I end up with a 250kb and 120kb bundles (+ all the <10kb ones)... the whole [icons](https://bundlephobia.com/package/@material-ui/icons@4.11.2) package is apparently only 300kb.. this is very strange 🤔 -- Probably a good idea to keep the icons out of the shared package, otherwise we'd have to download all of them at the same time on app launch?

host/webpack.config.js
```diff
@@ -75,7 +75,6 @@ module.exports = {
         },
         {
           '@openmsupply-client/common': {
-            import: path.join(__dirname, '../common'), // '@openmsupply-client/common',
```
joshxg, line 78: Removing these lines removed a 600kb bundle file in the dist folder when doing `yarn build`... I am unsure if that bundling would have ever been served and not sure what this line was meant to be doing - really I shouldnt be touching things I don't understand!

joshxg (approved):

mark-prins:
not sure how it was working for me without those imports.. very odd. I'll clean and start again now that you've fixed it up.

Thanks heaps - that's super helpful.

What I was expecting was that the imported package couldn't be tree shook / shaken, but that within the package it could tree shake, and therefore _common_ could import parts of material without bringing in the whole thing. I'm not certain now.. will see if I can try some refactoring to get that where I want it. 

I also want to be able to share the icons - perhaps building an icon package with only the required icons and including that. If it's already built then it won't have the bundling issues. Slightly more annoying to develop with, so will just experiment for now.

host/webpack.config.js
mark-prins: I'm not sure how that works for you! for me, that won't build
I get `can't resolve '@openmsupply-client/common'...` if I do. In another branch I've also changed this to `import: @openmsupply-client/common/src` and that resolves build errors for me.

joshxg: Weird..! Sorry, made it worse :) Were you planning to do anymore commits on this branch? If you are i'll leave it otherwise I'll revert it!

joshxg:
I think that it is tree shaking material etc. For example if I `export * from @material-ui/icons`, then I get a `2.5mb` bundle in the dist folder. I think where it is is not too bad? If we remove the icons we get a couple hundred kb bundle, non gzipped.. it's.. ok?

PR #17, opened 2021-07-30
#16 enable husky hook

### LINKED ISSUE #16: Enable husky pre commit hook
A few errors still but after fixing I think enabling the hook would be good

joshxg:
Fixes #16


.eslintrc.js
```diff
@@ -21,6 +21,9 @@ module.exports = {
   plugins: ['react', , '@typescript-eslint'],
   rules: {
     'require-jsdoc': 0,
+    'react/display-name': 'off',
```
joshxg, line 24: display-name arguably should be kept on.. 🤔

host/src/useLocalStorageSync.ts
```diff
@@ -1,17 +1,23 @@
 /* eslint-disable no-unused-vars */
 import React from 'react';
 
-declare global {
-  const localStorage: {
-    getItem: (key: string) => string;
-    setItem: (key: string, object: any) => void;
-  };
-}
+type LocalStorageSetter<T> = {
```
joshxg, line 4: I put the base tsconfig root to be `.` rather than `./src` and its giving the correct types for `window` and `localStorage` etc now

host/tsconfig.json
```diff
@@ -2,11 +2,5 @@
...
   "outDir": "./dist",
   "rootDir": "./src",
-  "strict": false,
-  "compilerOptions": {
-    "declaration": true,
-    "declarationDir": "./types",
-    "declarationMap": true,
-    "emitDeclarationOnly": true
-  }
+  "strict": false
```
joshxg, line 5: dunno what im doing here, it was just generating a lot of files.. 🤷

.eslintrc.js
mark-prins: probably should, but I do find it annoying

common/src/components/RemoteComponent.tsx
```diff
@@ -8,12 +8,12 @@ interface RemoteComponentProps {
...
 }
 
-export const RemoteComponent = ({
+export const RemoteComponent: FC<RemoteComponentProps> = ({
   url,
   scope,
   module,
   ...rest
-}: RemoteComponentProps): JSX.Element => {
+}: RemoteComponentProps) => {
```
mark-prins: this can be implied, now that you've fixed the `FC<RemoteComponentProps>` syntax

common/tsconfig.json
```diff
@@ -1,9 +1,3 @@
 {
-  "extends": "../tsconfig.base.json",
-  "compilerOptions": {
-    "declaration": true,
-    "declarationDir": "./types",
-    "declarationMap": true,
-    "emitDeclarationOnly": true
-  }
+  "extends": "../tsconfig.base.json"
```
mark-prins, line 2: In which case, might as well delete all of the generated declaration files, as they will get stale quickly. 
Just delete the `types` folders from host and dashboard.

host/src/useLocalStorageSync.ts
mark-prins: interesting. I had issues, I think with the type generation, which is why I changed it to `./src`
Great that it's working though, it's much better

host/tsconfig.json
mark-prins: I suspect we don't need the defs. and just to test the hook and prove the point.. bfcd849

mark-prins (approved):
great! very nice to have the typings working better, thanks. and good to have the hook enabled.
I've added a couple of changes; you're welcome to remove them, was tidying up defs files which aren't needed now, and corrected the syntax of a couple of components

joshxg:
🎉

PR #20, opened 2021-08-01
#12 Transaction MFE + Hot replacement

### LINKED ISSUE #12: Hot module replacement
Hot module replacement can work for a standalone app, but not through different apps (Doesn't seem to work for remotes and wpmf).

Setting this up for standalone apps would be helpful for me

joshxg:
Fixes #12  - kind of

- Aim was to get HMR 'working'. This only works when you are working on the standalone MFE. I.e. if we run this branch and go to `localhost:3005` - HMR is working, but it wouldn't work if I had a `TransactionService` imported into `Host`- the `TransactionService` wouldn't auto-replace.
- I think this isn't ideal but it is still useful. I could see us developing the 'main container' - i.e. with the appbar and top bar as components, and the `transactions` MFE could use that container as it's primary app, so that when you're developing transactions, you just use the specific url for that MFE. I think there's issues when a MFE starts utilizing other MFE's, though ..
- I chose `transactions` because the ui designs and also the backend team seem to be looking at customer invoices, so I figure we should just start with that.
- The `pmmmwh` dep is a bit dodge - needed this for it to work, which isn't in the latest release.. : https://github.com/pmmmwh/react-refresh-webpack-plugin/pull/392

### DIFF
--- a/packages/transactions/tsconfig.json
+++ b/packages/transactions/tsconfig.json
@@ -1,0 +1,6 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "outDir": "./dist",
+  "rootDir": "./src",
+  "strict": false
+}
--- a/packages/transactions/tsconfig.json
+++ b/packages/transactions/tsconfig.json
@@ -1,6 +1,5 @@
 {
   "extends": "../../tsconfig.base.json",
   "outDir": "./dist",
-  "rootDir": "./src",
-  "strict": false
+  "rootDir": "./src"
 }
--- a/packages/dashboard/tsconfig.json
+++ b/packages/dashboard/tsconfig.json
@@ -1,6 +1,5 @@
 {
   "extends": "../../tsconfig.base.json",
   "outDir": "./dist",
-  "rootDir": "./src",
-  "strict": false
+  "rootDir": "./src"
 }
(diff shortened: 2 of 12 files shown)


mark-prins:
the opening comment from `react-refresh-webpack-plugin` is gold: "First and foremost, this plugin is not 100% stable"
I think a little instability is worth it - this is only for the dev environment anyway and the hot replacement is soo good

packages/transactions/tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "outDir": "./dist",
+  "rootDir": "./src",
+  "strict": false
```
mark-prins, line 5: should we move `strict` into the base?

mark-prins (approved):
🎉  that works nicely!  😃

packages/transactions/tsconfig.json
joshxg: Hm. I don't really know what I'm doing but I removed the strict: false from everywhere as strict: true is in the base and strict seems good to start off with and we can relax it later (I don't know what it enforces .. )🤔 ?

joshxg: 50e3ca6

PR #22, opened 2021-08-02
Add basic mock apollo server

joshxg:
Just went with the easy, javascript with babel node and nodemon option .. tried to get typescript working but I gave up super quick


joshxg:
This is cute 

image

joshxg:
Actually getting ts errors in the js files which is annoying, going to take a look at that

joshxg:
Sorry! Good to go now.

I'm not sure what to do about the fields of a transaction. I'd like to have the 'real' field names we'll have in the future, but that discussion seems to be .. ongoing.. so will just go with this for now 🤷

mark-prins (approved):
👍  love it!
The `faker` is huuge - but awesome.

PR #23, opened 2021-08-02
update theme and style shared elements

### LINKED ISSUE #2: Set up theme
Create a shared theme and implement in the current UI

### LINKED ISSUE #15: Style host elements
Implement styling for
- Background
- Drawer
  - Menu
  - Menu Items
- Appbar
- favicon

mark-prins:
Let's just say that this Fixes #2 and Fixes #15

Made some changes to the theme, in order to support custom items and implemented some of the styling from zeplin.


packages/host/src/AppDrawer.tsx
```diff
@@ -62,21 +96,63 @@ const ListItemLink: React.FC<ListItemLinkProps> = props => {
...
     <li>
-      <ListItem selected={!!selected} button component={CustomLink}>
-        <ListItemIcon>{props.icon}</ListItemIcon>
+      <ListItem
+        selected={!!selected}
+        button
+        component={CustomLink}
+        className={className}
+      >
+        {props.icon}
```
joshxg, line 112: Maybe wrap in `<ListItemIcon>`?

packages/host/src/AppDrawer.tsx
```diff
@@ -62,21 +96,63 @@ const ListItemLink: React.FC<ListItemLinkProps> = props => {
       )),
     [props.to]
   );
+  const className = clsx(
+    props.classes['drawerMenuItem'],
+    !!selected && props.classes['drawerMenuItemSelected']
+  );
 
   return (
     <li>
```
joshxg, line 105: this `li` seems unneeded?

joshxg (approved):
It's amazing!  Looking great

Just a small thing with the text when the drawer is closed:
image

Wrapping the icon in `ListItemIcon` seems to fix it, but then the selected and hovering background gets a bit ugly.. might need to use the drawer open state to conditionally render the text?

Spent some time figuring out theme stuff and made some useless comments about putting things in the theme, which could be done later, if they should be put there..?

Also maybe need to ask Manu to put an 'open' button? 🤔

mark-prins:
> Also maybe need to ask Manu to put an 'open' button?

His thought with the right hand drawer was to make it responsive; if the screen size was big enough it would be expanded otherwise collapsed.. though it did have the option to open & close. 
I thought to make the msupply guy open & close this one. at least for now.

mark-prins:
> Just a small thing with the text when the drawer is closed

yeah, I did sort of give up on that; will do it now.
I was avoiding the `ListItemIcon` in order to remove some of the overhead. and that was doing stuff which I didn't want.. Will have a look.

packages/host/src/AppDrawer.tsx
```diff
@@ -1,38 +1,69 @@
...
+    '&:hover': {
+      backgroundColor: '#fff',
+      borderRadius: 16,
+      boxShadow:
+        '0 8px 16px 0 rgb(96 97 112 / 16%), 0 2px 4px 0 rgb(40 41 61 / 4%)',
+    },
+  },
+  drawerMenuItemSelected: {
+    backgroundColor: '#fff!important',
+    borderRadius: 16,
```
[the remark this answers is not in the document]
mark-prins: that gives a radius of `4` - I went with the specific value here, thinking that this lozenge menu item will only appear in the drawer. could well be wrong about that.

packages/host/src/AppDrawer.tsx
mark-prins: I took them out and bathed in the delight of minimalism, and then noticed that it was all wrapped in a `List` so added them back in to be syntactically complete 🤷‍♂️

mark-prins:
that was fun.. 
- I've added in the lower menu items
- and the divider between them
- hidden the text when the drawer is closed
- updated the style of the items when closed
- added back the `ListItemIcon`
- updated the mSupply Guy, wrapping him in an `IconButton` and toggling the drawer
- Used material shadows ( which aren't quite right, but I like the idea )
- Added colours to the palette
- Added mixins

packages/host/src/AppDrawer.tsx
joshxg: Ah I just read the docs - how silly that `*L*ist*I*tem` isn't rendering a `<*l**i*>` 😢

joshxg:
> that was fun..

Sarcasm or serious? 🤔

PR #24, opened 2021-08-02
#3 react query

### LINKED ISSUE #3: Implement react-query
Add the ability to fetch data using react-query

joshxg:
Fixes #3 

- This is really basic and I'm PRing to show the bones of an idea for editing
image
- Just showing a single transaction and able to edit the customer name, heh.
- General idea for the custom hook is there which has a lot of improvements to be made. The rest is boiler plate hooking things up


packages/transactions/src/api.ts
```diff
@@ -0,0 +1,108 @@
...
+ *
+ * - Easier editing mechanism. I don't want to have to use `setDraft({...draft, [myField]: newValue})
+ * every time - especially if there is more complexity - i.e. a customer requisition editing days of
+ * stock which might effect many columns. Possibly the draft could be a more complex class instance or
+ * we use a more functional approach and have functions i.e. updateDaysOfStock(draft, newValue)
+ *
+ * - Just use redux :thinking: :shrug:
+ */
+
+export const useDraftDocument = <DocumentType>(
```
joshxg, line 76: This is the meat and potatoes.

It's tiny but hopefully encapsulate what editing a record will look like

joshxg:
Merged in your host styling branch- i've left this on the `Transaction` route for now - not ideal, but im not sure what to do with the packages structures, exactly 🤔

packages/host/src/AppDrawer.tsx
```diff
@@ -78,7 +79,7 @@ const useStyles = makeStyles((theme: Theme) => ({
     [theme.breakpoints.up('sm')]: {
       width: theme.spacing(9),
     },
-    '& li > a': { borderRadius: 20, height: 40, padding: 10, width: 40 },
+    '& li > a': { borderRadius: 20, height: 32, padding: 10, width: 40 },
```
mark-prins, line 82: but why? now the circles aren't circles - in Manu's design the closed drawer had these as 40px circles and the larger version as 32px high lozenges

packages/transactions/src/TransactionService.tsx
```diff
@@ -1,5 +1,56 @@
-import { FC } from 'react';
+import React, { FC } from 'react';
+import { request } from 'graphql-request';
+import { query, mutation, useDraftDocument } from './api';
 
-const TransactionService: FC = () => null;
+interface Transaction {
```
mark-prins, line 5: Would be nice to have these interfaces shareable and exported by `common`

packages/transactions/src/TransactionService.tsx
```diff
@@ -1,5 +1,56 @@
...
+  customer: string;
+  supplier: string;
+  total: string;
+  id: string;
+  date: string;
+}
+
+const queryFn = async (): Promise<Transaction> => {
+  const result = await request('http://localhost:4000', query);
```
mark-prins, line 14: would be good to implement `.env` or similar at some point

packages/transactions/src/App.tsx
```diff
@@ -1,10 +1,16 @@
...
 import { LoadingApp } from '@openmsupply-client/common';
+import TransactionService from './TransactionService';
+
+const queryClient = new QueryClient();
 
 const App: FC = () => (
-  <React.Suspense fallback={<LoadingApp />}>
-    <span>Transactions</span>
-  </React.Suspense>
+  <QueryClientProvider client={queryClient}>
```
mark-prins, line 9: Does the provider have to be in here as well as in the host?

packages/transactions/src/TransactionService.tsx
```diff
@@ -1,5 +1,56 @@
...
+  const patch = { transactionPatch: updated };
+  const result = await request('http://localhost:4000', mutation, patch);
+
+  const { updateTransaction } = result;
+
+  return updateTransaction;
+};
+
+const TransactionService: FC = () => {
+  const { draft, setDraft, save } = useDraftDocument<Transaction>(
```
mark-prins, line 30: that's very nice

packages/transactions/src/api.ts
```diff
@@ -0,0 +1,112 @@
...
+ * - Easier editing mechanism. I don't want to have to use `setDraft({...draft, [myField]: newValue})
+ * every time - especially if there is more complexity - i.e. a customer requisition editing days of
+ * stock which might effect many columns. Possibly the draft could be a more complex class instance or
+ * we use a more functional approach and have functions i.e. updateDaysOfStock(draft, newValue)
+ *
+ * - Might be nice/easy/better/worse? to be able to more globally access the current draft. For example,
+ * the draft hook will be used at the 'root' of some page, and will be required to be drilled down through
+ * components which want to know the current draft values. It's not too difficult to make the value 'global'
+ * i.e. can use context, or a simple state management tool like Jotai or Recoil or..
+ * - Just use redux :thinking: :shrug: :lol:
```
mark-prins, line 73: tempting to just use redux! I agree though.. nice to avoid that overhead if we don't _actually_ need it and I don't think that we do. The wiring up of events, such as failure to save or potential conflict would be interesting to get in.

packages/transactions/src/api.ts
```diff
@@ -0,0 +1,112 @@
...
+ */
```
mark-prins, line 74: some sort of discard / cancel mechanism? I don't see it as necessary, as navigating away will lose the state.. just wonder if it would be nice for completeness

mark-prins (approved):
woo! there's a lot going on here, with a lot to think about and extend. 
The approach is great - I like where you're heading with this

packages/host/src/AppDrawer.tsx
joshxg: Ah right, sorry. Reverted: b3995da

I was trying to adjust this thing I noticed which irrationally annoyed me.. the shifting of the menu items due to the height change

height-change

V

height-change-no

packages/transactions/src/App.tsx
joshxg: Yeah. This is the `Transaction` MFE `App` which will load when navigating to `localhost:3005` - it doesn't have the host but does have HMR. The `TransactionService` is what is exposed as a remote.

packages/transactions/src/api.ts
joshxg: Oh great idea - "Reset"/"Discard changes" would be great to revert back to the version the server currently has.

packages/host/src/AppDrawer.tsx
mark-prins: yeah, bothered me too.
it isn't perfect - but this is pretty good:
```
    '& li > a': {
      borderRadius: 20,
      height: 40,
      margin: '12.5px 0',
      padding: 10,
      width: 40,
    },
```
under `drawerPaperClose`. Makes the circles bigger and round, while keeping their vertical position

PR #33, opened 2021-08-05
#shift nav component

joshxg:
So..

I was working on the customer container, and I wanted the `CustomerContainer` to expose the navigation for itself such that if you wanted to add a new route in the customers package, you only need to touch the customers package and not add a new route to the host.

So.. I shifted the link component into common and all hell broke loose.

I started having two material ui packages downloaded which was causing issues with the theme+styles. Then I had issues with multiple `react-router-dom`s as that was also being shared.

Confusing.

So this is what I've done to get this working.

All imports of material ui components *within* the `common` package are importing from `@material-ui/core` and from other packages, importing from `@openmsupply/common`. This has made the bundle of material ui bloat from `~2mb` development bundle to `~2.9mb`. However, the built size is still small as the development bundle has a lot of code marked as `unused` which is only trimmed in a production build.


packages/common/src/index.ts
```diff
@@ -1,16 +1,9 @@
...
-import * as utils from './utils';
-
-import { Link as ReactRouterLink } from 'react-router-dom';
+export * from 'graphql-request';
+export * from 'react-query';
+export * from 'react-query/devtools';
 
-export * from './components';
+export * from './utils';
+export * from './ui';
```
joshxg, line 6: removed react-router from being shared here, and instead added it as an explicit dependency for each package and 'hoisted it' with `lerna boostrap`.

Before I did this, I was getting an error that I couldn't use `Routes` outside of a `Router` which indicated that the `Routes` I was importing in `TransactionService` was from a different package than what the `BrowserRouter` was imported from.

I'm still figuring out what exactly went wrong here and why I seem to have to have `react-router-dom` be an explicit dependency in the lerna hoisted modules rather than importing from the common package..

packages/common/src/components/index.ts
```diff
@@ -1,11 +0,0 @@
-import CircularProgress from '@material-ui/core/CircularProgress';
-import Divider from '@material-ui/core/Divider';
-import Typography from '@material-ui/core/Typography';
-import ListItemText from '@material-ui/core/ListItemText';
-
-export { LoadingApp } from './LoadingApp';
-export { DataGrid } from '@material-ui/data-grid';
-export { RemoteComponent } from './RemoteComponent';
-export * from './icons';
-export * from './layout';
```
joshxg, line 10: I renamed `components` -> `ui` because I wanted a `components` directory for the floating components..

packages/common/src/styles/ThemeProvider.tsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import { ThemeProvider as MuiThemeProvider } from '@material-ui/core/styles';
+import { ThemeProvider as MuiThemeProvider } from '@material-ui/core';
```
joshxg, line 2: I don't know. Still experimenting, but importing everything I can from `@material-ui/core` seems to help ensure there are not multiple styles modules imported.

packages/common/src/styles/index.ts
```diff
@@ -1,5 +1,3 @@
-import makeStyles from '@material-ui/core/styles/makeStyles';
-import CssBaseline from '@material-ui/core/CssBaseline';
-import ThemeProvider from './ThemeProvider';
+import AppThemeProvider from './ThemeProvider';
 
-export { CssBaseline, makeStyles, ThemeProvider };
+export { AppThemeProvider };
```
joshxg, line 3: I am experimenting *again* with exporting all from `@material-ui/core`, which includes these packages..

packages/common/src/styles/theme.ts
```diff
@@ -1,9 +1,10 @@
 import { MixinsOptions as MuiMixinsOptions } from '@material-ui/core/styles/createMixins';
 import {
+  createTheme,
   Theme as MuiTheme,
   ThemeOptions as MuiThemeOptions,
-} from '@material-ui/core/styles';
-import createTheme from '@material-ui/core/styles/createTheme';
+} from '@material-ui/core';
```
joshxg, line 6: Again just trying to export everything from `@material-ui/core` was helping me..

packages/common/src/styles/theme.ts
```diff
@@ -45,7 +46,7 @@ export type ThemeOptions = Modify<
   }
 >;
 
-export type Theme = Modify<
+export type AppTheme = Modify<
```
joshxg, line 49: I'm getting issues here.

If I export this from the `styles/index.ts` like:

image

and then try to import it from elsewhere like:
image


I get no errors and the type is resolved in the IDE fine.

However when I try to compile I get

image

Possibly something to do with import order or something 🤔

packages/common/src/ui/components/NavLink.tsx
```diff
@@ -0,0 +1,62 @@
+import React, { FC } from 'react';
```
joshxg, line 1: Just copy pasted.

I would like to try and use the styled-components style for this 😁

joshxg (commented):
Getting a bit annoyed with this 😁 

Going to play around more today to try and really figure out what's going on!

joshxg:
Working through https://material-ui.com/guides/minimizing-bundle-size/ at the moment

packages/common/src/styles/index.ts
```diff
@@ -1,5 +1,8 @@
+import AppThemeProvider from './ThemeProvider';
 import makeStyles from '@material-ui/core/styles/makeStyles';
 import CssBaseline from '@material-ui/core/CssBaseline';
-import ThemeProvider from './ThemeProvider';
 
-export { CssBaseline, makeStyles, ThemeProvider };
+import { ApplicationTheme } from './theme';
+type AppTheme = ApplicationTheme;
```
joshxg, line 6: Without this I keep getting a 'type' not found error but compilation is fine...

joshxg (commented):
Hm.

Back more or less to what it was now.. 🙄

packages/common/src/styles/index.ts
mark-prins: Not sold on the name, `AppTheme` relates to which app?
I was wondering if we should call it `MSupplyTheme` or similar. This is the global, customised theme, right? which is why the rename?

packages/customers/package.json
```diff
@@ -25,11 +24,13 @@
...
-    "serve": "serve dist -p 3006",
+    "serve": "serve dist -p 3007",
     "clean": "rm -rf dist"
   },
   "dependencies": {
+    "history": "^5.0.0",
     "react": "^16.13.0",
-    "react-dom": "^16.13.0"
+    "react-dom": "^16.13.0",
+    "react-router-dom": "^6.0.0-beta.0"
```
mark-prins, line 34: it's a pity that we have to add this to each app, it might make version updates a problem later. I tried removing though.. and that breaks.

mark-prins (approved):
needed to be done it seems.  Merge away!

joshxg:
re `AppTheme`. Seems my comment went away!

I was trying to solve the problem you so elegantly solved in your PR with `export type` which i've fixed here 
df8fc60

PR #35, opened 2021-08-06
Add type augmentation for theme

### LINKED ISSUE #34: Make Theme type inferred by makeStyles
In https://github.com/openmsupply/openmsupply-client/pull/33 i am having some trouble with exporting the theme.

I realise we shouldn't really need to export the theme and that `makeStyles` should be inferring the type.

After reading the docs it seems the recommended approach is to use 'module augmentation'?

https://next.material-ui.com/guides/typescript/#customization-of-theme

joshxg:
Fixes #34 

I combined the two approaches and `makeStyles` now infers the correct theme type 🤷


mark-prins (approved):
Very nice! that's much better with the augmentation

PR #37, opened 2021-08-06
#36 customer package navigation

### LINKED ISSUE #36: CustomerContainer package navigation
Now that we can use the `NavLink` component, have `CustomerContainer` expose a component `Nav`, which will be what `Host` renders in the `AppDrawer`, rather than explicitly defining `Customers` routes

joshxg:
Fixes #36

This is adding the customer navigation to the app drawer

image

image

Thought this was enough for the PR for now. Some things in new PRs

- The drawer state is not correctly shared and can get munged. Need to essentially share the state between all `useDrawer` hook usages. Few options here
- Similar to the drawer state, the 'title' needs to be updated better. Can continue using the service context that's in the host package, however that will cause a re-render of the whole app which isn't ideal and I think a solution for both the drawer and title could be used.
- Breadcrumbs - a bit annoying to do this one but likely going to be a similar state that needs to be updated from many different places, same as the above two
- Styling: Not looking correct right now
- 'selected state' - need an update to the `common/AppNavLink` component
- Routes/navigation tidy in general - constants for routes/components to encapsulate more of the code

### DIFF
--- a/packages/common/src/hooks/useDrawer.ts
+++ b/packages/common/src/hooks/useDrawer.ts
@@ -1,0 +1,23 @@
+import { useLocalStorageSync } from './useLocalStorageSync';
+
+interface DrawerController {
+  open: boolean;
+  closeDrawer: () => void;
+  openDrawer: () => void;
+}
+
+export const useDrawer = (): DrawerController => {
+  const { value, setItem } = useLocalStorageSync<boolean>(
+    '@openmsupply-client/appdrawer/open'
+  );
+
+  return {
+    open: !!value,
+    closeDrawer() {
+      setItem(false);
+    },
+    openDrawer() {
+      setItem(true);
+    },
+  };
+};
--- a/packages/common/src/ui/components/NavLink.tsx
+++ b/packages/common/src/ui/components/NavLink.tsx
@@ -8,7 +8,7 @@ import { useMatch, Link } from 'react-router-dom';
 const useStyles = makeStyles(theme => ({
   drawerMenuItem: {
     height: 32,
-    marginTop: '20px',
+    marginTop: 20,
     '& svg': { ...theme.mixins.icon.medium },
     '&:hover': {
       backgroundColor: theme.palette.background.white,
--- a/packages/common/src/ui/components/NavLink.tsx
+++ b/packages/common/src/ui/components/NavLink.tsx
@@ -8,7 +8,7 @@ import { useMatch, Link } from 'react-router-dom';
 const useStyles = makeStyles(theme => ({
   drawerMenuItem: {
     height: 32,
-    margin: '20px 0',
+    marginTop: '20px',
     '& svg': { ...theme.mixins.icon.medium },
     '&:hover': {
       backgroundColor: theme.palette.background.white,
@@ -22,7 +22,6 @@ const useStyles = makeStyles(theme => ({
 }));
 
 interface ListItemLinkProps {
-  classes: Record<string, string>;
   to: string;
   icon: JSX.Element;
   text?: string;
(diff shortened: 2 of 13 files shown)


packages/common/src/ui/components/NavLink.tsx
```diff
@@ -8,7 +8,7 @@ import { useMatch, Link } from 'react-router-dom';
 const useStyles = makeStyles(theme => ({
   drawerMenuItem: {
     height: 32,
-    margin: '20px 0',
+    marginTop: '20px',
```
joshxg, line 11: I read [this](https://mxstbr.com/thoughts/margin/) the other day from another article talking about the old `spacer.gif` .. then playing around with this margin made me think I agree 🤔

packages/host/src/types.d.ts
```diff
@@ -9,3 +9,7 @@ declare module 'transactions/TransactionService' {
 declare module 'customers/CustomerContainer' {
   export default function (): JSX.Element;
 }
+
+declare module 'customers/Nav' {
+  export default function (): JSX.Element;
+}
```
joshxg, line 15: this also getting tedious already 🤔

packages/common/src/ui/components/index.ts
```diff
@@ -1,8 +1,9 @@
+import Collapse from '@material-ui/core/Collapse';
 import Divider from '@material-ui/core/Divider';
 import Typography from '@material-ui/core/Typography';
 import CircularProgress from '@material-ui/core/CircularProgress';
 
 export * from './LoadingApp';
 export * from './RemoteComponent';
 export { AppNavLink } from './NavLink';
-export { Typography, CircularProgress, Divider };
+export { Typography, CircularProgress, Divider, Collapse };
```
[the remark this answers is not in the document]
mark-prins: already is 😉 
Just trying to keep the bundle small. I hoped that we hit a point where we are just reusing things rather than adding.. but that might be a long way down the road?

packages/common/src/ui/components/NavLink.tsx
mark-prins: If it's just a single number, then better to:

```suggestion
    marginTop: 20,
```

packages/host/src/types.d.ts
mark-prins: yep

packages/common/src/hooks/useDrawer.ts
```diff
@@ -0,0 +1,23 @@
...
+interface DrawerController {
+  open: boolean;
+  closeDrawer: () => void;
+  openDrawer: () => void;
+}
+
+export const useDrawer = (): DrawerController => {
+  const { value, setItem } = useLocalStorageSync<boolean>(
+    '@openmsupply-client/appdrawer/open'
+  );
```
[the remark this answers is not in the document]
mark-prins: yes - and like you say, the title too. and language? theme?

packages/common/src/ui/components/NavLink.tsx
mark-prins: the margin article was interesting - and good food for thought. I tend to agree - but margins are soooo useful.

packages/customers/src/Nav.tsx
```diff
@@ -0,0 +1,42 @@
...
+const Nav: FC = () => {
+  const { isActive } = useNestedNav('customer/*');
+
+  return (
+    <>
+      <AppNavLink to="customers" icon={<Customers />} text="Customers" />
+      <Collapse in={isActive}>
+        <List>
+          <AppNavLink
+            to="/customers/customer-invoice"
```
mark-prins, line 32: `customer-invoices`?

mark-prins: or just `invoices`? it is under customers after all.

mark-prins: is it time to add a global enum for routes? I see `customer-invoice` a number of times already in this PR

mark-prins (approved):
super!

packages/customers/src/Nav.tsx
joshxg: yes - issue so I can merge this https://github.com/openmsupply/openmsupply-client/issues/43

PR #42, opened 2021-08-09
#29 localisation

### LINKED ISSUE #29: Localisation
Before getting to far would be good to add in some localisation.

Think the requirements are
- Able to localise 🙃 
- Can allow specific deployments to augment the localisations. I.e A specific customer would prefer the english string for `item` be `product` [there are a lot of conditionals in current mSupply for this purpose].
- Limit the amount of strings needing to be downloaded. Current mSupply localisations files are 2.3Mb. We don't want a user to have to download all of this in one go and bloat the bundle size- so localisations should be package specific
- We want to be able to *share* strings as much as possible between packages to limit the amount of translations required.

mark-prins:
Fixes #29 

Went around in circles and in the end, didn't end up where I wanted.
What I was trying to do, was to have each app be able to pick out the translations it requires from common and only load the subset of requires translations.

Turns out to be a tricky problem. Best I had was to export each string in common - so storing as modules rather than json ( tedious ) and then import each one individually in the consuming app ( more tedious ) but then had trouble with the selected locale.

Also had an issue with `swc` which does not allow dynamic imports as a webpack loader when using the `.swc` configuration. Had to push the config into `webpack.config.js` to get that working. It now bundles each language individually and the client app only downloads the current language. Not ideal but better than nothing.

This also creates an issue with the standard components, which don't recognise the `IntlProvider` - so we have to use `useFormattedMessage()` hook instead. That's nice as it also checks the keys at compile time, so will catch usage problems for us.


packages/common/src/intl/IntlProvider.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const { locale } = props;
+  const [messages, setMessages] = React.useState<Record<string, string> | null>(
+    null
+  );
+  React.useEffect(() => {
+    importMessages(locale as SupportedLocales).then(setMessages);
+  }, [locale]);
+
+  return messages ? (
+    <ReactIntlProvider locale={locale} messages={messages} key={locale}>
```
mark-prins, line 23: `key` is to encourage the component to re render - perhaps not necessary if we aren't using some kind of dynamic language selection

packages/common/src/intl/intlHelpers.ts
```diff
@@ -0,0 +1,48 @@
+import { useIntl } from 'react-intl';
+import type { PrimitiveType } from 'intl-messageformat';
+
+// "import type" ensures en messages aren't bundled by default
+import * as sourceOfTruth from './locales/en.json';
+
+export type SupportedLocales = 'en' | 'fr' | 'pt';
+export type LocaleMessages = typeof sourceOfTruth;
```
mark-prins, line 8: really like this idea - of taking the keys from the main language file and checking all usages against this set of allowable keys

packages/common/src/intl/intlHelpers.ts
```diff
@@ -0,0 +1,48 @@
...
+};
+
+// return type on this signature enforces that all languages have the same translations defined
+export const importMessages = (
+  locale: SupportedLocales
+): Promise<LocaleMessages> => {
+  switch (locale) {
+    case 'en':
+      return import(
+        /* webpackMode: "lazy", webpackChunkName: "en_json" */
```
mark-prins, line 34: perhaps don't need to break these out to separate bundles, but I rather like it. allows the browser to download and cache separately

packages/common/src/store/store.ts
```diff
@@ -87,7 +87,7 @@ interface FederatedStore
   reducerManager?: ReducerManager;
 }
 
-export const makeStore = () => {
+export const makeStore = (): FederatedStore => {
```
mark-prins, line 90: offroad tidy

packages/common/src/utils/cookies.ts
```diff
@@ -0,0 +1,6 @@
+export const readCookie = (name: string): string => {
```
mark-prins, line 1: was using state in the host app initially - am thinking that we would likely have display language tied to a user, as per desktop and mobile; in which case the language is set prior to loading the app. have implemented as a cookie to simulate that approach and keep the state a little cleaner. Easy enough to have it dynamic if we want to though 🤷‍♂️

packages/dashboard/webpack.config.js
```diff
@@ -27,6 +27,16 @@ module.exports = {
         test: /\.[t|j]sx?$/,
         loader: 'swc-loader',
         exclude: /node_modules/,
+        options: {
```
mark-prins, line 30: same options as before. they work when like this though..

packages/host/src/AppBar.tsx
```diff
@@ -10,11 +12,12 @@ const useStyles = makeStyles(theme => ({
   appBar: {
     left: 72,
     position: 'absolute',
-    zIndex: theme.zIndex.drawer + 1,
     transition: theme.transitions.create(['width', 'margin'], {
       easing: theme.transitions.easing.sharp,
       duration: theme.transitions.duration.leavingScreen,
     }),
+    width: 'calc(100% - 72px)',
```
mark-prins, line 19: another offroad tidy up

packages/host/src/AppDrawer.tsx
```diff
@@ -45,7 +45,7 @@ const useStyles = makeStyles((theme: Theme) => ({
   },
   drawerMenuItem: {
     height: 32,
-    margin: '20px 0',
+    margin: '16px 0',
```
mark-prins, line 48: tweaked these styles to make the left navigation nicer when opening and closing the drawer

joshxg (approved):
Wow its awesome.

I really like the hooks. Think its a good thing to define our own api for localization anyway so we dont have to change app code throughout if formatjs turns out to be the wrong one.

The key checking is also wicked.

Some things i thought aboot
- merge the localization json with a 'customization' json? Lets anyone override strings
- is en.json gonna get too big?
- should be able to pass down a `setLanguage` func using the provider to set the language  to rerender rather than reload the page similar to theme and dark mode?
- With this, if we only need to add a string, we need to bump the common package? Would be nice to be able to inject localisation strings from packages into the translations provider when they load but probably not needed right now..

mark-prins:
> merge the localization json with a 'customization' json? Lets anyone override strings

oh yeah, forgot that bit. will make another issue for that

> is en.json gonna get too big?

perhaps. I thought maybe worth giving it a go. wondered about splitting it up, kinda like mobile and desktop both do, but it is a pain knowing which bit to pull strings from. and desktop has deviated a long way from the original categorising.

> should be able to pass down a setLanguage func using the provider to set the language to rerender rather than reload the page similar to theme and dark mode?

I did have that in, and took it out. Had the language in state on the host app and decided that we probably wouldn't have a need for that. I'm now thinking that there's no reason not to - if we have a global state then lets just put language in there.

> With this, if we only need to add a string, we need to bump the common package? Would be nice to be able to inject localisation strings from packages into the translations provider when they load but probably not needed right now..

true. that's where I wandered for a while - having the ability for client apps to set messages. it got tricky, particularly with validating the translation keys. something to look at again I think

Issue #39, opened 2021-08-06
Implement global storage for drawer and title states


joshxg:
With #38 we can implement a global storage solution for the drawer and title states

PR #44, opened 2021-08-11
Add global state hook thing

### LINKED ISSUE #38: Implement global local storage setting solution
It would be very useful to set a local storage variable from different places in the app and ensure the value is reactive within the app.

Essentially integrating the `useLocalStorageSync` hook with a global state variable.

joshxg:
Fixes #38 and #39 

I went with zustand 🤣  It covers quite a few edge cases, the code is nice and easy and the bundle is also tiny.

valtio is my favourite, but it just seems less mature for some reason 🤷  

If we want to swap to something else, we just need to change the implementation of `useDrawer` (or whatever other hooks we make that should be global) and keep the same interface. Seems fine 🤷 

Also couple other bonuses/possible things..
- https://github.com/roadmanfong/zustand-persist - rather than the subscribe for local storage..
- https://github.com/beerose/simple-zustand-devtools 
- https://github.com/Nowsta/zustand-saga ( ... 🚶  )
- https://github.com/Tom-Julux/shared-zustand 

Also realise that the `titles` will be replaced for breadcrumbs


mark-prins (approved):
wicked! simple, elegant and easy to use. and it works :)

PR #46, opened 2021-08-12
#41 update navlink

### LINKED ISSUE #41: Update NavLink component
The styling for `NavLink`  needs updating as seen in #37 - Also should more correctly handle the selected state

mark-prins:
Fixes #41 - somewhat.
there's still a problem where you can
- open the drawer
- click on 'Customers'
- reload the page
- close the drawer

and get: 
Screen Shot 2021-08-12 at 11 47 10 AM

For that one, I'd like to use the global state - so will merge that PR before fixing this bit.

I also don't like the sub menu style, and had added an icon and indentation before realising that the already implemented style matched zeplin. To me the sub menu is lost; and I added the icon to cope with closing the drawer which is funky when there's no icon.

To that end, I also added tooltips, also not in zeplin: 

Screen Shot 2021-08-11 at 6 20 04 PM

when collapsed, it is hard to know what each menu icon represents. Thought this was a low-cost way to help web based users. 
Have reverted the other off-road design tweaks.


packages/host/src/AppDrawer.tsx
```diff
@@ -92,15 +92,14 @@ const Menu: React.FC<MenuProps> = ({ classes }) => {
...
-        <React.Suspense fallback={null}>
-          <CustomersNav />
-        </React.Suspense>
         <AppNavLink
           to="dashboard"
           icon={<Dashboard />}
           text={formatMessage('app.dashboard')}
         />
-
+        <React.Suspense fallback={null}>
```
mark-prins, line 100: moved down in the list, as the dashboard should be the top, I think

packages/common/src/ui/components/NavLink.tsx
```diff
@@ -22,21 +22,24 @@ const useStyles = makeStyles(theme => ({
 }));
 
 interface ListItemLinkProps {
-  to: string;
-  icon: JSX.Element;
+  end?: boolean; // denotes lowest level menu item, using terminology from useMatch
```
mark-prins, line 25: it's a terrible name for a parameter

packages/common/src/ui/components/NavLink.tsx
```diff
@@ -22,21 +22,24 @@ const useStyles = makeStyles(theme => ({
...
+  icon?: JSX.Element;
   text?: string;
+  to: string;
 }
 
 export const AppNavLink: FC<ListItemLinkProps> = props => {
   const classes = useStyles();
-  const selected = useMatch({ path: props.to });
+  const { end, icon = <span style={{ width: 20 }} />, text, to } = props;
```
mark-prins, line 33: the spacer might need replacing at some point? In which case I thought to move it up here

packages/customers/src/Nav.tsx
```diff
@@ -22,16 +23,25 @@ const useNestedNav = (path: string) => {
 
 const Nav: FC = () => {
   const { isActive } = useNestedNav('customer/*');
-
+  const formatMessage = useFormatMessage();
```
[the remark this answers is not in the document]
mark-prins: I like it - and have gone ahead. sets the pattern which will be copied, so good to change now.
Technically, the 'formatMessage' is a little closer to what the function does.. it is much better using `t('something)` though :)

PR #47, opened 2021-08-16
#39 global storage title states

### LINKED ISSUE #39: Implement global storage for drawer and title states

mark-prins:
Fixes #39 

Updating the app bar title with the current route. Had to switch the host context over to store the `titleKey` rather than the `title` as that wasn't updating - easier to store the key and display the title in the `AppBar` using the current locale when it re-renders. This way, changes made in the language menu update correctly.

There is one spot where things don't behave - it's pretty minor so haven't dug into it


packages/customers/src/CustomerContainer.tsx
```diff
@@ -1,11 +1,36 @@
...
+
+  useEffect(() => setTitleKey(titleKey), [titleKey]);
+  switch (titleKey) {
+    case 'app.customer_invoices':
+      return <TransactionService />;
+    case 'app.customer_requisitions':
+      return <RequsitionsService />;
+    default:
+      return <></>;
+  }
```
joshxg, line 33: Should this be using a router? The `TransactionService` has one, which might be a little confusing..

Maybe a router here which can match on the sub menu items i.e. `customer-invoice`, and then rather than a `TransactionService`, there is a `CustomerInvoiceService` which has a router which matches on `/` and `:id`..?

Something for later, I think.. to figure out exactly how the routing should work..

packages/customers/src/CustomerContainer.tsx
```diff
@@ -1,11 +1,36 @@
...
+    : isRequsition
+    ? 'app.customer_requisitions'
+    : 'app.customers';
+
+  useEffect(() => setTitleKey(titleKey), [titleKey]);
+  switch (titleKey) {
+    case 'app.customer_invoices':
+      return <TransactionService />;
+    case 'app.customer_requisitions':
+      return <RequsitionsService />;
```
joshxg, line 30: oooo my first ever spelling comment- `Requisition` 🥇 

Alsoo, plural and non-plurals (Is that the right word???) 🤔

packages/common/src/hooks/useHostContext.ts
```diff
@@ -0,0 +1,21 @@
+import create from 'zustand';
+import { LocaleKey, SupportedLocales } from '../intl/intlHelpers';
+
+type HostContext = {
+  locale: SupportedLocales;
+  titleKey?: LocaleKey;
+  setLocale: (locale: SupportedLocales) => void;
+  setTitleKey: (titleKey: LocaleKey) => void;
```
joshxg, line 8: Also have realised.. the title should be `breadcrumbs`? https://material-ui.com/components/breadcrumbs/

So.. maybe should be a `useBreadcrumbs` hook? Similar to the title one but possibly will need a way to append rather than set, and also a way to reset?

joshxg (approved):

packages/common/src/hooks/useHostContext.ts
```diff
@@ -0,0 +1,21 @@
...
+};
+
+const localStorageKey = '@openmsupply-client/localisation/locale';
+
+export const useHostContext = create<HostContext>(set => ({
+  locale: (localStorage.getItem(localStorageKey) as SupportedLocales) ?? 'en',
```
joshxg, line 14: I wonder actually- what if we wrapped the provider in our own provider.. 



```
const TranslationContext.Provider = React.createContext(getFromLocalStorageOrSomething?)

const TranslationProvider = ({children}) => {
 const [locale, setLocale] = useLocalStorageState(getLocaleFromLocalStorage)
 return <TranslationContext.Provider><IntlProvider locale={locale}>{children}</IntlProvider> </TranslationContext.Provider>
}

const useChangeLocale = () => useContext(TranslationContext)
```

I dunno

mark-prins: yup. I went on that journey and had too many problems, so wound it all back again.
I thought to expose the other helpers too.. so I could `const { t, setLocale, etc } = useTranslationContext()` but I think now that this wouldn't get past the review stage.

Still like the idea of wrapping the `IntlProvider` in our own context.

packages/common/src/hooks/useHostContext.ts
mark-prins: yeah, a little unclear from the designs. I'll have another go.

packages/customers/src/CustomerContainer.tsx
mark-prins: Nice, thanks!

packages/customers/src/CustomerContainer.tsx
mark-prins: 😄  Yes, I did this already too!
My problem was that adding a `Router` here affects the routes which `TransactionService` sees. In the way that the route is `/customers/customer-invoice` but is `/customer-invoice` when the `TransactionService` sees it. I also thought that this was not a permanent solution. 

Will have another look ..

packages/common/src/hooks/useHostContext.ts
mark-prins: hehehe.. have gone with a different approach. I rather like it: now using the navigation to drive the breadcrumb rather than have the pages themselves push their title.
d166be8
300f61e

PR #55, opened 2021-08-17
#testing library

joshxg:
2k Lines 😱 

- All dev deps so shouldn't have an impact on production..
- Adds jest and https://testing-library.com/docs/queries/about and a couple other packages to help with matching and assertions
- The config to get jest to work was a bit.. ugh.. so I changed the name of the `tsconfig.base.json` to just `tsconfig.json` and it worked and seems to be ok... ?
- The tests are maybe not the best examples? But meant to show a little bit.. I really like testing generic js/ts code with jest and if I can test ui elements easily also i'd be pretty happy..
- testing-library 'principles' here: https://testing-library.com/docs/guiding-principles - basically trying to help you test the the web page rather than the implementation which is where the false positives come from (apparently!)


.eslintrc.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
     es2021: true,
   },
   extends: [
+    'plugin:jest-dom/recommended',
```
joshxg, line 7: Just a few preferences in the tests.. https://github.com/testing-library/eslint-plugin-jest-dom#recommended-configuration

__mocks__/zustand.ts
```diff
@@ -0,0 +1,24 @@
...
+  const initialState = store.getState();
+  stores.add(() => act(() => store.setState(initialState, true)));
+
+  return store;
+};
+
+afterEach(() => {
+  stores.forEach(resetFn => resetFn());
+});
+
```
joshxg, line 23: ensuring if you continue to use a zustand store the state is reset after each test

jest-setup.ts
```diff
@@ -0,0 +1,21 @@
...
+});
+
+afterAll(() => {
+  (global.Storage.prototype.setItem as jest.Mock).mockReset();
+  (global.Storage.prototype.getItem as jest.Mock).mockReset();
+});
+
+beforeEach(() => {
+  mockStorage = {};
+});
```
joshxg, line 21: ensuring the store is reset after tests

jest.config.js
```diff
@@ -0,0 +1,16 @@
...
+module.exports = {
+  preset: 'ts-jest',
+  testEnvironment: 'jsdom',
+  setupFilesAfterEnv: ['<rootDir>/jest-setup.ts'],
+  globals: {
+    'ts-jest': {
+      // Without this, there is a warning about imports which I can't remove. Everything seems
+      // to be working fine, though! https://github.com/kulshekhar/ts-jest/issues/748
+      diagnostics: {
+        ignoreCodes: [151001],
```
joshxg, line 12: 😭

package.json
```diff
@@ -17,14 +18,21 @@
...
     "@typescript-eslint/eslint-plugin": "^4.28.4",
     "@typescript-eslint/parser": "^4.28.4",
     "eslint": "^7.31.0",
     "eslint-config-google": "^0.14.0",
+    "eslint-plugin-jest-dom": "^3.9.0",
     "eslint-plugin-react": "^7.24.0",
+    "jest": "^27.0.6",
     "lerna": "^3.22.1",
     "swc-loader": "^0.1.14",
+    "ts-jest": "^27.0.4",
```
joshxg, line 35: I put all this in the root, and this means I can run (on my machine!!!) `yarn test` anywhere and all the tests run.. maybe we want more finer grained control to just run tests for a particular package, but there are jest cli options for that type of thing anyway 🤔

packages/common/src/intl/IntlTestProvider.tsx
```diff
@@ -0,0 +1,11 @@
+import React, { FC } from 'react';
+import { IntlProvider as ReactIntlProvider } from 'react-intl';
+import en from './locales/en.json';
+
+export const IntlTestProvider: FC = ({ children }) => {
+  return (
+    <ReactIntlProvider locale="en" messages={en}>
+      {children}
+    </ReactIntlProvider>
+  );
```
joshxg, line 10: The testing-library doesn't like it when you change state internally 🤔  (The real intl provider does this to lazy load the translations).

it gives you functionality to, for example, press a button or input text etc. The preference is that these state changing invocations are done within a method `act` - which ensures that the component is fully rendered and everything is all good before trying to do state changes.

So, if you change internal state outside of an `act`, the library gives a warning. I thought that this component was OK and that because we require quite a few different providers for components, it might be best to make a `TestProvider` which can actually pass a lot more things initially - we could then pass through props the `messages` to test different localisations etc.

packages/common/src/intl/locales/en.json
```diff
@@ -15,5 +15,6 @@
   "app.transactions": "Transactions",
   "app.welcome": "Hello, {name}!",
   "breadcrumb.item": "#{id}",
-  "button.ok": "ok"
-}
+  "button.ok": "ok",
+  "button.open-the-menu": "Open the menu"
```
joshxg, line 19: I realise i'm unsure of the convention with the keys here

packages/common/src/intl/locales/pt.json
```diff
@@ -15,5 +15,6 @@
   "app.transactions": "Transactions",
   "app.welcome": "Hello, {name}!",
   "breadcrumb.item": "#{id}",
-  "button.ok": "ok"
-}
+  "button.ok": "ok",
+  "button.open-the-menu": "Abra o menu"
```
joshxg, line 19: I also can see having to put the key in all translation files to get tedious quickly?

packages/host/src/AppDrawer.test.tsx
```diff
@@ -0,0 +1,79 @@
...
+      <AppThemeProvider>
+        <IntlTestProvider>
+          <BrowserRouter>
+            <AppDrawer />
+          </BrowserRouter>
+        </IntlTestProvider>
+      </AppThemeProvider>
+    );
+
+    const button = screen.getByRole('button', { name: /Open the menu/i });
```
joshxg, line 21: it is certainly difficult to create tests which aren't fragile, that's for sure. If we change this label the test breaks which isn't great!

packages/host/src/AppDrawer.tsx
```diff
@@ -21,7 +21,11 @@ import {
 } from '@openmsupply-client/common';
 import clsx from 'clsx';
 
-const CustomersNav = React.lazy(() => import('customers/Nav'));
+const CustomersNav = React.lazy(() =>
+  process.env['NODE_ENV'] === 'test'
+    ? import('../../customers/src/Nav')
+    : import('customers/Nav')
```
joshxg, line 27: Hm. This seemed easiest but is probably not the best way to go..

packages/host/src/AppDrawer.tsx
```diff
@@ -130,10 +134,13 @@ const Menu: React.FC<MenuProps> = ({ classes }) => {
 const AppDrawer: React.FC = () => {
   const classes = useStyles();
   const drawer = useDrawer();
+  const t = useTranslation();
 
   return (
     <Drawer
+      data-testid="drawer"
```
joshxg, line 141: argument for `data-testid` is that you should generally try not to, but if you can't access the element then if you use a class name, or css id then you're using implementation details, so just use `data-testid` 🤷

packages/host/src/AppDrawer.tsx
```diff
@@ -143,7 +150,10 @@ const AppDrawer: React.FC = () => {
       open={drawer.isOpen}
     >
       <div className={classes.toolbarIcon}>
-        <IconButton onClick={drawer.toggle}>
+        <IconButton
+          aria-label={t('button.open-the-menu')}
```
joshxg, line 154: I learnt a bit about aria stuff today!!

jest.config.js
mark-prins: oh, typescript 🙄

package.json
mark-prins: yeah, agree. certainly at this stage having them in the root is best

packages/common/src/intl/locales/en.json
mark-prins: I should write up where I've gone with it.. I don't really know what's best and haven't given it enough thought yet.

packages/common/src/intl/locales/pt.json
mark-prins: I'm already over it.
Started with the approach similar to mobile, where `en` is the base and you just add in parts that you have of other languages. Missing translations fall back to the base language. had trouble getting that to fly with lazy loading and decided that maybe adding the key to all translation files was ok.

packages/host/src/AppDrawer.tsx
mark-prins: what about lifting it out to a helper method. still just as clumsy, but the code looks nicer! ( and if we think of another way to make it work, then easy to change )
```
const lazyLoad = (path: string) => React.lazy(() =>
  process.env['NODE_ENV'] === 'test'
    ? import(`../../${path}`)
    : import(path);

const CustomersNav = lazyLoad('customers/Nav');
```

packages/host/src/AppDrawer.tsx
mark-prins: nice! I've always skipped over aria

mark-prins (approved):
wowsers! that's awesome
agree with the comment about fragility of tests, particularly UI related tests which can be a real pain. and the other aspect of needing to put test hooks into the code which isn't great - what you've got is not intrusive though and great to have some test examples

packages/host/src/AppDrawer.tsx
joshxg: Nice. I change this a little in my next PR so will address it soon!

joshxg:
pretty minor changes to the tests to have the pass again which i'm happy with, ha 
6cbf660

PR #57, opened 2021-08-17
#51 storybook

### LINKED ISSUE #51: Add storybook
Before creating so many packages, would it be useful to have story book set up?

I know when working in a big code base like Grafana it was incredibly useful.

I know this will take some time to setup. Will take more time if we want to set it up.. later?

@mark-prins think this is useful?

joshxg:
Fixes #51 

image

cooooool

- I left the default `stories` folder for now because it's easier than looking at the docs..
- if we have a `test` and a `story` and an actual component file for each component we might need a better structure... unsure if we should just use the `stories` folder and put all the stories in there or if having a folder per component with the test and story in is nice also.. :(
- I added `react` and `react-dom` to dev deps of the root as story book needed it there.. when I added them react 17 was added.. so I just updated all of them.. mighhhht need to `rm -rf node_modules` and `yarn install`..


.storybook/main.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  stories: [
+    '../packages/**/*.stories.mdx',
+    '../packages/**/*.stories.@(js|jsx|ts|tsx)',
+  ],
```
joshxg, line 5: unsure where to actually put stories

.storybook/theme.js
```diff
@@ -0,0 +1,36 @@
...
+  barTextColor: '#f2f2f5',
+  barSelectedColor: 'black',
+  barBg: '#e95c30',
+
+  // Form colors
+  inputBg: 'blue',
+  inputBorder: 'blue',
+  inputTextColor: 'blue',
+  inputBorderRadius: 4,
+});
```
joshxg, line 36: I was just messing around it was a bit of fun. Feel free to tweak any values you want

package.json
```diff
@@ -8,14 +8,23 @@
     "compile": "lerna run --scope @openmsupply-client/* --parallel tsc",
     "prepare": "husky install",
     "pre-commit-lint": "yarn compile && yarn lint-staged",
-    "test": "jest --config ./jest.config.js"
+    "test": "jest --config ./jest.config.js",
+    "storybook": "start-storybook -p 6006",
+    "build-storybook": "build-storybook"
```
joshxg, line 13: unsure if we should use a `@openmsupply-client/ui` package with all the building blocks and just use story book for that..

But on the other hand, it is a super nice way for 'visual testing' - easily be able to switch through components and check that strings fit etc.

package.json
```diff
@@ -8,14 +8,23 @@
...
     ]
   },
   "devDependencies": {
+    "@babel/core": "^7.15.0",
+    "@storybook/addon-actions": "^6.3.7",
+    "@storybook/addon-essentials": "^6.3.7",
+    "@storybook/addon-links": "^6.3.7",
+    "@storybook/addons": "^6.3.7",
+    "@storybook/react": "^6.3.7",
+    "@storybook/theming": "^6.3.7",
```
joshxg, line 27: thats a lot!

packages/common/src/ui/components/LoadingApp.stories.tsx
```diff
@@ -0,0 +1,12 @@
...
+import { LoadingApp } from '.';
+
+const Template: ComponentStory<typeof LoadingApp> = () => <LoadingApp />;
+
+export const Primary = Template.bind({});
+
+export default {
+  title: 'Example/Loading',
+  component: LoadingApp,
+} as ComponentMeta<typeof LoadingApp>;
```
joshxg, line 12: simple example

packages/common/src/utils/testing.tsx
```diff
@@ -0,0 +1,22 @@
...
+  locale = 'en',
+}) => {
+  return (
+    <AppThemeProvider>
+      <IntlTestProvider locale={locale}>
+        <BrowserRouter>{children}</BrowserRouter>
+      </IntlTestProvider>
+    </AppThemeProvider>
+  );
+};
```
joshxg, line 22: a lot of stuff will need this so makes writing tests/stories easier.. not sure if this is a good place for it

packages/host/src/AppDrawer.tsx
```diff
@@ -22,7 +22,7 @@ import {
 import clsx from 'clsx';
 
 const CustomersNav = React.lazy(() =>
-  process.env['NODE_ENV'] === 'test'
+  process.env['NODE_ENV'] !== 'production'
```
joshxg, line 25: 🏃

stories/Button.stories.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
```
joshxg, line 1: I left all these in to reference.. can remove?

.storybook/theme.js
```diff
@@ -0,0 +1,36 @@
...
+  colorSecondary: '#e95c30',
+
+  // UI
+  appBg: '#555770',
+  appContentBg: 'white',
+  appBorderColor: 'white',
+  appBorderRadius: 4,
+
+  // Typography
+  fontBase: '"Open Sans", sans-serif',
```
mark-prins, line 16: I spent a while trying to 'fix' this and bring in a better font.. but have given up.

.storybook/theme.js
mark-prins: maybe later 😉  all good.

package.json
mark-prins: I really like the idea of a ui package but do worry that updating, bumping revisions, updating usages.. all too much hassle at the mo. maybe later?
agree though, that storybook does help with checking out components which are built of items in the ui package, and so it is good to have it at this level

packages/common/src/utils/testing.tsx
mark-prins: nice! not sure either.. seems ok to me

stories/Button.stories.tsx
mark-prins: can we comment somewhere to say that they are here for reference? rename the folder?
agree that they're good to have, just threw me at first, wondering why there were stories here, and next to components

mark-prins (approved):
badass!
am now keen to get it in and create stories for shared components and stuff like the appbar to show the options available.

.storybook/main.js
joshxg: Think you're suggesting that having the stories at the same level is no good?

joshxg:
Merging into my other pr to do the tidy ups there.

PR #58, opened 2021-08-18
responsive app drawer and styles on appbar

### LINKED ISSUE #54: Implement viewport
An initial implementation of supporting screen sizes - there are designs in zeplin to work with.

At this stage the noticeable implementation is support for having the drawers collapsed on load. I think if we implement a framework and then add changes for the drawers that will be enough for now.

mark-prins:
fixes #54 

Opens and closes the drawer based on screen size. Not much else in here really, just a small tidy and the example of adjusting components based on the media query.

### DIFF
--- a/packages/common/src/hooks/useDrawer.ts
+++ b/packages/common/src/hooks/useDrawer.ts
@@ -7,15 +7,11 @@ type DrawerController = {
   toggle: () => void;
 };
 
-const localStorageKey = '@openmsupply-client/appdrawer/open';
-
-export const useDrawer = create<DrawerController>(set => ({
-  isOpen: !!JSON.parse(localStorage.getItem(localStorageKey) ?? 'false'),
-  open: () => set(state => ({ ...state, isOpen: true })),
-  close: () => set(state => ({ ...state, isOpen: false })),
-  toggle: () => set(state => ({ ...state, isOpen: !state.isOpen })),
-}));
-
-useDrawer.subscribe(({ isOpen }) => {
-  localStorage.setItem(localStorageKey, JSON.stringify(isOpen));
+export const useDrawer = create<DrawerController>(set => {
+  return {
+    isOpen: true,
+    open: () => set(state => ({ ...state, isOpen: true })),
+    close: () => set(state => ({ ...state, isOpen: false })),
+    toggle: () => set(state => ({ ...state, isOpen: !state.isOpen })),
+  };
 });
--- a/packages/common/src/hooks/useDrawer.ts
+++ b/packages/common/src/hooks/useDrawer.ts
@@ -1,17 +1,23 @@
 import create from 'zustand';
 
+const localStorageKey = '@openmsupply-client/appdrawer/open';
+
 type DrawerController = {
   isOpen: boolean;
+  isSet: boolean;
   open: () => void;
   close: () => void;
   toggle: () => void;
 };
 
-export const useDrawer = create<DrawerController>(set => {
-  return {
-    isOpen: true,
-    open: () => set(state => ({ ...state, isOpen: true })),
-    close: () => set(state => ({ ...state, isOpen: false })),
-    toggle: () => set(state => ({ ...state, isOpen: !state.isOpen })),
-  };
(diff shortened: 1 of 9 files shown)


packages/common/src/hooks/useDrawer.ts
```diff
@@ -7,15 +7,11 @@ type DrawerController = {
...
-export const useDrawer = create<DrawerController>(set => ({
-  isOpen: !!JSON.parse(localStorage.getItem(localStorageKey) ?? 'false'),
-  open: () => set(state => ({ ...state, isOpen: true })),
-  close: () => set(state => ({ ...state, isOpen: false })),
-  toggle: () => set(state => ({ ...state, isOpen: !state.isOpen })),
-}));
-
-useDrawer.subscribe(({ isOpen }) => {
-  localStorage.setItem(localStorageKey, JSON.stringify(isOpen));
+export const useDrawer = create<DrawerController>(set => {
```
mark-prins, line 10: Could continue to keep the choice in local storage but I thought that it was more useful to have the drawer show based on screen size

packages/common/src/styles/theme.ts
```diff
@@ -41,5 +48,6 @@ const themeOptions = {
 };
 
 const theme = createTheme(themeOptions);
-
+theme.shadows[1] =
```
mark-prins, line 51: slightly iffy maybe?
all of the material shadows are much darker than the designs we have

packages/host/src/Host.tsx
```diff
@@ -60,45 +60,47 @@ const Host: FC = () => {
           <AppThemeProvider>
             <BrowserRouter>
               <Viewport>
+                <AppBar />
```
mark-prins, line 63: not necessary to move it now - I was thinking to change all of these to flex layout.
Kept this as a fixed position so that it is always on the top of the page and as such, doesn't matter where it is in the dom

packages/host/src/Host.tsx
```diff
@@ -60,45 +60,47 @@ const Host: FC = () => {
...
-                        path="messages/*"
-                        element={<Heading locale={locale}>messages</Heading>}
-                      />
-                      <Route
-                        path="*"
-                        element={<Navigate to="/dashboard" replace />}
-                      />
-                    </Routes>
-                  </React.Suspense>
+                  <Box flex={1} style={{ marginTop: 90 }}>
```
joshxg, line 66: Should this be `72`, the height of the `AppBar`?

joshxg (approved):
Nice! I am a little unsure which zepplin designs you're following, there's a lot of different ones

packages/host/src/Host.tsx
mark-prins: was looking mainly at [Framework / Table](https://app.zeplin.io/project/60e44fff6fa942168092f1ff/screen/60e498d63f93c6b91456c0b9) which is `90px` - the later screens all seem to be `90` so went with that

mark-prins:
> Could have the initial 'never seen it in local storage' value be hidden for small screen and visible for big ones, then if the user ever changes it the state is persistent? 🤷

yeah, was wondering about that - but what happens if someone hides it and then resizes the browser to be on a bigger screen? Do we keep it hidden ( like they asked ) or show it ( now that there is more room )?

the question is: "is the decision to hide based on screen size?" and we don't know that
should we support resizing? ( I think yes ) or just set an initial value when the site loads?

current thinking:
- open & close based on screen size, and watch for resizing
- if the user ever changes then the state is persistent and never change

since the open/close button is pretty obscure - for most people the responsive will happen. only those who really want it closed or open will click and therefore keep it that way.

packages/common/src/hooks/useDrawer.ts
mark-prins: Ok - it is now unresponsive :)
and slightly harder to test.

Required `localStorage.removeItem('@openmsupply-client/appdrawer/open');` in console in order to check the responsiveness after opening and closing like I have.

joshxg: Works well and I like it a lot, thanks

Issue #45, opened 2021-08-11
Typed and safer local storage interface


joshxg:
Would be good and helpful to have a more type-safe and parsing-safe local storage interface to utilize as a util in common

PR #62, opened 2021-08-19
#45 type safe localstorage

joshxg:
#45 

- getItem / setItem should be typed dependent on the key passed and the key in `keys`


.eslintrc.js
```diff
@@ -28,6 +28,7 @@ module.exports = {
   settings: { react: { version: 'detect' } },
   plugins: ['react', , '@typescript-eslint'],
   rules: {
+    camelcase: ['error', { allow: ['_ONLY_FOR_TESTING'] }],
```
joshxg, line 31: 🤷 is this a good idea? functions only to be used for testing having this suffix..?

package.json
```diff
@@ -8,7 +8,7 @@
     "compile": "lerna run --scope @openmsupply-client/* --parallel tsc",
     "prepare": "husky install",
     "pre-commit-lint": "yarn compile && yarn lint-staged",
-    "test": "jest --config ./jest.config.js",
+    "test": "jest --config ./jest.config.js --maxWorkers=50%",
```
joshxg, line 11: why are the jest tests so slow!

joshxg (commented):
Hm. I like this, but I can see a few things that are debatable!

mark-prins (approved):
very nifty. certainly like the addition of type safety on the local storage access, I think it will be useful to have local storage so worth getting usage tidy.

PR #63, opened 2021-08-20
Add app bar focusing/selecting update

### LINKED ISSUE #59: AppBar focus highlighting of nested sub menus when collapsed is incorrect
- step1 expando the drawo
image

- step2 navigate to a nested submenu
image

- step3 collapso the drawo

image

Should `Customers` be the one highlighted now?

joshxg:
Fixes #59 

- The customer nav icon should be highlighted when you're viewing a customer invoice

### DIFF
--- a/packages/common/src/ui/components/NavLink.tsx
+++ b/packages/common/src/ui/components/NavLink.tsx
@@ -4,6 +4,7 @@ import clsx from 'clsx';
 import { ListItem, ListItemText, Tooltip } from '@material-ui/core';
 import { makeStyles } from '@material-ui/core/styles';
 import { useMatch, Link } from 'react-router-dom';
+import { useDrawer } from '../../hooks/useDrawer';
 
 const useStyles = makeStyles(theme => ({
   drawerMenuItem: {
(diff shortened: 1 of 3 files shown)


packages/common/src/ui/components/NavLink.tsx
```diff
@@ -48,9 +62,9 @@ export const AppNavLink: FC<ListItemLinkProps> = props => {
 
   return (
     <li>
-      <Tooltip title={text || ''}>
+      <Tooltip disableHoverListener={drawer.isOpen} title={text || ''}>
```
mark-prins, line 65: that's probably fair enough. no need to show the tooltip on the label which has the same text!

packages/customers/src/Nav.tsx
```diff
@@ -27,22 +27,28 @@ const Nav: FC = () => {
...
-            end={true}
-            to="/customers/customer-invoice"
-            text={t('app.customer-invoice')}
-          />
-          <AppNavLink
-            end={true}
-            to="/customers/customer-requisition"
-            text={t('app.customer-requisition')}
-          />
+          {isActive && (
```
mark-prins, line 37: I'm not sure why these shouldn't be rendered now.. but find to do

packages/common/src/ui/components/NavLink.tsx
```diff
@@ -28,11 +29,24 @@ interface ListItemLinkProps {
...
 }
 
+const useSelectedNavMenuItem = (to: string, end: boolean): boolean => {
+  // This nav menu item should be selected when lower level elements
+  // are selected. For example, the route /customer-invoices/{id} should
+  // highlight the nav menu item for customer-invoices.
+  const highlightLowerLevels = !end || to.endsWith('*');
+
+  // If we need to highlight on lower levels, append a wildcard.
+  const path = highlightLowerLevels ? to : `${to}/*`;
```
mark-prins, line 39: it works, though I'm confused by the wording. There's "If we need to highlight on lower levels, append a wildcard." but we append a wildcard if `highlightLowerLevels` is false

shouldn't the comment be more like 'If we need to highlight on higher levels'?

mark-prins (approved):
it works! the logic and words confused me.. I think that's a 'me' problem.
good catch, I hadn't even noticed.

packages/customers/src/Nav.tsx
joshxg: Ugh I know! I wanted to comment but commenting within the JSX sucks.. it's to avoid a weird animation thing 🙄

packages/common/src/ui/components/NavLink.tsx
joshxg: Ugh I know!

Why this simple change was so hard for me 🙄 

All of the wording was difficult. I even tried a version where we just passed a `matchPath` in as a prop rather than doing this wildcard stuff here because I found it confusing and confusing to 'name'..

packages/customers/src/Nav.tsx
mark-prins: wondered if it was something like that - just didn't see it when testing. I wasn't as observant as you :)

PR #64, opened 2021-08-21
#43 routes to enums

### LINKED ISSUE #43: Add route enums
Need to add some enums for routes.

Think the issue of where to put these is gonna be a common one coming up.

### Defined commonly
- + No conflicts
- + Packages can more easily access them
- - Need to bump the common version every route added

### Defined independently
- The opposite of common..?
- Not sure if conflicts are really an issue.

joshxg:
Fixes #43 

OTT?

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -30,6 +30,8 @@ module.exports = {
   rules: {
     'require-jsdoc': 0,
     'react/display-name': 'off',
+    'no-unused-vars': 'off',
+    '@typescript-eslint/no-unused-vars': 'error',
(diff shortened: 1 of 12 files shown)


mark-prins (approved):
🎉  that is very pleasing! Like the approach, that's very elegant! 😄 

Had trouble initially.. as I had to manually `yarn install` in the `packages/config` to get that working. without that init, the response is `Module not found: Error: Can't resolve '@openmsupply-client/config' in '/Users/m2/GitHub/openmsupply-client/packages/customers/src'` which isn't super helpful

PR #66, opened 2021-08-23
Add basic error handling component

### LINKED ISSUE #52: Add generic error handling
Would be good to have a generic error boundary we can use, especially for lazily loaded components, just incase? So if a lazily loaded page doesn't load we can show a 'oops!' type thing?

joshxg:
Fixes #52 

I wanted to explore error handling a little bit.. if bugsnag was supporting offline logging (it doesn't) and generally what the bugsnag react plugin does and how we can extend it... I ended up with some of this so I thought i'd just commit it.. Can wrap lazily loaded components individually etc in the future..

The bugsnag plugin basically gives an errorboundary like the one written. However since bugsnag won't send logs offline, if we want to handle that, we'd need to store them in local storage (or just send it to the remote server) or something when offline so we'd need to basically not use their boundary.. 

The actual bugsnag calling can then be more appropriately used in a logging interface..?

The API Key.. well.. uhh.. we have it committed in mobile too? 😁


packages/common/src/ui/components/errors/GenericErrorFallback.tsx
```diff
@@ -0,0 +1,25 @@
+import { Box, Typography, Button } from '@material-ui/core';
+import React, { FC } from 'react';
+import { ErrorBoundaryFallbackProps } from './types';
+import UnhappyMan from '../../icons/UnhappyMan';
```
mark-prins, line 4: that name still makes me smile

packages/common/src/ui/components/errors/GenericErrorFallback.tsx
```diff
@@ -0,0 +1,25 @@
...
+    <Box
+      display="flex"
+      flexDirection="column"
+      justifyContent="center"
+      alignItems="center"
+      height="100%"
+      flex={1}
+    >
+      <UnhappyMan />
+      <Typography variant="h3">Oops! Somethings gone wrong.</Typography>
```
mark-prins, line 19: this looks better if the `Typography` component has 20px of padding on it
and a localisation for it.. please?

mark-prins: I also think it should be `Something's gone wrong` with the apostrophe as a contraction for 'has'

mark-prins (approved):
👍  great idea! thanks - like the ErrorBoundary.
had to read the docs to see that event handlers aren't caught, and also cannot get a lazy load to fail appropriately. For me the app doesn't render if there are any issues with load. perhaps I'm testing that wrong 🤷

PR #68, opened 2021-08-23
Add swc remove ts-jest

joshxg:
Down to ~7s for me. Still seems long 🤔


mark-prins (approved):
that fairly flies along compared the last time!
yes, maybe could be faster, but still.. much better. and acceptably quick I think

PR #69, opened 2021-08-23
#4 data table

### LINKED ISSUE #4: Implement a data table
When we decide which one to use.. add to project and style appropriately.

mark-prins:
Fixes #4
An initial implementation of react-table as a replacement for data-grid.
Where I've ended up is creating two table components
- one that works with 'local' data i.e. the calling component fetches data and hands that to the table, which then internally manages sorting and pagination (if required)
- and one that uses 'remote' data, in the sense that the server handles pagination and sorting and filtering

To test that out, I've extended the mock server to support sorting and pagination; though I expect the format will change when this is settled. I've used the `{first: number, offset: number}` format which is suggested for graphQL and used in universal codes. The sorting provides a field to sort by and asc/desc indication. I've used `desc` as ascending is the usual default, and so a truthy value for `desc` which defaults to false fits the normal usage.

There's a lot I'm not happy with but this works nicely, is the best version so far, and I feel like I've spent long enough kicking around options ...and don't want to spend too long polishing this turd if you think there's better alternatives.


packages/common/src/ui/icons/SortAsc.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import SvgIcon, { SvgIconProps } from '@material-ui/core/SvgIcon';
+
+export const SortAsc = (props: SvgIconProps): JSX.Element => {
```
mark-prins, line 4: these weren't svgs in zeplin, so I made my own which look about right to me.

packages/common/src/ui/layout/tables/RemoteDataTable.tsx
```diff
@@ -0,0 +1,179 @@
...
+    cursor: 'pointer',
+  },
+  headerCell: {
+    ...theme.typography.th,
+  },
+  loadingIndicator: { marginLeft: 'auto', marginRight: 'auto' },
+  loadingIndicatorContainer: { display: 'flex', marginTop: 50 },
+}));
+
+interface TableProps<T extends Record<string, unknown>> {
```
mark-prins, line 59: oh, typescript... what a glorious pain you can be sometimes.
I really like ts and the ability here to define a table of a certain data type is nice - but the hoops that you have to jump through some times can end up a little clumsy. Here I'm just trying to add a generic to the column and data, and it insists that I extend `object` and then the linter doesn't like the use of `object`. And cannot use `React.FC` with generics, which is why the prop type and return type of `JSX.Element` instead

packages/common/src/ui/layout/tables/RemoteDataTable.tsx
```diff
@@ -0,0 +1,179 @@
...
+  return !!column.isSortedDesc ? <SortDesc /> : <SortAsc />;
+};
+
+export const RemoteDataTable = <T extends Record<string, unknown>>({
+  columns,
+  onFetchData,
+  onRowClick,
+}: TableProps<T> & { children?: ReactNode }): JSX.Element => {
+  const classes = useStyles();
+  const [pageSize, setPageSize] = useState(DEFAULT_PAGE_SIZE);
```
mark-prins, line 79: I'm not happy with a lot of `useState` calls here but like it slightly better than mutating an object

packages/common/src/ui/layout/tables/RemoteDataTable.tsx
```diff
@@ -0,0 +1,179 @@
...
+      );
+      setIsLoading(false);
+    });
+  }, [pageSize, pageIndex]);
+
+  useEffect(() => setIsLoading(!tableData), [tableData]);
+
+  return isLoading ? (
+    <Box className={classes.loadingIndicatorContainer}>
+      <CircularProgress className={classes.loadingIndicator} />
```
mark-prins, line 117: probs not needed; not sure yet how slow the queries will be. Material claims to debounce if under 100ms so I haven't unbounced, but maybe that would be good to add in too?

packages/common/src/ui/layout/tables/react-table-config.d.ts
```diff
@@ -0,0 +1,26 @@
+import {
```
mark-prins, line 1: another TS issue. Had to add this to get the linter to smile. And then had to duplicate in `host` to get the pre-commit linting hook to allow me to commit. Need to see if I can get around that

packages/host/src/Host.tsx
```diff
@@ -19,6 +20,11 @@ import AppDrawer from './AppDrawer';
 import AppBar from './AppBar';
 import Viewport from './Viewport';
 
+const Content = styled(Box)({
```
mark-prins, line 23: the table was bigger than the page, but you couldn't scroll down and see the rest of it! added the overflow to allow the scroll within this container

packages/transactions/package.json
```diff
@@ -35,4 +35,4 @@
     "react-dom": "^17.0.2",
     "react-router-dom": "^6.0.0-beta.0"
   }
-}
\ No newline at end of file
+}
```
mark-prins, line 38: haven't removed data-grid yet.. will wait for feedback first

packages/transactions/src/TransactionService.tsx
```diff
@@ -54,71 +62,52 @@ const Transaction: FC = () => {
...
-}
-  `
-  );
+const Transactions: FC = () => {
+  const queryProps = { first: 10, offset: 0, sort: undefined, desc: false };
+  const listQuery = async () => {
+    const { first, offset, sort, desc } = queryProps;
+    const sortParameters = sort ? `, sort: ${sort}, desc: ${!!desc}` : '';
+    const { transactions } = await request(
+      'http://localhost:4000',
```
mark-prins, line 71: I thought about env vars at some point. haven't done that yet.

joshxg:
@mark-prins 

Couple Q's while i'm looking: 
- Sorting: Is per page sorting the goal? Was expecting per data set sorting via the server but I don't have much experience with best practise with pagination
- Data/state: The data is returned from the `useQuery` hook and could be passed down and the prop passed to `useTable`, as well as the pagecount calculated on the fly (?), what are we getting from putting it (the rows) in a `useState`?
- Data set in memory: My initial expectation was that the data was to be appended when we fetch more pages rather than having a full new data set. I can see that could eventually bring to much into memory, is that why we don't do that?

mark-prins:
> Couple Q's while i'm looking:

- Sorting: I was planning per data set on the server; just gave up before implementing that part. The PR was dragging on a little.
- Data/state: I would prefer not to have it in state. there were some troubles with managing the loading state within the table component, while the query hook did the fetch outside it. I think that process needs refining. It also bothers me that the `useQuery` is disabled in order to prevent the initial data request being made outside of the table component. I don't think that's very intuitive, nor how the hook should be used. The `refetch` is fine I think, just the initial fetch that is odd.
- Not sure what this gives, other than the ability for users to go from a page in the result set to the previous page without a server query; and it requires a bit more work to track locally which pages have been fetched, and then invalidate if rowsPerPage or sort or filter options are changed. Happy to bung that in if you think it's useful

mark-prins:
Haven't done the tests yet either. Will make more issues for the remainder of things

joshxg (approved):
> Happy to bung that in if you think it's useful

Nope, just trying to understand a couple things!

It's awesome and I can see it's going to work well. I'm not sure I see why we need separate remote and local data tables, though.

Lets get it in!

PR #76, opened 2021-08-28
#75 unused webpack packages

### LINKED ISSUE #75: Remove unused webpack configs
After #73 we have a few unused webpack configs we can remove

joshxg:
Fixes #75

Jeeeez I need a shower.

Sorry this has taken me far too long. There's really a lot to learn in js bundling/modules, ts compiling, working within yarn workspaces, lerna 😭 

After removing the webpack configs we end up with a bunch of libraries and a single web app that lazy loads them.

### Problem 1

There is still a problem that the `common` package is still not being tree shaken which results in the whole thing being put into a single bundle, which means that it's bundle is large and not spread out into other bundles, and that it is larger than it should be as is contains a bunch of code we aren't using.. this results in a 200kb gzipped bundle - not the end of the world, but still a lot larger than it should be. The main problem being needing to download that whole 200kb before interaction.

<details>
<summary>bundle analyzer</summary>
<image src="https://user-images.githubusercontent.com/35858975/131203151-bcf18c95-1973-4917-b458-9f46f8affdb4.png"/>
</details>

As mentioned we can get rid of this by being super explicit in the imports.. i.e. from `@openmsupply-client/common/ui/components/Typography` - that is super brittle, but we can mitigate that with github actions checking bundle sizes.

I *think* we need to have `common` be built as 'es modules', which brings with it tree-shaking, but I am not 100% sure on how to do that right now.

https://www.npmjs.com/package/babel-plugin-import - we can also use this plugin which we should be able to make imports like `import { Typography } from '@openmsupply-client/common'` into `import Typography from '@openmsupply-client/common/ui/components/Typography` .. I think 🤔  - as shown in the mui docs, option 2 - https://material-ui.com/guides/minimizing-bundle-size/#option-2 - I think this is needed because we are bundling a package which is exporting a package 🙄 

### Problem 2

I seem to be getting  correct intellisense for non-material ui components in `host`.. for example if I try to use `LocalStorage`, I get intellisense 👍  If I try for `Typography`, it doesn't work. Unless! I import and export `Typography` at the top level index. Annoying.

### Problem 3

I'm not sure how to do better imports than this for the lazy loads which require a default export: `'@openmsupply-client/customers/src/CustomerContainer` - that's super ugly :(

### The Good

- `yarn start` is super quick 🤣 
- I can edit any of the react in any of the packages and hot module replacement is working


joshxg:
`lerna clean` and `rm -rf node_modules`, then `lerna bootstrap` 

Then check if `yarn compile` and `yarn build` work?

Also hot module replacement

tsconfig.json
```diff
@@ -1,4 +1,5 @@
 {
+  "files": ["./types/react-table-config.d.ts"],
```
joshxg, line 2: I was having an issue with `dashboard` not compiling because it didn't have this. Seems it's required in every package which uses the table. So, made this folder? 🤷

package.json
```diff
@@ -36,18 +35,14 @@
...
     "eslint-plugin-jest-dom": "^3.9.0",
     "eslint-plugin-react": "^7.24.0",
     "jest": "^27.0.6",
     "lerna": "^3.22.1",
     "react": "^17.0.2",
-    "react-dom": "^17.0.2",
-    "swc-loader": "^0.1.14",
-    "webpack-bundle-analyzer": "^4.4.2",
-    "webpack-dev-server": "4.0.0-beta.3"
+    "react-dom": "^17.0.2"
```
joshxg, line 45: just removed some deps that didn't seem to need to be here (I did this a bit and i'm still learning about all this building, so very might be wrong in many cases...)

packages/common/package.json
```diff
@@ -11,18 +11,13 @@
...
   "dependencies": {
     "@material-ui/core": "^4.9.10",
-    "@material-ui/data-grid": "^4.0.0-alpha.33",
-    "@material-ui/icons": "^4.9.1",
     "@openmsupply-client/config": "^0.0.0",
-    "@reduxjs/toolkit": "^1.6.0",
     "graphql": "^15.5.1",
     "graphql-request": "^3.4.0",
     "history": "^5.0.0",
     "react-query": "^3.19.1",
```
joshxg, line 18: not using this redux stuff

packages/common/src/utils/testing.tsx
```diff
@@ -1,5 +1,5 @@
 import React, { FC } from 'react';
-import { AppThemeProvider } from '@openmsupply-client/common';
+import { AppThemeProvider } from '../styles';
```
joshxg, line 2: was giving me compiler issues

packages/common/tsconfig.json
```diff
@@ -1,3 +1,6 @@
 {
-  "extends": "../../tsconfig.json"
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "noEmit": true
```
joshxg, line 5: Since these are just libraries, having them built and bundled by the only app we have, `host`? Unless we are releasing these as separate packages (which we very well might do one day) no point in emiting anything.. I had hoped it would speed up tsc but it doesn't seem to

packages/host/src/LanguageMenu.tsx
```diff
@@ -43,9 +42,7 @@ export const LanguageMenu: React.FC = () => {
 
   return (
     <div>
-      <IconButton onClick={handleClick}>
-        <TranslateIcon />
```
joshxg, line 47: Where did this go 😱

packages/host/webpack.config.js
```diff
@@ -1,17 +1,22 @@
+const ReactRefreshWebpackPlugin = require('@pmmmwh/react-refresh-webpack-plugin');
+const webpack = require('webpack');
+const HotModuleReplacementPlugin = webpack.HotModuleReplacementPlugin;
 const HtmlWebpackPlugin = require('html-webpack-plugin');
-const ModuleFederationPlugin =
-  require('webpack').container.ModuleFederationPlugin;
+// const ModuleFederationPlugin = webpack.container.ModuleFederationPlugin;
```
joshxg, line 5: I commented this out.. it's likely we'll still need to use federation for plugins/later..

packages/common/src/ui/icons/index.ts
```diff
@@ -14,4 +13,3 @@ export { SortDesc } from './SortDesc';
 export { Stock } from './Stock';
```
mark-prins, line 13: I feel some conflicts coming. happy to sort those out.

packages/host/package.json
```diff
@@ -4,17 +4,19 @@
   "main": "./src/index.ts",
   "private": true,
   "devDependencies": {
+    "@pmmmwh/react-refresh-webpack-plugin": "^0.5.0-beta.8",
```
mark-prins, line 7: this failed to install with yarn and causes issues. hoping for a release soon, so we can bump version.
For now, you could up it to `^0.5.0-rc.5`

packages/host/src/LanguageMenu.tsx
mark-prins: that's odd & another good conflict coming up with this one :)

packages/transactions/tsconfig.json
```diff
@@ -1,5 +1,7 @@
 {
   "extends": "../../tsconfig.json",
-  "outDir": "./dist",
-  "rootDir": "./src"
+  "include": ["./src"],
+  "compilerOptions": {
+    "noEmit": true
```
mark-prins, line 5: `noEmit: true` is in the base config - does it need to be in all of the app configs too?

tsconfig.json
mark-prins: yeah, it is required - this is a much nicer solution thanks.

mark-prins (approved):
Had some issues after cleaning due to that unreleased version of react-refresh webpack plugin.
That aside - hot reloading is working and the build is much faster.. go for it!!

PR #77, opened 2021-08-28
#48 appbar buttons

### LINKED ISSUE #48: AppBar per page buttons
Generic solution for these buttons:

image

- Seems these buttons would be a part of the `AppBar` component but will be page specific. 
- So the implementation of the `onClick` needs to be determined by the page but the `AppBar` renders the components?

mark-prins:
Fixes #48 

adds some context-dependent buttons to the AppBar
Exported a general button style as the imaginatively named `Button` and used that in the customer invoices list.

Went a little off road to add in notifications in order to connect an action to the buttons.


example-stories-not-used/Header.tsx
```diff
@@ -4,17 +4,27 @@ import { Button } from './Button';
 import './header.css';
 
 interface HeaderProps {
-  user?: {};
+  user?: Record<string, unknown>;
```
mark-prins, line 7: made the mistake of looking at the file and it complained, so I changed it. oops.

packages/common/src/hooks/useHostContext.ts
```diff
@@ -1,14 +1,17 @@
+import { createRef } from 'react';
 import create from 'zustand';
 import { SupportedLocales } from '../intl/intlHelpers';
 
 type HostContext = {
+  appBarButtonsRef: React.MutableRefObject<null>;
```
mark-prins, line 6: I felt ok about stuffing a ref into the store: you may have a better idea

packages/common/src/hooks/useNotification.ts
```diff
@@ -0,0 +1,31 @@
+import { useSnackbar } from 'notistack';
+
+interface NotificationHook {
+  error: (message: string) => () => void;
+  info: (message: string) => () => void;
+  success: (message: string) => () => void;
+  warning: (message: string) => () => void;
+}
+
+export const useNotification = (): NotificationHook => {
```
mark-prins, line 10: just simplifying the usage - we may want the message removal and the closing of the snackbar.. for now this is enough

packages/transactions/src/TransactionService.tsx
```diff
@@ -101,13 +114,42 @@ const Transactions: FC = () => {
...
+        <>
+          <Button
+            startIcon={<Book />}
+            onClick={() =>
+              (window.location.href = 'https://docs.msupply.foundation')
+            }
+          >
+            {t('button.docs')}
+          </Button>
+          <Button
```
mark-prins, line 128: as this point I wondered about creating a simpler to use `Button` component: just pass in the LocaleKey and the icon and the onClick handler 🤷‍♂️

packages/transactions/src/TransactionService.tsx
```diff
@@ -101,13 +114,42 @@ const Transactions: FC = () => {
...
   return (
-    <RemoteDataTable<Transaction>
-      columns={columns}
-      onFetchData={fetchData}
-      onRowClick={row => {
-        navigate(`/customers/customer-invoice/${row.id}`);
-      }}
-    />
+    <>
+      <Portal container={appBarButtonsRef.current}>
```
joshxg, line 118: Getting cut off when we shrink the appbar

image

Added a flex on the div and then the language button goes funky also (but maybe we are moving the language button anyway? 🤔 )

Maybe in the future something like the container takes an array of buttons and it can have some sort of mechanism to use a drop down if there's too many 🤷

packages/common/src/ui/components/Button.tsx
```diff
@@ -0,0 +1,23 @@
...
+  color: theme.palette.primary.main,
+  fontWeight: 'bold',
+  height: 40,
+  marginRight: 10,
+  minWidth: 115,
+  textTransform: 'none',
+  boxShadow: theme.shadows[1],
+}));
+
+const Button: React.FC<ButtonProps> = props => {
```
joshxg, line 17: Wondering if we have a wide range of buttons if the label should be more specific?

I am thinking of separate buttons for having an icon and not (rather than props) and other things to avoid bloated components with a bunch of props and conditions defining behaviour

packages/common/src/ui/components/Button.tsx
```diff
@@ -0,0 +1,23 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const { variant, ...otherProps } = props;
+  return <StyledButton variant="contained" {...otherProps} />;
```
joshxg, line 20: Can do `{...otherProps} variant="contained"` to always override variant

packages/transactions/src/TransactionService.tsx
joshxg: Ah yup.. good idea

Maybe we could think about integrating muis `Typography` components and  FormatJS 🤷 

The extraction in formatjs seems nice also

joshxg (approved):
Sooo coool 😀

packages/transactions/src/TransactionService.tsx
mark-prins: bother! didn't check that. thanks for noticing
Not sure what to do with the language button. It's not in the design, and I've just bunged it in for now

I wondered about passing in an array of buttons: in Manu's mind, we simply limit the number that of buttons we use up there, and that may work for now at least

Fixed up the size and reworked the language button in 8b7c9e8

packages/common/src/ui/components/Button.tsx
mark-prins: agree - and partly held off because I didn't want to name the buttons 🙃

packages/common/src/ui/components/Button.tsx
mark-prins: that would result in a duplicated prop which I think doesn't matter, but the linter will complain about. Have changed anyway, thanks

packages/common/src/hooks/useNotification.ts
mark-prins: Sorry, don't follow what you mean. Is that an enum of messages or locale message keys?

PR #80, opened 2021-08-30
#61 rtl

### LINKED ISSUE #61: Make RTL work
https://material-ui.com/guides/right-to-left/

Make it work?

joshxg:
Fixes #61 

- Technically makes rtl work, but it doesn't
- Did the instructions as per https://next.material-ui.com/guides/right-to-left/
- There are extra steps needed if you're using `makeStyles`/`jss` - I ignored these in the hope we consolidate on `styled` and it would then be a waste of time.
- So the styles aren't all being flipped. Can test by removing `StyledEngineProvider` in `ThemeProvider`, and then all the `margin-left` get flipped to `margin-right` etc and works well

image
image


packages/common/package.json
```diff
@@ -21,6 +21,8 @@
     "react-query": "^3.19.1",
     "react-router-dom": "^6.0.0-beta.0",
     "react-table": "^7.7.0",
+    "stylis": "^4.0.10",
+    "stylis-plugin-rtl": "^2.1.0",
```
joshxg, line 25: these are tiny but still makes me sad 🤔

packages/common/src/hooks/useHostContext.ts
```diff
@@ -1,18 +1,24 @@
...
 
 useHostContext.subscribe(({ locale }) => {
-  localStorage.setItem(localStorageKey, locale);
+  localStorage.setItem('/localisation/locale', locale);
+});
+
+LocalStorage.addListener<SupportedLocales>((key, value) => {
+  if (key === '/localisation/locale') {
+    useHostContext.setState(state => ({ ...state, locale: value }));
+  }
```
joshxg, line 23: was wondering if this is more just a 'locale' store now rather than the 'host store' but thought to wait to see if we add anymore stuff into it

packages/common/src/intl/IntlTestProvider.tsx
```diff
@@ -4,11 +4,13 @@ import { LocaleMessages, SupportedLocales } from './intlHelpers';
 import en from './locales/en.json';
 import fr from './locales/fr.json';
 import pt from './locales/pt.json';
+import ab from './locales/ab.json';
 
 const locales: Record<SupportedLocales, LocaleMessages> = {
   en,
   fr,
   pt,
+  ab,
```
joshxg, line 13: `ab` is the code for arabic and arabic is rtl, right?

packages/common/src/intl/locales/ab.json
```diff
@@ -0,0 +1,20 @@
...
+    "app.reports": "التقارير",
+    "app.stock": "الأوراق المالية",
+    "app.suppliers": "الموردين",
+    "app.sync": "تزامن",
+    "app.tools": "أدوات",
+    "app.transactions": "المعاملات",
+    "app.welcome": "أهلا،",
+    "breadcrumb.item": "#{id}",
+    "button.ok": "نعم",
+    "button.open-the-menu": "افتح القائمة"
```
joshxg, line 19: I just copy pasted some stuff from google

packages/common/src/styles/RTLProvider.tsx
```diff
@@ -0,0 +1,8 @@
+import React, { FC } from 'react';
+import { useRtl } from '../intl/intlHelpers';
+
+export const RTLProvider: FC = props => {
+  const isRtl = useRtl();
+
+  return <div style={{ direction: isRtl ? 'rtl' : 'ltr' }} {...props} />;
+};
```
joshxg, line 8: Maybe should be using a prop rather than the hook, as it couples the rtl implementation to how we're handling locales in this app pretty tightly

packages/common/src/styles/useAppTheme.ts
```diff
@@ -0,0 +1,22 @@
...
+    setTheme(newTheme);
+  };
+
+  useEffect(() => {
+    const newDirection = isRtl ? 'rtl' : 'ltr';
+    setDirection(newDirection);
+  }, [isRtl]);
+
+  return currentTheme;
+};
```
joshxg, line 22: Maybe this should be in the `host` package, given it's pretty app specific - dependent so heavily on the host context..

packages/host/src/Viewport.tsx
```diff
@@ -1,35 +1,28 @@
...
+            flexDirection: 'column',
+          },
+          html: { position: 'fixed' },
+          'html, body': {
+            height: '100%',
+            width: '100%',
+          },
+        }}
+        {...props}
+      />
```
joshxg, line 25: I had made some other changes and tried a different way, but I didn't like it so I reverted it. That way had used this `GlobalStyles` instead of `makeStyles` so I left it like that

packages/common/package.json
mark-prins: hm. seems necessary though..

packages/common/src/hooks/useHostContext.ts
mark-prins: good thing the appbarbuttoncontainerref was added then!

packages/common/src/intl/IntlTestProvider.tsx
mark-prins: I thought it was `ar`
Closes to a definitive list I can find is the [unicode cldr](https://github.com/unicode-org/cldr-json/tree/master/cldr-json/cldr-localenames-full)

packages/common/src/intl/locales/ab.json
mark-prins: 😉  It's what I would have done.
Still would like to implement a fallback in some way, so that we don't have to populate the whole list.. later though.

packages/common/src/styles/RTLProvider.tsx
mark-prins: yeah, just wasn't sure how a prop would help. still tightly bound

packages/common/src/styles/useAppTheme.ts
mark-prins: yes, I think so

packages/host/src/Viewport.tsx
```diff
@@ -1,35 +1,28 @@
...
-      '#root': {
-        height: '100%',
-        display: 'flex',
-        flexDirection: 'column',
-      },
-    },
-  }),
-  { name: 'ViewportGlobals' }
-);
+import { GlobalStyles, CssBaseline } from '@openmsupply-client/common';
```
mark-prins, line 2: I don't have a `GlobalStyles`

packages/host/src/Viewport.tsx
mark-prins: can it be `styled` rather than adding the styles prop?

mark-prins (changes requested):
It works - but I don't have a `GlobalStyles` exported. Has that been missed out?

packages/common/src/ui/components/index.ts
```diff
@@ -10,6 +11,7 @@ export * from './LoadingApp';
 export * from './RemoteComponent';
 export { AppNavLink } from './NavLink';
 export {
+  GlobalStyles,
```
joshxg, line 14: @mark-prins `GlobalStyles` is exported here?

packages/host/src/Viewport.tsx
joshxg: I tried, but it can't be. `styled` makes a `className` and applies it to the component, where as this seems to need this object passed through put in a var instead.

joshxg:
@mark-prins think we're good now. Storybook still has a bit of an issue but i'll do another issue for that so we can get this in?

packages/common/src/ui/components/index.ts
mark-prins: weird. I did a global search for Global styles and it only found the usage in Viewport.
Fine now though 🤷

mark-prins (approved):

PR #111, opened 2021-09-02
#81 remove make styles 2

### LINKED ISSUE #81: Remove uses of makeStyles
If we remove the uses of `makeStyles` we can get rid of `@material-ui/styles` package (I think)

Requires changing the styles to use `styled` for:
- `LoadingApp` (Not a great name for this one?)
- `NavLink` (Also not a great name..?)
- `AppBar`
- `AppDrawer`
- `DashboardService` (Can just remove this, it's placeholder stuff)
- `RemoteDataTable`

Can also try and more flexing? 

Probably should be a couple of issues

joshxg:
Fixes #81 

- Phew.
- Removed the last `makeStyles` which meant I could remove `@material-ui/styles` and trim the bundle a bit, RTL works, storybook working all good
- Had an issue where the RTL plugin had some weird behaviour when changing the direction during runtime. Took me awhile to figure out :( In the end I just did a refresh.. You can see the weird not correcting styles correctly behaviour when you change to arabic and before the refresh


packages/common/src/ui/layout/tables/RemoteDataTable.tsx
```diff
@@ -131,19 +115,29 @@ export const RemoteDataTable = <T extends Record<string, unknown>>({
   }, [sortBy]);
 
   return isLoading ? (
-    <Box className={classes.loadingIndicatorContainer}>
-      <CircularProgress className={classes.loadingIndicator} />
+    <Box
```
joshxg, line 118: trying to work with what this says https://next.material-ui.com/system/basics/

This is basically inline styles.

... I kinda.. like it....? I think?

joshxg:
Have noticed the snackbar toast things are going craaazy when in rtl.

Working on a new issue so will either look at that when I revisit this after a review or in a new issue

packages/dashboard/src/DashboardService.tsx
```diff
@@ -25,40 +21,33 @@ const SalesTodayWidget = () => (
...
-          </Grid>
-          <Grid item xs={12} md={4} lg={3}>
-            <SalesDepositsWidget />
-          </Grid>
-          <Grid item xs={12}>
-            <RecentInvoiceWidget />
-          </Grid>
+const Dashboard: React.FC = () => (
+  <Content>
+    <StyledContainer maxWidth="lg">
```
mark-prins, line 37: smooth.. that's much nicer

packages/host/src/LanguageMenu.tsx
```diff
@@ -7,20 +7,23 @@ import {
...
 export const LanguageMenu: React.FC = () => {
+  const navigate = useNavigate();
   const [anchorEl, setAnchorEl] = React.useState<null | HTMLElement>(null);
   const handleClick = (event: React.MouseEvent<HTMLButtonElement>) => {
     setAnchorEl(event.currentTarget);
   };
 
   const handleClose = () => {
     setAnchorEl(null);
+    navigate(0);
```
mark-prins, line 26: required for the RTL / LTR swapping?

mark-prins (approved):
🎉 
works really nicely and is much easier to read. thanks for unpicking all of that, sets things up nicely for using material styles well

packages/host/src/LanguageMenu.tsx
joshxg: Yeah, for a couple of the things acting super weird when switching to rtl at runtime that I spent way to long trying to fix. This is a hack that I landed on. Can take it away and check the styles which aren't switched.. just a couple and super annoying 😭

mark-prins: while it is neater without the reload - in practice I'm not expecting people to regularly switch language and the reload is pretty quick (at this stage) anyway. so it's all good - still quicker than editing a user, logging out and logging in again :wink:

PR #114, opened 2021-09-04
#112 New outbound

### LINKED ISSUE #112: New outbound shipment
- Add a button on the outbound shipment page for creating a new outbound shipment
- Send a request to the server to create the shipment record which is then returned with details filled - i.e. status, dates
    - We will need 'create `OS`' style method in the backend which we'll basically need to keep in sync with the frontend otherwise.
- Create the outbound in memory and set the query cache to be equal to the in memory outbound.
- Both of these options would be better if the id was created client side 🤔 
- unsure which is better. Think the first

joshxg:
Fixes #112 

- Creates a new invoice in memory and navigates to `/new`
- When saved, changes the url to `/{id}`
- Shifted around the `transaction` package - was getting a bit annoying to navigate
- made changes to `useDocumentDraft` to handle the `new` thing
- Added a zustand store mixing in with `useDocumentDraft`, which means that we can use `useOutboundDraft` to use the same draft from any component and we'll be using the same draft - i.e in the RHS Drawer, rather than passing the draft through props or a context, we can just `useOutboundDraft` everywhere, if needed - I think that's pretty cool


packages/common/src/index.ts
```diff
@@ -8,3 +10,6 @@ export * from './hooks';
 export * from './intl';
 export * from './styles';
 export * from './localStorage';
+export * from './types';
+
+export { zustand, UseStore };
```
joshxg, line 15: will likely move the hooks back into the common later, just haven't yet

packages/common/src/types/index.ts
```diff
@@ -0,0 +1,7 @@
+export type Transaction = {
+  id?: string;
+  customer: string;
+  supplier: string;
+  total: string;
+  date: string;
+};
```
joshxg, line 7: Just starting the `common/types` - likely will need two types one for the list view and one for the detail, so we don't have so many optionals

packages/common/src/ui/icons/PlusCircle.tsx
```diff
@@ -0,0 +1,8 @@
+import SvgIcon, { SvgIconProps } from '@material-ui/core/SvgIcon';
+import * as React from 'react';
+
+export const PlusCircle: React.FC<SvgIconProps> = props => (
+  <SvgIcon viewBox="0 0 24 24" color="primary" {...props}>
+    <path d="M14.7 10c0 .552-.47 1-1.05 1h-2.1v2c0 .552-.47 1-1.05 1-.58 0-1.05-.448-1.05-1v-2h-2.1c-.58 0-1.05-.448-1.05-1s.47-1 1.05-1h2.1V7c0-.552.47-1 1.05-1 .58 0 1.05.448 1.05 1v2h2.1c.58 0 1.05.448 1.05 1m-4.2 8c-4.632 0-8.4-3.589-8.4-8s3.768-8 8.4-8c4.632 0 8.4 3.589 8.4 8s-3.768 8-8.4 8m0-18C4.7 0 0 4.477 0 10s4.7 10 10.5 10S21 15.523 21 10 16.3 0 10.5 0" />
+  </SvgIcon>
+);
```
joshxg, line 8: This is not the one from feather .. I couldn't get it to work 😢  Seems SVGIcon only wants a path, and the feather one had `<line/>` components

packages/transactions/src/OutboundShipment/DetailView.tsx
```diff
@@ -0,0 +1,87 @@
+import React, { FC } from 'react';
```
joshxg, line 1: I thought `DetailView` and `ListView` were OK names, but maybe not

packages/transactions/src/OutboundShipment/DetailView.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { useQueryClient } from 'react-query';
+import { useNavigate, useParams } from 'react-router';
+import { request, Transaction } from '@openmsupply-client/common';
+import { getMutation, getDetailQuery } from '../api';
+import { createDraftStore, useDraftDocument } from '../useDraftDocument';
+
+const queryFn = (id: string) => async (): Promise<Transaction> => {
+  const result = await request('http://localhost:4000', getDetailQuery(), {
```
joshxg, line 9: Need some common `http client` type thing to abstract this url and auth and stuff away later

packages/transactions/src/OutboundShipment/DetailView.tsx
```diff
@@ -0,0 +1,87 @@
...
+  const { upsertTransaction } = result;
+  return upsertTransaction;
+};
+
+const placeholderTransaction: Transaction = {
+  customer: '',
+  total: '',
+  date: '',
+  supplier: '',
+};
```
joshxg, line 28: Found this to be the nicest way to do the new thing.. seed with this placeholder and that becomes the `data` value returned from `useQuery`.. so when it's a `new` outbound, use this placeholder, otherwise don't pass one. Then everything else is the same as editing, kinda

packages/transactions/src/OutboundShipment/DetailView.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+const useDraft = createDraftStore<Transaction>();
+
+const useDraftOutbound = (id: string) => {
```
joshxg, line 32: Created this abstraction so that we could pass down a 'global' store for a draft for each 'record type'.

packages/transactions/src/OutboundShipment/DetailView.tsx
```diff
@@ -0,0 +1,87 @@
...
+    // and once saved, we replace the url with the new invoice number
+    // outbound-shipment/{invoice_number}
+    // This will cause the query key to update, and everything from this
+    // point is exactly the same as when editing an existing invoice.
+    (data, variables) => {
+      if (!variables.id) {
+        navigate({ pathname: `../${data.id}` }, { replace: true });
+      }
+
+      queryClient.invalidateQueries('transaction');
```
joshxg, line 56: This is likely going to be used for all record types, but couldn't think of a nice way to abstract this yet

packages/transactions/src/OutboundShipment/ListView.tsx
```diff
@@ -0,0 +1,105 @@
+import React, { FC, useState } from 'react';
```
joshxg, line 1: This has just been shifted from the original file. Slight change to the query to use a variable and added the new shipment button

packages/transactions/src/TransactionService.tsx
```diff
@@ -167,8 +19,19 @@ const TransactionService: FC = () => {
...
+      <Route
+        path={customerInvoicesRoute}
+        element={<OutboundShipmentListView />}
+      />
+      <Route
+        path={customerInvoiceRoute}
+        element={
+          <>
+            <OutboundShipmentDetailView />
+            <OutboundShipmentDetailView />
```
joshxg, line 31: Two of these for seeing how using two `useDraftOutbound` works

packages/transactions/src/api.ts
```diff
@@ -1,14 +1,6 @@
...
 
-import {
-  useQuery,
-  useMutation,
-  useQueryClient,
-  QueryKey,
-} from '@openmsupply-client/common';
-
-export const getQuery = (): string => gql`
+export const getDetailQuery = (): string => gql`
```
joshxg, line 3: just put the queries in here for now. Put the hook in its own file

packages/transactions/src/useDraftDocument.ts
```diff
@@ -0,0 +1,121 @@
+import { useEffect } from 'react';
```
joshxg, line 1: Just put this in its own file and a few changes added ill comment below on

packages/transactions/src/useDraftDocument.ts
```diff
@@ -0,0 +1,121 @@
...
+ * while creating a new invoice.
+ */
+
+// eslint-disable-next-line @typescript-eslint/ban-types
+export const createDraftStore = <T>(): UseStore<DraftStore<T>> => {
+  return zustand<DraftStore<T>>(set => ({
+    draft: null,
+    setDraft: draft => set(state => ({ ...state, draft })),
+  }));
+};
```
joshxg, line 28: So as mentioned, this is for replacing what was just a `useState` call in the `useDraftDocument` hook. 

This means whenever `useDraftDocument` is used, it returns the same `draft` and also is hooked up to the same `api` calls through react-querys cache

packages/transactions/src/useDraftDocument.ts
```diff
@@ -0,0 +1,121 @@
...
+ * we use a more functional approach and have functions i.e. updateDaysOfStock(draft, newValue)
+ *
+ * - Might be nice/easy/better/worse? to be able to more globally access the current draft. For example,
+ * the draft hook will be used at the 'root' of some page, and will be required to be drilled down through
+ * components which want to know the current draft values. It's not too difficult to make the value 'global'
+ * i.e. can use context, or a simple state management tool like Jotai or Recoil or..
+ * - Just use redux :thinking: :shrug: :lol:
+ */
+
+type MutateFn<DocumentType> = (
```
joshxg, line 71: The types are getting nutty so added these to try help

packages/transactions/src/useDraftDocument.ts
```diff
@@ -0,0 +1,121 @@
...
+  variables: Partial<DocumentType>
+) => void;
+
+export const useDraftDocument = <DocumentType>(
+  key: unknown[],
+  queryFn: QueryFn<DocumentType>,
+  mutateFn: MutateFn<DocumentType>,
+  onSuccess: OnSuccessCallback<DocumentType>,
+  useDraftState: UseStore<DraftStore<DocumentType>>,
+  placeholderData?: DocumentType
```
joshxg, line 88: Here i've added the props `onSuccess`, `useDraftState` and `placeholderData`.

- `onSuccess` is for helping with the changing of the url once a `new` record has been saved. Also invalidates all related queries in the react-query cache, generally. Couldn't think of a nice way to abstract this so just passing it in for now.
- `useDraftState` - passing in a zustand store to use as state for the draft. Again couldn't think of a nice way to abstract this such that only one store per entity is created. Maybe we could pass in some sort of `entity` string identifier, but not sure that's so greater
- placeholderData: As above, used to seed the `data` returned from `useQuery` for a `new` record

joshxg (commented):
Wondering if `useDocumentDraft` should be `useRecordDraft` or something.

mark-prins (approved):
🎉  that's sweet!
I think you've covered everything in your self-review - mainly around abstracting the `useDraftDocument` and such. possibly will have to pass in a text string, though at least we can enforce allowable values with a type

PR #116, opened 2021-09-06
#mws

### LINKED ISSUE #115: MSW
I just went to write a test for a component which does an http request and it becomes annoying to have the mock server running.

Same with storybook.

I don't want to spend more time setting things up but I also want to spend a bit of time setting mock service worker up so we can have mocks for testing, storybook and general development..

joshxg:
Fixes #115 

- Didn't consult, if you think this is a bad idea feel free to bin it - luckily this didn't take very long
- Has mocking for storybook/jest/browser working which I think is awesome
- My tests are lame, more about getting this in place to be able to hopefully write some half decent and useful tests
- Also think the mocks are way easier to write than a graphql server. Also had some annoyances with my setup of the mock server with babel-node not really reflecting changes nicely etc

image


.storybook/main.js
```diff
@@ -1,7 +1,23 @@
+const path = require('path');
+const toPath = filePath => path.join(process.cwd(), filePath);
+
 module.exports = {
   stories: [
     '../packages/**/*.stories.mdx',
     '../packages/**/*.stories.@(js|jsx|ts|tsx)',
   ],
   addons: ['@storybook/addon-links', '@storybook/addon-essentials'],
+  webpackFinal: async config => {
```
mark-prins, line 10: would still like to get the correct font working in storybook

mark-prins (approved):
wowsers! that's quite something.
I think it's a great idea - awesome to have the data service mocked

joshxg (commented):
I forgot to push submit on my review oops, a lot of help that is..

PR #129, opened 2021-09-07
production build

### LINKED ISSUE #122: Environment settings
couple of things to enable a production release
- set the API URL which is currently hard coded
- set the production mode value in the host webpack config

would like these to be configured at runtime by picking up env vars or a .env file or similar

mark-prins:
Fixes #122 

Adds the option to override environment based settings at runtime. Possibly the API url will just be on the same url with a specific port and we won't need this for production, instead a version which overrides for local dev. Not sure at this stage - this option allows the same code to run in different environments, so meets my need.

Also tidied the 'production' mode which wasn't being set when a build was run.

### DIFF
--- a/packages/host/webpack.config.js
+++ b/packages/host/webpack.config.js
@@ -1,4 +1,5 @@
 const ReactRefreshWebpackPlugin = require('@pmmmwh/react-refresh-webpack-plugin');
+const CopyPlugin = require('copy-webpack-plugin');
 const webpack = require('webpack');
 const HotModuleReplacementPlugin = webpack.HotModuleReplacementPlugin;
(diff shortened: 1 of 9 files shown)


packages/host/webpack.config.js
```diff
@@ -8,104 +9,110 @@ const path = require('path');
...
-    //       },
-    //     },
-    //   ],
-    // }),
-  ],
+      new HtmlWebpackPlugin({
+        favicon: './public/favicon.ico',
+        template: './public/index.html',
+      }),
+      new CopyPlugin({
```
mark-prins, line 83: tried to use assets, but could not get that to actually copy the file to the output folder. had to use this instead

joshxg (approved):
That's awesome, thanks!

Issue #128, opened 2021-09-06
Outbound: Add Row select mechanism


joshxg:
- Add a row selection mechanism like:
![image](https://user-images.githubusercontent.com/35858975/132265331-c0c6fcd3-8983-438d-92bf-f0e122b52c09.png)

- Adds 'selected rows' to some state, from which we can perform actions on afterward
- Add a delete mechanism.
- Add a 'select all' mechanism

joshxg:
Playing with this a bit

- using the `useRowSelect` plugin for react-table
- A little annoying/hard to know what the current 'selection' state is as the state is managed internally within the useTable hook, so lifting that state up to the parent of say, the listview page is a bit annoying to manage. Can have an initial state, and then pass down a callback to update some state in the parent. This works, but has the state in two places - not ideal. Also means you need to trigger a re-render on the parent to update the state - not ideal.
- Implement the functionality without the plugin - not so bad, really you just need to manage a `Record<id, boolean>` and when you render a row, check if your row is selected
- use `manualRowSelection` which means having a field on each row `isSelected` - not too bad
- Third option is using an imperative handle, something like:

```
useImperativeHandle(
    ref,
    () => ({
      selectedRows,
    }),
    [input],
  )
```

so when we render the table, we pass a `ref`, which we can then access things about the table from the parent.. we can add all the returned values: https://react-table.tanstack.com/docs/api/useRowSelect#instance-properties into this `ref` so we can have full control from the parent, kinda.

This is what mui datagrid exposes https://material-ui.com/api/data-grid/grid-api/ - where you can pass a ref to `<DataGrid ref={myGridApi} .. />`

And then in the parent `myGridApi.applyFilter()` .. it's a bit against the grain of react, but I think I like this the most

PR #131, opened 2021-09-07
#128 add row select mechanism

joshxg:
#128 partly - enough for now

- I went with a quick implementation with the `useImperativeHandle` hook to see if we like it. It's against the 'react grain' (passing props down, rather than passing a thing down to access the child) so I can see it not being liked. I am still undecided
- couple offroads
- Think best to change the icons of the checkbox for the nice border radius (which I played with and cant get to work) and color. Probably better to be using the blue as the secondary, I guess

image
image

### DIFF
--- a/packages/common/src/styles/theme.ts
+++ b/packages/common/src/styles/theme.ts
@@ -44,6 +44,8 @@ const themeOptions = {
   palette: {
     darkGrey: '#555770',
     primary: { 500: '#e95c30' },
+    secondary: { main: '#555770' },
+
     background: {
       menu: '#f2f2f5',
       white: '#fff',
--- a/packages/common/src/ui/layout/tables/RemoteDataTable.tsx
+++ b/packages/common/src/ui/layout/tables/RemoteDataTable.tsx
@@ -1,12 +1,11 @@
 /* eslint-disable react/jsx-key */
-import React, { ReactNode, useEffect, useState } from 'react';
+import React, { useEffect, useState } from 'react';
 
 import {
-  Column,
   ColumnInstance,
-  Row,
   SortingRule,
   usePagination,
+  useRowSelect,
   useSortBy,
   useTable,
 } from 'react-table';
@@ -28,34 +27,15 @@ import { useTheme } from '@material-ui/core/styles';
 
 import { SortAsc, SortDesc } from '../../icons';
 import { DEFAULT_PAGE_SIZE } from '.';
+import { TableProps } from './types';
+import { useSetupDataTableApi } from './hooks/useDataTableApi';
 
 export { SortingRule };
-export interface QueryProps<D> {
-  first: number;
-  offset: number;
-  sortBy?: SortingRule<D>[];
-}
-
-export interface QueryResponse<T> {
-  data: T[];
-  totalLength: number;
-}
-
-interface TableProps<T extends Record<string, unknown>> {
-  columns: Column<T>[];
-  data?: T[];
-  initialSortBy?: SortingRule<T>[];
-  isLoading?: boolean;
-  onFetchData: (props: QueryProps<T>) => void;
-  onRowClick?: <T extends Record<string, unknown>>(row: Row<T>) => void;
-  totalLength?: number;
-}
 
 const renderSortIcon = <D extends Record<string, unknown>>(
   column: ColumnInstance<D>
 ) => {
   if (!column.isSorted) return null;
-
   return !!column.isSortedDesc ? <SortDesc /> : <SortAsc />;
 };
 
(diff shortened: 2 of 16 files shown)


packages/common/src/styles/theme.ts
```diff
@@ -44,6 +44,8 @@ const themeOptions = {
   palette: {
     darkGrey: '#555770',
     primary: { 500: '#e95c30' },
+    secondary: { main: '#555770' },
```
joshxg, line 47: don't like but can fix with using better svg icons for the checkbox

packages/common/src/ui/layout/tables/RemoteDataTable.tsx
```diff
@@ -28,34 +27,15 @@ import { useTheme } from '@material-ui/core/styles';
 
 import { SortAsc, SortDesc } from '../../icons';
 import { DEFAULT_PAGE_SIZE } from '.';
+import { TableProps } from './types';
+import { useSetupDataTableApi } from './hooks/useDataTableApi';
```
joshxg, line 31: put the types in a file they were annoying my navigation .. 😁

packages/common/src/ui/layout/tables/RemoteDataTable.tsx
```diff
@@ -67,20 +47,14 @@ export const RemoteDataTable = <T extends Record<string, unknown>>({
   onFetchData,
   onRowClick,
   totalLength = 0,
-}: TableProps<T> & { children?: ReactNode }): JSX.Element => {
+  tableApi,
+}: TableProps<T>): JSX.Element => {
```
joshxg, line 51: played around with `forwardRef` but couldn't keep the nice generic typing and also made a little more sense to name the prop something other than `ref` (although unsure if `tableApi` is that much better.. )

packages/common/src/ui/layout/tables/RemoteDataTable.tsx
```diff
@@ -137,7 +123,10 @@ export const RemoteDataTable = <T extends Record<string, unknown>>({
               {headers.map(column => (
                 <TableCell
                   {...column.getHeaderProps(column.getSortByToggleProps())}
-                  sx={theme.typography.th}
+                  sx={{
+                    backgroundColor: 'transparent',
```
joshxg, line 127: the header row was making the drawer shadow go hunchy funchy

packages/common/src/ui/layout/tables/RemoteDataTable.tsx
```diff
@@ -162,7 +151,8 @@ export const RemoteDataTable = <T extends Record<string, unknown>>({
                     <TableCell
                       {...cell.getCellProps()}
                       sx={{
-                        padding: '9.5px',
+                        padding: 0,
+                        paddingLeft: '16px',
```
joshxg, line 155: I dunno why this started playing funny buggers

packages/common/src/ui/layout/tables/columns/CheckboxSelectionColumn.tsx
```diff
@@ -0,0 +1,26 @@
...
+        color="secondary"
+        size="small"
+        onClick={event => {
+          event.stopPropagation();
+        }}
+        {...row.getToggleRowSelectedProps()}
+      />
+    );
+  },
+});
```
joshxg, line 26: this was surprisingly very easy

packages/common/src/ui/layout/tables/hooks/useColumns.ts
```diff
@@ -17,34 +18,45 @@ export interface ColumnDefinition<T> {
...
-  <T extends Record<string, unknown>>(
-    columns: ColumnDefinition<T>[]
-  ): Column<T>[] => {
-    const t = useTranslation();
-    const formatDate = useFormatDate();
+interface ColumnOptions {
+  useSelectionColumn: boolean;
+}
+
+// eslint-disable-next-line @typescript-eslint/ban-types
```
joshxg, line 25: Had some real issues here with types! how annoying!

packages/common/src/ui/layout/tables/hooks/useColumns.ts
```diff
@@ -17,34 +18,45 @@ export interface ColumnDefinition<T> {
...
+export const useColumns = <T extends object>(
+  columnsToMap: ColumnDefinition<T>[],
+  options: ColumnOptions = { useSelectionColumn: false }
```
joshxg, line 28: I don't think this is good, but OK for now, maybe - intention wasn't for a `useX` for every column, but the more generic ones use in different places.. 🤷

packages/host/src/Host.tsx
```diff
@@ -25,7 +25,14 @@ const Content = styled(Box)({
...
   height: '100vh',
 });
-const queryClient = new QueryClient();
+
+const queryClient = new QueryClient({
+  defaultOptions: {
+    queries: {
+      refetchOnWindowFocus: false,
+    },
+  },
```
joshxg, line 34: was annoying me that whenever I click dev tools and back on the app it refetches 🤔

packages/common/src/ui/layout/tables/columns/CheckboxSelectionColumn.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { Checkbox } from '../../../components/inputs/Checkbox';
+import { Column } from 'react-table';
+
+export const getCheckboxSelectionColumn = (): Column => ({
+  id: 'selection',
+  Header: ({ getToggleAllRowsSelectedProps }) => (
```
joshxg, line 7: Looks like maybe there isn't a header with a checkbox like this and it'd actually be in some other component in the toolbar above, but left it for now

mark-prins (approved):
sweeeet!
agree about the ref - not exactly reacty but it works. Not sure about the name though.
The types are a pain, would be nice to revisit those.
I've made some checkbox icons - will do them separately though, so I can add the storybook for icons too.

packages/common/src/ui/layout/tables/hooks/useColumns.ts
joshxg: Yup much better
75e26c5

PR #134, opened 2021-09-07
Add rename customer-invoice -> outbound shipments / outbounds

### LINKED ISSUE #126: Outbound: Change URL from customer-invoice to outbound-shipment
Change URL from customer-invoice to outbound-shipment

joshxg:
Fixes #126

### DIFF
--- a/packages/common/src/intl/locales/ar.json
+++ b/packages/common/src/intl/locales/ar.json
@@ -1,7 +1,7 @@
 {
     "app.admin": "مشرف",
-    "app.customer_invoices": "فواتير العميل",
-    "app.customer-invoice": "فاتورة العميل",
+    "app.outbound-shipment": "فواتير العميل",
+    "app.outbounds": "Outbound shipments",
     "app.customer-requisition": "طلبات الشراء",
     "app.customers": "عملاء",
     "app.dashboard": "لوحة القيادة",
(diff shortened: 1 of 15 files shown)


mark-prins:
I think the replacement should be for customers. Having `Customers` / `Outbounds` in the menu doesn't make sense. If anything it would be `Customers` / `Shipments` but I would think more properly it should be:

- Dashboard
- Outbound
  - Shipments
  - Requisitions
- Inbound
  - Shipments
  - Requisitions
- Stock
- etc

will see if I can find Richard's docs on the renaming

joshxg:
Ahh yes that does make more sense.

Although I'm not sure with the additional sub headers `customers`/`credit`/`receipts`. I suppose they still make sense as `outbound credit`, `outbound receipt` (I guess that's where we are headed..?)

W

richardmoizeau:
the overall idea would be to have a physical flow associated with a financial flow. The financial flow will be included in the "physical flow" transaction: 
- Outbound Shipments / Customer Invoices
- Inbound Shipments / Supplier Invoices
- Inbound Returns / Customer Credits
- Outbound Returns / Supplier Credits

Here is the link of the suggested new workflow and some naming
https://docs.google.com/presentation/d/1eEe0uBGvkXbYnKc2oLO2U0qRwFv4l0ws4QwFZa6e74s/edit#slide=id.p

Also, the docs about the navigation panel:
https://docs.google.com/spreadsheets/d/1XRN0qhbj4i8Uo982vSAEKnwA7IyxAwbKPc0qQa66YxQ/edit#gid=1914400837

Issue #117, opened 2021-09-06
OutboundShipment: Expando toolbar with invoice details


joshxg:
- Need to put a bunch of info in the 'toolbar'/'appbar' area.

![image](https://user-images.githubusercontent.com/35858975/132264191-e600be00-691f-4e22-8a10-fa87215de357.png)

- Have the toolbar/appbar expand
- Use a portal for the outbound page to 'inject' into the appbar in the shell?
- Can just 'setup' the mechanism first before making fancy pants

joshxg:
Also on the list view:

![image](https://user-images.githubusercontent.com/35858975/132405150-cf684e8c-83eb-483a-84d1-d1e5360d595b.png)

Will start by making the `AppBar` height dynamic and adding a portal ref thing

PR #136, opened 2021-09-07
#117 expando app bar

joshxg:
Does some things for #117 

image

- Really I just added a ref and adjusted the styles for the `AppBar` to push down the content below when it expands

### DIFF
--- a/.storybook/theme.js
+++ b/.storybook/theme.js
@@ -4,7 +4,7 @@ export default create({
   base: 'light',
 
   colorPrimary: '#e95c30',
-  colorSecondary: '#e95c30',
+  colorSecondary: '#CDCDCD',
 
   // UI
   appBg: '#555770',
@@ -29,8 +29,8 @@ export default create({
   barBg: '#e95c30',
 
   // Form colors
-  inputBg: 'blue',
-  inputBorder: 'blue',
-  inputTextColor: 'blue',
+  inputBg: 'transparent',
+  inputBorder: 'transparent',
+  inputTextColor: 'black',
   inputBorderRadius: 4,
 });
--- a/packages/common/src/hooks/useHostContext.ts
+++ b/packages/common/src/hooks/useHostContext.ts
@@ -6,6 +6,7 @@ import { SupportedLocales } from '../intl/intlHelpers';
 
 type HostContext = {
   appBarButtonsRef: React.MutableRefObject<null>;
+  appBarExtra: React.MutableRefObject<null>;
   locale: SupportedLocales;
   setLocale: (locale: SupportedLocales) => void;
 };
@@ -13,6 +14,7 @@ type HostContext = {
 export const useHostContext = create<HostContext>(set => ({
   locale: LocalStorage.getItem('/localisation/locale') ?? 'en',
   appBarButtonsRef: createRef(),
+  appBarExtra: createRef(),
   setLocale: locale => set(state => ({ ...state, locale })),
 }));
 
--- a/packages/common/src/hooks/useHostContext.ts
+++ b/packages/common/src/hooks/useHostContext.ts
@@ -6,7 +6,7 @@ import { SupportedLocales } from '../intl/intlHelpers';
 
 type HostContext = {
   appBarButtonsRef: React.MutableRefObject<null>;
-  appBarExtra: React.MutableRefObject<null>;
+  appBarExtraRef: React.MutableRefObject<null>;
   locale: SupportedLocales;
   setLocale: (locale: SupportedLocales) => void;
 };
@@ -14,7 +14,7 @@ type HostContext = {
 export const useHostContext = create<HostContext>(set => ({
   locale: LocalStorage.getItem('/localisation/locale') ?? 'en',
   appBarButtonsRef: createRef(),
-  appBarExtra: createRef(),
+  appBarExtraRef: createRef(),
   setLocale: locale => set(state => ({ ...state, locale })),
 }));
 
--- a/packages/common/src/ui/components/navigation/Breadcrumbs/Breadcrumbs.stories.tsx
+++ b/packages/common/src/ui/components/navigation/Breadcrumbs/Breadcrumbs.stories.tsx
@@ -1,0 +1,72 @@
+import React from 'react';
+import { ComponentMeta, Story } from '@storybook/react';
+import { Route } from 'react-router-dom';
+
+import { Breadcrumbs } from './Breadcrumbs';
+import { TestingProvider, TestingRouter } from '../../../../utils/testing';
+import { Box } from '@material-ui/system';
+import { RouteBuilder } from '../../../..';
+import { AppRoute } from '@openmsupply-client/config';
+
+export default {
+  title: 'Components/Breadcrumbs',
+  component: Breadcrumbs,
+} as ComponentMeta<typeof Breadcrumbs>;
+
+const Template: Story<{ initialEntries: string[] }> = ({ initialEntries }) => {
+  return (
+    <TestingProvider>
+      <TestingRouter initialEntries={initialEntries}>
+        <Route
+          path="*"
+          element={
+            <Box>
+              <Breadcrumbs />
+            </Box>
+          }
+        ></Route>
+      </TestingRouter>
+    </TestingProvider>
+  );
+};
+
+export const Short = Template.bind({});
+Short.args = {
+  initialEntries: [RouteBuilder.create(AppRoute.Customers).build()],
+};
+
+export const Medium = Template.bind({});
+Medium.args = {
+  initialEntries: [
+    RouteBuilder.create(AppRoute.Customers)
+      .addPart(AppRoute.CustomerInvoice)
+      .addPart('3')
+      .build(),
+  ],
+};
+
+export const TooLong = Template.bind({});
+TooLong.args = {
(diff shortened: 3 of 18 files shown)


.storybook/theme.js
```diff
@@ -29,8 +29,8 @@ export default create({
   barBg: '#e95c30',
 
   // Form colors
-  inputBg: 'blue',
-  inputBorder: 'blue',
-  inputTextColor: 'blue',
+  inputBg: 'transparent',
+  inputBorder: 'transparent',
+  inputTextColor: 'black',
```
joshxg, line 34: offroad I couldn't see what some things were in the storybook

packages/common/src/hooks/useHostContext.ts
```diff
@@ -6,13 +6,15 @@ import { SupportedLocales } from '../intl/intlHelpers';
...
   appBarButtonsRef: React.MutableRefObject<null>;
+  appBarExtraRef: React.MutableRefObject<null>;
   locale: SupportedLocales;
   setLocale: (locale: SupportedLocales) => void;
 };
 
 export const useHostContext = create<HostContext>(set => ({
   locale: LocalStorage.getItem('/localisation/locale') ?? 'en',
   appBarButtonsRef: createRef(),
+  appBarExtraRef: createRef(),
```
joshxg, line 17: I couldn't think of a name 😭  

I also wondered if instead of just a ref for "all this area below" it should be more specific.. lhs/rhs/center kind of thing

packages/common/src/ui/components/navigation/Breadcrumbs/Breadcrumbs.stories.tsx
```diff
@@ -0,0 +1,72 @@
+import React from 'react';
```
joshxg, line 1: stories and tests for `NavLink` and `Breadcrumbs` which I also put into a `components/navigation` folder

packages/common/src/ui/components/navigation/Breadcrumbs/Breadcrumbs.test.tsx
```diff
@@ -0,0 +1,84 @@
+import React from 'react';
```
joshxg, line 1: stories and tests for `NavLink` and `Breadcrumbs` which I also put into a `components/navigation` folder

packages/common/src/ui/components/navigation/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -0,0 +1,59 @@
+import React, { useState, useEffect } from 'react';
```
joshxg, line 1: Copy past of breadcrumbs component from `AppBar` file

packages/common/src/ui/components/navigation/NavLink/NavLink.stories.tsx
```diff
@@ -0,0 +1,61 @@
+import React, { FC, useEffect } from 'react';
```
joshxg, line 1: stories and tests for `NavLink` and `Breadcrumbs` which I also put into a `components/navigation` folder

packages/common/src/ui/components/navigation/NavLink/NavLink.test.tsx
```diff
@@ -0,0 +1,77 @@
+import React, { FC, useEffect } from 'react';
```
joshxg, line 1: stories and tests for `NavLink` and `Breadcrumbs` which I also put into a `components/navigation` folder

packages/common/src/ui/components/navigation/NavLink/NavLink.tsx
```diff
@@ -1,17 +1,16 @@
 import React, { FC } from 'react';
-
 import {
   ListItem,
   ListItemIcon,
   ListItemText,
   Tooltip,
   ListItemButton,
   Box,
+  ListItemProps,
```
joshxg, line 9: shifted nav link -> components/navigation/navlink

packages/config/src/routes.ts
```diff
@@ -21,3 +21,7 @@ export enum AppRoute {
 
   Logout = 'logout',
 }
+
+export enum ExternalURL {
+  PublicDocs = 'https://docs.msupply.foundation',
+}
```
joshxg, line 27: offroad constant

packages/customers/src/Nav.tsx
```diff
@@ -30,22 +30,22 @@ const Nav: FC = () => {
...
-          <AppNavLink
+          <NavLink
             end={true}
             to={RouteBuilder.create(AppRoute.Customers)
               .addPart(AppRoute.CustomerInvoice)
               .build()}
             text={t('app.customer-invoice')}
           />
-          <AppNavLink
+          <NavLink
```
joshxg, line 48: Changed from `AppNavLink` -> `NavLink` because there wasn't enough offroad diffing in this PR

packages/host/src/AppBar.tsx
```diff
@@ -1,108 +1,43 @@
-import React, { useState } from 'react';
+import React from 'react';
```
joshxg, line 1: In here I took out the Breadcrumbs component, the `ArrowIcon` I just used `sx` and the `H6` component I tried hard to understand what `flexGrow` was doing but I couldn't figure it out so I removed it so that I can add it back later when you school me.

Also the toolbar I just used `sx` for the single prop

packages/host/src/AppBar.tsx
```diff
@@ -1,108 +1,43 @@
...
 })<{ isOpen: boolean }>(({ isOpen, theme }) => ({
-  left: 72,
-  position: 'absolute',
+  marginLeft: 80,
+  marginRight: 0,
+  minHeight: 90,
+
+  zIndex: theme.zIndex.drawer - 1,
+  boxShadow: theme.shadows[1],
+
```
joshxg, line 33: I changed this to not use `position` absolute (and am hoping thats OK) and just did some adjustments to the width/margin/left offset and it seemed to be ok..

I saw mui had an `AppBar` component and tried figuring out if we should be using that but css is so confusing 😢

packages/host/src/AppDrawer.tsx
```diff
@@ -21,9 +21,9 @@ import {
   useMediaQuery,
   useTheme,
   useTranslation,
+  NavLink,
```
joshxg, line 24: Changed from `AppNavLink` -> `NavLink` because there wasn't enough offroad diffing in this PR

packages/transactions/src/OutboundShipment/ListView/ListView.tsx
```diff
@@ -62,8 +72,20 @@ export const OutboundShipmentListView: FC = () => {
     { id: 'date', desc: true },
   ];
 
+  const [height, setHeight] = useState(300);
```
joshxg, line 75: to play with adjusting the app bar height and squashing the content

.storybook/theme.js
```diff
@@ -4,7 +4,7 @@ export default create({
   base: 'light',
 
   colorPrimary: '#e95c30',
-  colorSecondary: '#e95c30',
+  colorSecondary: '#CDCDCD',
```
mark-prins, line 7: the boss colour! or three bosses perhaps. or not. bye.

packages/common/src/hooks/useHostContext.ts
mark-prins: `appBarContentRef` ?
`appBarBodyRef` - more like an HTML document or `<Table />`
:shrug:

at this stage no need for a left/right/centre that I know of, so I'm thinking to avoid the flex hell that will give us and just use a big box for now. could be convinced though..

packages/host/src/AppBar.tsx
mark-prins: I actually think better to not use absolute if it can be avoided. as far as I can tell, we don't need it - the page length is constrained, so we don't have a need to stick the app bar to the top.

I had a look at the Mui AppBar to see if it gave us anything and from what I could see, it only gave us pain. So I stuck with this approximation instead.

packages/host/src/AppBar.tsx
```diff
@@ -129,12 +64,14 @@ const AppBar: React.FC = () => {
...
             icon={<Book />}
             labelKey="button.docs"
-            onClick={() => (location.href = 'https://docs.msupply.foundation')}
+            onClick={() => (location.href = ExternalURL.PublicDocs)}
           />
           <LanguageMenu />
         </ButtonContainer>
-      </StyledToolbar>
+      </Toolbar>
+      <ButtonContainer ref={appBarExtraRef} />
```
mark-prins, line 72: the `<ButtonContainer />` could do with a new name to avoid confusion

mark-prins (approved):
nice! I agree with the various offroads too - looking really good.

I'm getting some hunchy funchy due to the table header which is screwing with the shadow of the drawer. is that the thing you mentioned in the other PR?

Screen Shot 2021-09-08 at 3 43 45 PM

packages/host/src/AppBar.tsx
joshxg: Will do - will go and add some things into it and do it in the next pr

joshxg:
Yup the table header will (should) be fixed in one of the other prs

Issue #140, opened 2021-09-08
Outbound Shipment: Add list view actions to toolbar


joshxg:
Possibly getting ahead of myself as the designs for this seem undecided. 

However, this part has been pretty consistent for awhile in the designs:

![image](https://user-images.githubusercontent.com/35858975/132480808-7f1f986f-2217-43cc-bc1d-dae2e76522e1.png)

And I think would be nice to at least have that as a start. 

Note: Also in the `DetailView` so generic?

- `Dropdown` as a generic component
- Hook `checkbox` up to toggle all the table
- `ListViewActionBar` component? Where this dropdown and checkbox are on the RHS? 🤔

mark-prins:
yes - it would be generic I think. used in all of the list views so far
might be time for a RHS part to that appbar portal area after all :)

joshxg:
Hmm.. the drop down is there now (for the list view) with only the delete 'working' - was unsure what the `edit` was supposed to do and also exactly what the `export to csv` was supposed to do.

Still need to handle errors on the delete also

mark-prins:
I don't believe that we've talked about the `export to csv` specifically.
Some questions about it
- should it export the entire data set, or just the page shown?
- should it respect the current options selected for
  - sort
  - filter
  - columns shown

my thoughts are (a) yes and (b) yes 

delete errors.. 🤔

richardmoizeau:
delete errors = a way to cancel the latest action ? (ctrl + Z style)

joshxg:
- `Ctrl+z` / `undo` is likely going to be very difficult.
- Sorry, I meant errors as in: "You're trying to delete a finalised/allocated/shipped invoice, you can't do that bro"

richardmoizeau:
- Ok so I guess a good old fashion dialog window to say that "You're about to do something stupid. Would you like to proceed?" (https://zpl.io/25z0ore)
- for the delete errors, I think notifications could do the trick ? "You can't delete, go home"

<img width="479" alt="Screen Shot 2021-09-27 at 12 49 12 PM" src="https://user-images.githubusercontent.com/74992958/134828568-c93f1348-4c91-4096-970d-e8dcdb6ddbef.png">

PR #142, opened 2021-09-09
#140 listview actionbar

joshxg:
Fixes nothing!

- I renamed the abbbar ref
- I made a `portals` folder with the intention that we can make portal 'pair' components and have the 'source' rendered somewhere and then we just need to wrap some content in `<AppbarContentPortal>` which is neat and potentially OTT
- I put a select in a `Dropdown` file which is super helpful. Looking to style it later. Also why did I name it `Dropdown`?
- Also added some methods to the `tableApi` ref and it was at this point I started googleing. Then I found this issue: https://github.com/tannerlinsley/react-table/issues/2459 and decided that we were going to need to a different approach for managing the 'table state' outside of `react-table` itself (i.e. to have a checkbox above the `table` component in the dom tree).
- I basically want to touch a bit in the table to handle the state better than the `tableRef` approach, but thought this work was at least somewhat useful

### DIFF
--- a/packages/common/src/ui/components/portals/AppBarContent/AppBarContent.test.tsx
+++ b/packages/common/src/ui/components/portals/AppBarContent/AppBarContent.test.tsx
@@ -35,14 +35,17 @@ describe('AppBarContent', () => {
   });
 
   it('Portal children dismount if the portal dismounts', () => {
-    const { getByText, getByRole } = render(<TestAppBarContent initialShow />);
+    const { queryByText, getByRole } = render(
+      <TestAppBarContent initialShow />
+    );
 
-    const node = getByText(/josh/);
     const button = getByRole(/button/);
 
     act(() => {
       userEvent.click(button);
     });
+
+    const node = queryByText(/josh/);
 
     expect(node).not.toBeInTheDocument();
   });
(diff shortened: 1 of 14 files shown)


packages/common/src/ui/components/portals/AppBarContent/AppBarContent.test.tsx
```diff
@@ -0,0 +1,57 @@
+import React, { FC } from 'react';
+import { act, render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { AppBarContentPortal, AppBarContent } from './AppBarContent';
+
+describe('AppBarContent', () => {
```
mark-prins, line 6: this is genius!

packages/common/src/ui/components/portals/AppBarContent/AppBarContent.tsx
```diff
@@ -17,14 +17,22 @@ const Container = styled('div')({
 // mounted at any one time as mounting this in multiple locations would cause
 // some pretty weird behaviour
 export const AppBarContent: FC = () => {
-  const { appBarContentRef } = useHostContext();
+  const { setAppBarContentRef } = useHostContext();
+  const ref = useRef(null);
+
+  useEffect(() => {
+    setAppBarContentRef(ref);
+    () => setAppBarContentRef(null);
```
mark-prins, line 25: how is this working without returning the function? the test passes, so something is happening. 😕

mark-prins (approved):
👍  yep. that's awesome
I got into a bit of a tangle with creating a ref for the button portal in the calling component. this is what I was trying to do - nice work!

joshxg:
Ah! That was an important lesson, nice, thanks.

- Turns out the order of the `act` seems relevant
- Also a ref is set to null automatically by react when dismounting

PR #145, opened 2021-09-11
#144 row selection and delete

### LINKED ISSUE #144: Outbound shipment: Change selection/row state
- The current `tableApi` method of toggling a selected state on a row doesn't work, as we can't trigger re-renders above the `table` when the `table` state changes

Options:

- Add row state into the row data objects. For example, if we have a trans line: `{id: 'x', quantity: 1}`, we can add `isSelected`. In addition to this, we would need to add something like this to set the table state:
```
useEffect(() => {
    rows.forEach(({ id, original }) => {
        if ("isSelected" in original) {
            toggleRowSelected(id, original.isSelected);
        }
    });
}, [rows, toggleRowSelected]);
```
- Use `rowState` and check the state when rendering a row- I think the way to go

joshxg:
Fixes #144 

- Adds a dropdown which does some things
image
image
- Wanted to align the checkbox to the right, so made some changes around that area. Extracted the row to `DataRow` (bad name?) and add an `align` into the `column` definition.

Some issues to make from here:
- Create a hook `useListData` (?) - This can encapsulate the query for some list data and mutations/actions performed on the data to tidy things up a bit.
- Extract out `HeaderRow` component
- Made me think about what is the behaviour when a row is finalised. We want to be able to select it, but we don't want to be able to delete it. Also if there's a mix of finalised/not finalised in the list of selected things. Possibly showing on the dropdown "disabled as you have finalised things and we can't delete them" 🤷

### DIFF
--- a/packages/common/src/intl/locales/pt.json
+++ b/packages/common/src/intl/locales/pt.json
@@ -27,6 +27,7 @@
   "button.try-again": "Try again",
   "button.go-back": "Go back",
   "button.delete-lines": "Delete selected lines",
+  "button.export-to-csv": "Export to CSV",
   "error.something-wrong": "Oops! Somethings gone wrong.",
   "label.customer": "Customer",
   "label.date": "Date",
--- a/packages/common/src/intl/locales/pt.json
+++ b/packages/common/src/intl/locales/pt.json
@@ -26,6 +26,7 @@
   "button.open-the-menu": "Abra o menu",
   "button.try-again": "Try again",
   "button.go-back": "Go back",
+  "button.delete-lines": "Delete selected lines",
   "error.something-wrong": "Oops! Somethings gone wrong.",
   "label.customer": "Customer",
   "label.date": "Date",
--- a/packages/common/src/ui/components/inputs/Dropdown/Dropdown.test.tsx
+++ b/packages/common/src/ui/components/inputs/Dropdown/Dropdown.test.tsx
@@ -1,0 +1,60 @@
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import React from 'react';
+import { act } from 'react-dom/test-utils';
+import { Dropdown, DropdownItem } from '.';
+import { TestingProvider } from '../../../../utils';
+
+describe('Dropdown', () => {
+  it('Renders the dropdown item children when the dropdown is clicked', () => {
+    const { getByRole, getByText } = render(
+      <TestingProvider>
+        <Dropdown label="dropdown">
+          <DropdownItem>One</DropdownItem>
+        </Dropdown>
+      </TestingProvider>
+    );
+
+    const button = getByRole('button');
+
+    act(() => {
+      userEvent.click(button);
+    });
+
+    const node = getByText(/one/i);
+
+    expect(node).toBeInTheDocument();
+  });
+
+  it('Renders the dropdown item children when the dropdown is clicked', () => {
+    const TestDropdown = () => {
+      const [text, setText] = React.useState('one');
+
+      return (
+        <TestingProvider>
+          <Dropdown label="dropdown">
+            <DropdownItem onClick={() => setText('two')}>{text}</DropdownItem>
+          </Dropdown>
+        </TestingProvider>
+      );
+    };
+
+    const { getByRole, getByText } = render(<TestDropdown />);
+
+    const button = getByRole('button');
+
+    act(() => {
+      userEvent.click(button);
+    });
+
+    let node = getByText(/one/i);
+
+    act(() => {
+      userEvent.click(node);
+    });
+
+    node = getByText(/two/i);
+
+    expect(node).toBeInTheDocument();
+  });
+});
--- a/packages/common/src/ui/components/inputs/Dropdown/Dropdown.test.tsx
+++ b/packages/common/src/ui/components/inputs/Dropdown/Dropdown.test.tsx
@@ -1,4 +1,4 @@
-import { render } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import React from 'react';
 import { act } from 'react-dom/test-utils';
@@ -53,7 +53,7 @@ describe('Dropdown', () => {
       userEvent.click(node);
     });
 
-    node = getByText(/two/i);
+    node = screen.getByText(/two/i);
 
     expect(node).toBeInTheDocument();
   });
--- a/packages/common/src/ui/components/inputs/Dropdown/Dropdown.test.tsx
+++ b/packages/common/src/ui/components/inputs/Dropdown/Dropdown.test.tsx
@@ -1,4 +1,4 @@
-import { render, screen } from '@testing-library/react';
+import { render } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import React from 'react';
 import { act } from 'react-dom/test-utils';
@@ -53,7 +53,7 @@ describe('Dropdown', () => {
       userEvent.click(node);
     });
 
-    node = screen.getByText(/two/i);
+    node = getByText(/two/i);
 
     expect(node).toBeInTheDocument();
   });
--- a/packages/common/src/ui/components/inputs/Dropdown/Dropdown.tsx
+++ b/packages/common/src/ui/components/inputs/Dropdown/Dropdown.tsx
@@ -1,15 +1,14 @@
-import React, { FC, useState } from 'react';
+import React, { FC } from 'react';
 import {
   MenuItem,
   MenuItemProps,
   FormControl,
   Select,
-  ListItemText,
-  Menu,
+  InputLabel,
+  outlinedInputClasses,
 } from '@material-ui/core';
-import { outlinedInputClasses } from '@material-ui/core/OutlinedInput';
-import InputLabel from '@material-ui/core/InputLabel';
 import { styled } from '@material-ui/system';
+
 import { ChevronDown } from '../../../icons';
 
 export const DropdownItem: FC<MenuItemProps> = props => {
@@ -18,7 +17,7 @@ export const DropdownItem: FC<MenuItemProps> = props => {
 
 const StyledSelect = styled(Select)(({ theme }) => ({
   borderRadius: '8px',
-  minWidth: 160,
+  width: 160,
   backgroundColor: 'white',
   '& .MuiSelect-icon': {
     // If left is not explicitly defined, sometimes the icon floats to the left
@@ -39,30 +38,15 @@ const StyledSelect = styled(Select)(({ theme }) => ({
     },
 }));
 
-interface Action {
-  label: string;
-  callback: () => void;
-}
-interface ActionDropdownProps {
-  actions: Action[];
+interface DropdownProps {
   label: string;
 }
 
 // Styled doesn't like `sx` prop being passed to it.
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-export const ActionDropdown: FC<ActionDropdownProps> = ({ label, actions }) => {
-  const [anchorEl, setAnchorEl] = useState<null | HTMLElement>(null);
-
-  const handleClick = (event: React.MouseEvent<HTMLDivElement>) => {
-    setAnchorEl(event.currentTarget);
-  };
-
-  const handleClose = () => {
-    setAnchorEl(null);
-  };
-
+export const Dropdown: FC<DropdownProps> = ({ label, children }) => {
   return (
-    <FormControl margin="dense" size="small">
+    <FormControl size="small">
       <InputLabel
         shrink={false}
         sx={{ color: '#8f90a6', '&.Mui-focused': { color: '#8f90a6' } }}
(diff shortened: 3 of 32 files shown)


packages/common/src/ui/components/inputs/Dropdown/Dropdown.tsx
```diff
@@ -1,16 +1,92 @@
...
+      ) : null}
+      <ListItemText
+        inset={props.inset}
+        primaryTypographyProps={{ sx: { fontSize: 'inherit' } }}
+        sx={{ '&.MuiListItemText-inset': { paddingLeft: '22px' } }}
+      >
+        {props.children}
+      </ListItemText>
+    </MenuItem>
+  );
```
joshxg, line 38: A little bit of extras in here. I only noticed the icons in the zeplin a little later

image

And then added a little bit to be able to inset the font and have a nicer mix of rows with icons and those without. 

It's a little restrictive in the font sizes.. somewhat intentionally (for consistency) and somewhat because I didn't want to spend anymore time on this..!

packages/common/src/ui/components/inputs/Dropdown/Dropdown.tsx
```diff
@@ -1,16 +1,92 @@
...
+    <MenuItem sx={{ fontSize: '10' }} {...props}>
+      {props.IconComponent ? (
+        <Box mr={1}>
+          <props.IconComponent fontSize="inherit" />
+        </Box>
+      ) : null}
+      <ListItemText
+        inset={props.inset}
+        primaryTypographyProps={{ sx: { fontSize: 'inherit' } }}
+        sx={{ '&.MuiListItemText-inset': { paddingLeft: '22px' } }}
```
joshxg, line 33: TIL is that sometimes these class overrides work with a space between `&` and `.` and sometimes they need no space. Unsure if intentional yet, but if it is intentional - the person who decided that needs a good chatting to.

packages/common/src/ui/components/inputs/Dropdown/Dropdown.tsx
```diff
@@ -1,16 +1,92 @@
...
+        {props.children}
+      </ListItemText>
+    </MenuItem>
+  );
 };
 
-const StyledSelect = styled(Select)({});
+const StyledSelect = styled(Select)(({ theme }) => ({
+  borderRadius: '8px',
+  width: 160,
```
joshxg, line 43: Again restrictive 🤔  probably not ideal for the width and maybe should be a min width instead. Then I thought that maybe using a `size` or `width` prop which was an enum would be better and then I gave up

packages/common/src/ui/components/inputs/Dropdown/Dropdown.tsx
```diff
@@ -1,16 +1,92 @@
...
+  backgroundColor: 'white',
+  '& .MuiSelect-icon': {
+    // If left is not explicitly defined, sometimes the icon floats to the left
+    left: 'calc(100% - 30px)',
+    color: theme.palette.primary.main,
```
joshxg, line 48: This was interesting.

I had an auto-import: `import {styled} from @material-ui/system` which made me use the color like: `theme.palette['primary']['500']`. which is technically correct? I am not sure exactly which one to be using!

packages/common/src/ui/components/inputs/Dropdown/Dropdown.tsx
```diff
@@ -1,16 +1,92 @@
...
+    borderColor: '#e4e4eb',
+  },
+  [`&:hover .${outlinedInputClasses.notchedOutline}`]: {
+    borderColor: '#e4e4eb',
+  },
+
+  [`&.${outlinedInputClasses.focused} .${outlinedInputClasses.notchedOutline}`]:
+    {
+      borderColor: theme.palette['darkGrey'],
+    },
```
joshxg, line 61: Ugh! These took me forever to override, and i'm not even sure I was meant to 🙃

packages/common/src/ui/components/inputs/Dropdown/Dropdown.tsx
```diff
@@ -1,16 +1,92 @@
...
+    // If left is not explicitly defined, sometimes the icon floats to the left
+    left: 'calc(100% - 30px)',
+    color: theme.palette.primary.main,
+  },
+
+  [`& .${outlinedInputClasses.notchedOutline}`]: {
+    borderColor: '#e4e4eb',
+  },
+  [`&:hover .${outlinedInputClasses.notchedOutline}`]: {
+    borderColor: '#e4e4eb',
```
joshxg, line 55: Should be in the theme as a `palette` I think?

packages/common/src/ui/components/inputs/Dropdown/Dropdown.tsx
```diff
@@ -1,16 +1,92 @@
...
 // Styled doesn't like `sx` prop being passed to it.
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-export const Dropdown: FC<SelectProps> = ({ sx, ...props }) => {
-  return <StyledSelect {...props} />;
+export const Dropdown: FC<DropdownProps> = ({ label, children }) => {
+  return (
+    <FormControl size="small">
+      <InputLabel
+        shrink={false}
+        sx={{ color: '#8f90a6', '&.Mui-focused': { color: '#8f90a6' } }}
```
joshxg, line 75: Theme? Wanted to try figure out typographys, sure can set something up to make this automatically the color.

Also: No space between `&` and `.` for this one.

packages/common/src/ui/components/inputs/Dropdown/Dropdown.tsx
```diff
@@ -1,16 +1,92 @@
...
+
+interface DropdownProps {
+  label: string;
+}
 
 // Styled doesn't like `sx` prop being passed to it.
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-export const Dropdown: FC<SelectProps> = ({ sx, ...props }) => {
-  return <StyledSelect {...props} />;
+export const Dropdown: FC<DropdownProps> = ({ label, children }) => {
```
joshxg, line 70: I'm not spreading props intentionally. 1) which props? For the label or the select or the form control or what? Need to add a prop for each of the components props?

I think I am leaning toward not spreading props and specific use cases for components.. (dreaming?)

packages/common/src/ui/layout/tables/RemoteDataTable.tsx
```diff
@@ -115,54 +104,59 @@ export const RemoteDataTable = <T extends Record<string, unknown>>({
...
+                  >
+                    <TableSortLabel
+                      hideSortIcon={column.id === 'selection'}
+                      active={column.isSorted}
+                      direction={column.isSortedDesc ? 'desc' : 'asc'}
+                      IconComponent={SortDesc}
+                    >
+                      {column.render('Header')}
+                    </TableSortLabel>
+                  </TableCell>
```
joshxg, line 137: Changes here were to try and be able to right align the checkbox header.

I went searching through the mui docs table examples and found this `TableSortLabel` component which flips the icon n stuff.

the `sortDirection` on the `TableCell` is apparently for blind peeps. 

will make an issue to encapsulate this elsewhere

packages/common/src/ui/layout/tables/RemoteDataTable.tsx
```diff
@@ -115,54 +104,59 @@ export const RemoteDataTable = <T extends Record<string, unknown>>({
...
-                      }}
-                    >
-                      {cell.render('Cell')}
-                    </TableCell>
-                  );
-                })}
-              </TableRow>
+              <DataRow<T>
+                cells={cells}
+                values={values as T}
```
joshxg, line 153: Jeez these types are really starting to get annoying!

packages/mock-server/src/schema/Transaction.js
```diff
@@ -71,6 +71,15 @@ const TransactionQueryResolvers = {
...
 const TransactionMutationResolvers = {
+  deleteTransactions: (_, { transactions }) => {
+    transactions.forEach(({ id: deleteId }) => {
+      const idx = TransactionData.findIndex(({ id }) => deleteId === id);
+      TransactionData.splice(idx, 1);
+    });
+
+    return transactions;
+  },
+
```
joshxg, line 82: I should've done this is a different pr :( - deleting transaction mock and handler in the mock server

packages/transactions/src/OutboundShipment/ListView/ListView.tsx
```diff
@@ -70,19 +87,39 @@ export const OutboundShipmentListView: FC = () => {
...
-          <DropdownItem value={10}>Ten</DropdownItem>
-          <DropdownItem value={20}>Twenty</DropdownItem>
+        <Dropdown label="Select">
+          <DropdownItem
+            IconComponent={ChevronDown}
+            onClick={() => {
+              const linesToDelete = tableApi?.current?.selectedRows;
+              if (linesToDelete && linesToDelete?.length > 0) {
+                mutateAsync(linesToDelete);
+                success(`Deleted ${linesToDelete?.length} invoices`)();
```
joshxg, line 102: Bit weird actually? Maybe we should return 

```
const success = (message:string) => enqueueSnackbar(message, {variant: 'success'})
```

rather than the function? 

I admit I was a bit silly and got a bit confused passing `onClick={() => success("..")}`  - seems like the most common usage

types/react-table-config.d.ts
```diff
@@ -49,11 +49,13 @@ import {
 } from 'react-table';
 
 declare module 'react-table' {
+  type Align = { align?: 'left' | 'center' | 'right' };
   export interface ColumnInstance<
     D extends Record<string, unknown> = Record<string, unknown>
   > extends UseFiltersColumnProps<D>,
       UseGroupByColumnProps<D>,
       UseResizeColumnsColumnProps<D>,
+      Align,
```
joshxg, line 58: Seems we can add anything into the columns we like which is nice. But the fields 1) have to be optional 2) need to be defined in a type explictly.

i.e.
```
...
UseResizeColumnsColumnProps<D>,
{ align: 'left' | 'center' | 'right' },
...
```

didn't work hehe

mark-prins:
there might be a missing commit. the DataRow is throwing `cannot read property 'align' of undefined` in the test and storybook

joshxg:
Ugh sorry about that- pushed

packages/common/src/ui/components/inputs/Dropdown/Dropdown.tsx
mark-prins: I hadn't noticed the icons before, though that looks like one of the older designs 🤷‍♂️

packages/common/src/ui/components/inputs/Dropdown/Dropdown.tsx
mark-prins: wondered similar.  leave for now though - looks great and there's plenty in this PR already

packages/common/src/ui/components/inputs/Dropdown/Dropdown.tsx
mark-prins: I prefer the version you've used.. easier to read.
The 500 is the font weight ( i.e. normal, with bold at about 700 ) which is why leaving it off will look the same as the `['500']` version

packages/common/src/ui/components/inputs/Dropdown/Dropdown.tsx
mark-prins: yeah, ideally would be.

packages/common/src/ui/components/inputs/Dropdown/Dropdown.tsx
mark-prins: very tricksy!! impressive css work there

packages/common/src/ui/components/inputs/Dropdown/Dropdown.tsx
mark-prins: if we can, then absolutely for the specific use cases in components. that will give us consistency of the look. allowing any props will lead to all sorts of chaos

packages/common/src/ui/layout/tables/RemoteDataTable.tsx
mark-prins: wicked.
I see that the icon has a tooltip 'toggle SortBy' which isn't super informative. not sure if that can be overridden

packages/mocks/src/handlers.ts
```diff
@@ -2,7 +2,7 @@ import { graphql } from 'msw';
 
 import faker from 'faker';
 
-const TransactionData = Array.from({ length: 500 }).map((_, i) => ({
+const TransactionData = Array.from({ length: 10 }).map((_, i) => ({
```
mark-prins, line 5: you don't like my big data set?

packages/transactions/src/OutboundShipment/ListView/ListView.tsx
mark-prins: probably should - the function was a convenience. which only seems to help in that one specific scenario.

If we do keep the function, then this might make this usage clearer:

```
                const notifyDeleteSuccess = success(`Deleted ${linesToDelete?.length} invoices`);

// and then, in the delete event:
                notifyDeleteSuccess();
```

mark-prins (commented):
that is awesome!
Just adding as a comment, so I can refresh..

mark-prins (approved):
magic!

packages/common/src/intl/locales/pt.json
```diff
@@ -26,6 +26,8 @@
   "button.open-the-menu": "Abra o menu",
   "button.try-again": "Try again",
   "button.go-back": "Go back",
+  "button.delete-lines": "Delete selected lines",
+  "button.export-to-csv": "Export to CSV",
```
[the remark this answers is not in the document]
joshxg: I've also been meaning to look through how the package can extract the strings from the code base also

packages/common/src/ui/components/inputs/Dropdown/Dropdown.test.tsx
```diff
@@ -0,0 +1,60 @@
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import React from 'react';
+import { act } from 'react-dom/test-utils';
+import { Dropdown, DropdownItem } from '.';
+import { TestingProvider } from '../../../../utils';
+
+describe('Dropdown', () => {
+  it('Renders the dropdown item children when the dropdown is clicked', () => {
```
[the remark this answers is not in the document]
joshxg: I am not getting this one.

When I remove 

```
act(() => {
      userEvent.click(button);
    });
```

The test fails

joshxg:
Renamed `Dropdown` -> `DropdownMenu` and put it in a `menus` folder (🤔 )

Reverted the handler reduction in data 😁  (accidental!!) and fixed the test description. Unsure about the flakey test?

PR #152, opened 2021-09-12
#149 update color

### LINKED ISSUE #149: Outbound shipments: Colour column
This column here:

image

Should be easy, likely isn't. Hopefully is a simple way to do it.

- A column-header-less column?
- Embedded into the cell of the first column?
- Do we send a request and update automatically, or do we have a save for this action also? (I assume that there will also be bulk actions with the drop down and selection - i.e. selection 3 rows, set as finalised for example, but that at least is for bulk editing, where as this is likely single request per row. Not a big deal though, imo).

joshxg:
Fixes #149 

image

- I think this approach is OK. Using a special column rather than doing some tricky embedding of the color. I think the color is used sparingly enough that we don't need some sort of generic solution for it.
- I tried a few approaches to passing the `onChange` to the cell
    - Tried to add in an `onChange` into the column definition. I wasn't able to correctly type this very well - I could add the function, but couldn't pass through more generics (for the `onChange` params). Would also likely need to add sorts of handlers into the column definition? (onSubmit, onClick) which would mean they all become optional and that's a bit lame.
    - Another option is to add into the columns a `getHandler` function, which the cell can call, then the `ListView` can have a function which returns the correct callback. This is basically the same as how mobiles works. I think that it is pretty good option
    - Tried using a context and then the cell can call `const {handler} = useCellHandler(TableContext, rowKey, columnKey)` - this is basically the same as the above approach using a context rather than a `hander` callback in the column definition. I like this also.
    - Landed on a cell factory kinda pattern. I think it's pretty decent, but I *think* I like the context solution better 😁  .. this seems to work, though 🙄

### DIFF
--- a/packages/common/src/ui/components/menus/ColorMenu/ColorMenu.tsx
+++ b/packages/common/src/ui/components/menus/ColorMenu/ColorMenu.tsx
@@ -62,7 +62,8 @@ export const ColorMenu: FC<ColorMenuProps> = ({
           role="button"
           aria-label={name}
           key={hex}
-          onClick={() => {
+          onClick={e => {
+            e.stopPropagation();
             onClick({ hex, name });
             onClose();
           }}
--- a/packages/common/src/ui/layout/tables/columns/NameAndColorColumn.tsx
+++ b/packages/common/src/ui/layout/tables/columns/NameAndColorColumn.tsx
@@ -46,7 +46,7 @@ export const getNameAndColorColumn = <T extends Colorable>(
           }}
         />
         <UnstyledIconButton
-          titleKey="app.admin"
+          titleKey="button.select-a-color"
           icon={
             <Circle
               htmlColor={row.original.color}
--- a/packages/common/src/ui/layout/tables/columns/NameAndColorColumn.tsx
+++ b/packages/common/src/ui/layout/tables/columns/NameAndColorColumn.tsx
@@ -1,0 +1,71 @@
+import React, { useState } from 'react';
+import { CellProps, Column } from 'react-table';
+import { TableCell } from '@material-ui/core';
+import { Color, ColorMenu, UnstyledIconButton } from '../../../components';
+import { Circle } from '../../../icons';
+import { Box } from '@material-ui/system';
+
+type Colorable = {
+  color: string;
+} & Record<string, unknown>;
+
+export const getNameAndColorColumn = <T extends Colorable>(
+  onChange: (row: T, color: Color) => void
+): Column<T> => ({
+  id: 'color',
+  align: 'left',
+
+  Cell: ({ row, value }: CellProps<T>) => {
+    const [anchorEl, setAnchorEl] = useState<HTMLButtonElement | null>(null);
+
+    const handleClick = (event: React.MouseEvent<HTMLButtonElement>) => {
+      setAnchorEl(event.currentTarget);
+    };
+
+    const handleClose = () => {
+      setAnchorEl(null);
+    };
+
+    return (
+      <TableCell
+        padding="none"
+        align="left"
+        sx={{
+          flexDirection: 'row',
+          borderBottom: 0,
+          alignItems: 'center',
+          display: 'flex',
+        }}
+      >
+        <ColorMenu
+          onClose={handleClose}
+          anchorEl={anchorEl}
+          onClick={color => {
+            handleClose();
+            onChange(row.original, color);
+          }}
+        />
+        <UnstyledIconButton
+          titleKey="app.admin"
+          icon={
+            <Circle
+              htmlColor={row.original.color}
+              sx={{
+                width: '12px',
+                margin: 'margin: 0 9px 0 10px',
+                overflow: 'visible',
+                cursor: 'pointer',
+              }}
+            />
+          }
+          onClick={e => {
+            e.stopPropagation();
+            handleClick(e);
+          }}
+        />
+        <Box ml={1} />
+        {value}
+      </TableCell>
+    );
+  },
+});
--- a/packages/mocks/src/handlers.ts
+++ b/packages/mocks/src/handlers.ts
@@ -48,7 +48,12 @@ const upsertTransaction = graphql.mutation(
       return response(context.data({ upsertTransaction: newTransaction }));
     }
 
-    return response(context.data({}));
+    const idx = TransactionData.findIndex(
+      ({ id: filterId }) => id === filterId
+    );
+    TransactionData[idx] = { ...TransactionData[idx], ...patch };
+
+    return response(context.data({ upsertTransaction: TransactionData[idx] }));
   }
 );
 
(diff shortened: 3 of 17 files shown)


packages/common/src/ui/components/menus/ColorMenu/ColorMenu.tsx
```diff
@@ -62,7 +62,8 @@ export const ColorMenu: FC<ColorMenuProps> = ({
           role="button"
           aria-label={name}
           key={hex}
-          onClick={() => {
+          onClick={e => {
+            e.stopPropagation();
```
joshxg, line 66: this stops the row event from triggers. Unsure if the best place

packages/common/src/ui/layout/tables/columns/NameAndColorColumn.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { Circle } from '../../../icons';
+import { Box } from '@material-ui/system';
+
+type Colorable = {
+  color: string;
+} & Record<string, unknown>;
+
+export const getNameAndColorColumn = <T extends Colorable>(
+  onChange: (row: T, color: Color) => void
+): Column<T> => ({
```
joshxg, line 14: Was going to create an interface for `ColumnFactory` .. but wanted to see what you thought about this approach rather than the other options

packages/common/src/ui/layout/tables/columns/NameAndColorColumn.tsx
```diff
@@ -0,0 +1,71 @@
...
+          display: 'flex',
+        }}
+      >
+        <ColorMenu
+          onClose={handleClose}
+          anchorEl={anchorEl}
+          onClick={color => {
+            handleClose();
+            onChange(row.original, color);
+          }}
```
joshxg, line 46: This is using the original data row we passed in. The `row` object has the values transformed using the `accessor` and also only contains the fields for the row (so color is technically missing in the `row` object, as the key for this row is `name`)

packages/mocks/src/handlers.ts
```diff
@@ -48,7 +49,12 @@ const upsertTransaction = graphql.mutation(
       return response(context.data({ upsertTransaction: newTransaction }));
     }
 
-    return response(context.data({}));
+    const idx = TransactionData.findIndex(
+      ({ id: filterId }) => id === filterId
+    );
+    TransactionData[idx] = { ...TransactionData[idx], ...patch };
+
+    return response(context.data({ upsertTransaction: TransactionData[idx] }));
```
joshxg, line 57: This handler was just handling new upserts. Took the rest of the code I forgot to add from the mock server to also handle actual upserts

packages/transactions/src/OutboundShipment/DetailView/DetailView.tsx
```diff
@@ -2,36 +2,17 @@ import React, { FC } from 'react';
...
 
-import {
-  request,
-  Transaction,
-  useQueryClient,
-} from '@openmsupply-client/common';
+import { Transaction, useQueryClient } from '@openmsupply-client/common';
 
-import { getMutation, getDetailQuery } from '../../api';
+import { detailQueryFn, updateTransactionFn } from '../../api';
```
joshxg, line 7: Extracted the functions in the `ListView` and `DetailView` out to the `api` file which contains all the code for the detail+list queries, delete mutation and upsert mutations as I reused the upsert mutation in the `ListView`

joshxg: Think these will pretty much all go into custom hooks at some point

packages/transactions/src/OutboundShipment/ListView/ListView.tsx
```diff
@@ -63,18 +42,54 @@ export const OutboundShipmentListView: FC = () => {
...
+
+      const existingRowIdx = previousData.findIndex(
+        ({ id }) => id === patch.id
+      );
+      previousData[existingRowIdx] = patch;
+
+      queryClient.setQueryData(key, { ...previousData, data: previousData });
+
+      return { previousCached };
+    },
```
joshxg, line 73: optimistically updating (and not handling the error case) - think there should be an easier way to do this. Likely need to do it for other things like deletion and any mutations on the `ListView` which aren't triggered by an `onSAve` event

joshxg: something like: Revert + snackbar to say "Nope!"

joshxg (commented):
Seems sluggish. Need to investigate more, but seems like there's 3-4 re-renders of the whole thing when selecting a color 🙄

packages/common/src/ui/components/menus/ColorMenu/ColorMenu.tsx
mark-prins: I think this is the right place

packages/transactions/src/api.ts
```diff
@@ -46,3 +49,41 @@ export const getListQuery = (): string => gql`
...
+
+export const detailQueryFn = (id: string) => async (): Promise<Transaction> => {
+  const result = await request(Environment.API_URL, getDetailQuery(), {
+    id,
+  });
+  const { transaction } = result;
+  return transaction;
+};
+
+export const updateTransactionFn = async (
```
mark-prins, line 82: the function names are inconsistent - in that most don't include `transaction` (presumably because they are in the `/transactions/api`) but the `updateTransactionFn` does. How about renaming to `updateFn`?

packages/transactions/src/OutboundShipment/ListView/ListView.tsx
```diff
@@ -63,18 +42,53 @@ export const OutboundShipmentListView: FC = () => {
...
+        );
+
+      const previousData = [...(previousCached?.data ?? [])];
+
+      const existingRowIdx = previousData.findIndex(
+        ({ id }) => id === patch.id
+      );
+      previousData[existingRowIdx] = patch;
+
+      queryClient.setQueryData(key, { ...previousData, data: previousData });
```
mark-prins, line 69: I'm getting confused. Shouldn't this be `...previousCached`?

packages/transactions/src/OutboundShipment/ListView/ListView.tsx
mark-prins: yeah, feels like a big manual process. if nothing else, good to extract it out if we can move this into generic hooks. logic looks good though 🤷‍♂️

mark-prins (approved):
🎉  I think it's great. covers a lot of interesting ground too, which is useful for other stuff.

I was thinking to have a separate column for color; but the name+color approach is good. likely more useful this way.

joshxg:
Updated the incorrect optimistic update that you caught (thanks) and renamed the func `updateFn`. Also added updates to the graphql schema for `color` that I had forgotten.

PR #155, opened 2021-09-13
#153 variable width columns

### LINKED ISSUE #153: Outbound shipment: Variable width columns
Want to add variable column widths/flexing. It makes things look better

image

- Add better mock data with more columns and better values
- use the `useFlexLayout` plugin

joshxg:
Fixes #153 

image

- Makes the columns flex - works OK. Comment column needs work, but I figure if we want to have that much text we should use a special column type with a character limit and somewhat to read it - even if it's just a hover so left it for now
- Changed up the transaction mock server and mock worker data - looks better..? Also should fix the demo server which I broke - well, it's all working - but so did the last branch that I merged so.. 🙃 !


packages/common/src/ui/layout/tables/RemoteDataTable.tsx
```diff
@@ -7,6 +7,7 @@ import {
   useRowSelect,
   useSortBy,
   useTable,
+  useFlexLayout,
```
joshxg, line 10: In the end all this really does is pass the width as a flex basis to the cells 🤔

packages/common/src/ui/layout/tables/RemoteDataTable.tsx
```diff
@@ -108,15 +110,31 @@ export const RemoteDataTable = <T extends Record<string, unknown>>({
...
+                ...getHeaderGroupProps().style,
+                display: 'flex',
+                flex: '1 0 auto',
+
+                height: '60px',
+                paddingLeft: '20px',
+                paddingRight: '20px',
+                alignItems: 'center',
+              }}
+            >
```
joshxg, line 125: Had made an issue to extract this row out

packages/common/src/ui/layout/tables/RemoteDataTable.tsx
```diff
@@ -145,19 +164,18 @@ export const RemoteDataTable = <T extends Record<string, unknown>>({
...
+            const { cells, original } = row;
             const { key } = row.getRowProps();
 
             return (
               <DataRow<T>
                 cells={cells}
-                values={values as T}
                 key={key}
                 onClick={onRowClick}
+                rowData={original}
```
joshxg, line 175: oops I got confused. `values` is the values actually in the table, which doesn't include `color` or `id`

packages/common/src/ui/layout/tables/columns/NameAndColorColumn.tsx
```diff
@@ -27,12 +26,10 @@ export const getNameAndColorColumn = <T extends Colorable>(
     };
 
     return (
-      <TableCell
-        padding="none"
-        align="left"
+      <Box
```
joshxg, line 29: <td> in a <td> is a no no apparently!

packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
```diff
@@ -7,32 +7,48 @@ import { Cell } from 'react-table';
...
+    <TableRow
+      sx={{
+        alignItems: 'center',
+        height: '40px',
+        boxShadow: 'inset 0 0.5px 0 0 rgba(143, 144, 166, 0.5)',
+        padding: '0px 20px',
+        display: 'flex',
+        flex: '1 0 auto',
+        minWidth: '1000px',
+      }}
```
joshxg, line 32: Flex for the layout - boxshadow/height from zepplin

packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
```diff
@@ -7,32 +7,48 @@ import { Cell } from 'react-table';
...
           <TableCell
             key={cellKey}
             align={cell.column.align}
             sx={{
+              borderBottom: 'none',
               justifyContent: 'flex-end',
+
               padding: 0,
-              paddingLeft: '16px',
+              paddingRight: '16px',
```
joshxg, line 49: Tried to make the header/rows consistent in that the rows themselves had 20px on each side, then there was 16px right padding between each cell.. probably should have taken it off the last cell in the row but it looks ok with a checkbox..

packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
```diff
@@ -7,32 +7,48 @@ import { Cell } from 'react-table';
...
               ...(hasOnClick && { cursor: 'pointer' }),
+              ...cellStyle,
```
joshxg, line 51: this has `width: 50` etc 🤷

packages/transactions/src/OutboundShipment/ListView/ListView.tsx
```diff
@@ -80,19 +80,74 @@ export const OutboundShipmentListView: FC = () => {
...
+      align: 'right',
+    },
+    {
+      label: 'label.comment',
+      key: 'comment',
+      width: 150,
+      minWidth: 300,
+      maxWidth: 450,
+      align: 'left',
+    },
```
joshxg, line 149: I couldn't decide what to do with these.

Could put all in a `columns` folder in the `tables` folder and use `getCommentColumn` and pass params.. put them all into a file in the `transactions` package or keep them in this file, leave them here 🤔

mark-prins:
might need an `overflow: hidden` and, to make it fancy, `text-overflow: ellipsis`

I'm getting a bit of a mash up with the comments:
Screen Shot 2021-09-14 at 9 54 05 AM

But have raised #158 which takes care of big comments

Have also fixed the demo server - just noticed that.. the hook and update of the site is working well, which is great. the mock server wasn't being updated which isn't great when there's a breaking change 😉 . will add it in though, it's pretty easy to copy that too.

packages/mock-server/src/schema/Transaction.js
```diff
@@ -1,22 +1,43 @@
...
+  return options[randomIdx];
+};
+
 const TransactionData = Array.from({ length: 500 }).map((_, i) => ({
   id: `${i}`,
-  customer: `${faker.name.firstName()} ${faker.name.lastName()}`,
-  supplier: `${faker.name.firstName()} ${faker.name.lastName()}`,
-  date: faker.date.past().toString(),
-  total: `${faker.commerce.price()}`,
+  name: `${faker.name.firstName()} ${faker.name.lastName()}`,
```
mark-prins, line 13: this is great! the new fields make the list look really good, and the randomness of the status is a nice touch.

packages/transactions/src/OutboundShipment/ListView/ListView.test.tsx
```diff
@@ -10,20 +10,39 @@ import { OutboundShipmentListView } from './ListView';
...
-    const { getByText } = render(
+    const { getByRole } = render(
       <TestingProvider>
         <TestingRouter initialEntries={['/customers/customer-invoice']}>
           <Route path="*" element={<OutboundShipmentListView />} />
         </TestingRouter>
       </TestingProvider>
     );
 
+    // TODO: Don't hard core the columns. Get them from a shared function with the
```
mark-prins, line 21: :nit I prefer `hard core` but suspect it was intended to be `hard code`

packages/transactions/src/OutboundShipment/ListView/ListView.tsx
mark-prins: me too! 

leaning towards putting them in a shared file and using those column definitions to compose the array of columns here. or possibly going full mobile and having collections of columns already defined - though there has been a few issues with the inflexibility of that ( such as combinations of preferences..using cash handling and vaccines and programs )

I think there's value in having each column look the same, regardless of which specific table it is used in. such as having the 'confirmed date' always with a particular format and width. 🤷‍♂️

mark-prins (approved):
🎉  it is soo much better!

I don't have great ideas on what to do with column definitions - apart from the observation that it would be good to do something, or at least decide which way to go

joshxg:
Nice, thanks

The comments just on the one line now but think fixing it with a better comment column later will do it

image

joshxg:
also sorry about the demo server. If we enabled the mock worker in production, will that mean we can remove the mock server and everything will be updated by the hook?

mark-prins:
> also sorry about the demo server. If we enabled the mock worker in production, will that mean we can remove the mock server and everything will be updated by the hook?

all good. I've just added the line to copy the mock server src to the hosted site. haven't tested but I'm sure it'll be fine. 
Could just enable the worker, that'd also fix the problem. I figured that we wouldn't be using the mock server long so shouldn't be much of an issue 🤷‍♂️

PR #159, opened 2021-09-13
#154 condensed milk buttons

### LINKED ISSUE #154: Condensed appbar buttons for small viewports
Catchy title!

I realised that we haven't implemented the condensed buttons as per this design:
Screen Shot 2021-09-13 at 5 26 24 PM

joshxg:
Fixes #154 

image
image

- Wasn't sure if I should have made a `ShrinkableButton` rather than adding the conditions. Wanted to get your opinion on that
- Went with 1200px as that's what the current drawer is using 🤔


packages/common/src/hooks/useIsSmallScreen.tsx
```diff
@@ -0,0 +1,6 @@
+import { useMediaQuery, useTheme } from '@material-ui/core';
+
+export const useIsSmallScreen = (): boolean => {
+  const theme = useTheme();
+  return useMediaQuery(theme.breakpoints.down('lg'));
```
mark-prins, line 5: we'll need a few more options sorry. The viewports page in zeplin has these values as defaults:
- `> 1536 wide`, the detail panel visible
- `< 1536 wide`, the detail panel hidden
- `> 1440 wide` app drawer expanded
- `< 1440 wide` app drawer collapsed
- `> 1024 wide` app bar RH button label visible
- `< 1024 wide` condensed milk

whaddya say that we continue this specific pattern that you've used, and just add in `useIsLargeScreen` and `useIsMediumScreen`? would that be too restrictive?

packages/common/src/ui/components/buttons/Button.tsx
```diff
@@ -1,30 +1,59 @@
...
 import { styled } from '@material-ui/core/styles';
 import { LocaleKey, useTranslation } from '../../../intl/intlHelpers';
 import { DefaultButtonStyles } from './styles';
+import { useIsSmallScreen } from '../../../hooks/useIsSmallScreen';
 
 interface ButtonProps {
   icon: React.ReactNode;
   labelKey: LocaleKey;
   onClick: (event: React.MouseEvent<HTMLButtonElement>) => void;
+  shouldShrink?: boolean;
```
mark-prins, line 12: I'm surprised- I thought you generally favoured specific versions of components rather than conditionals?
I hate to say it - but I really like the idea of a `ShrinkableButton` 

I was thinking it might end up a switch between an `IconButton` and a `Button` - instead you've done wonders! :smiley:  the transition on the width is fantastic

mark-prins (approved):
👍  love it

packages/common/src/hooks/useIsSmallScreen.tsx
joshxg: So I went a bit OTT and added `useXL`, `useLarge`, `useMed` : e99baa3

This was because I changed the breakpoints, and this was aligning somewhat with that. Additionally in hindsight it's actually probably only a matter of time before we support mobiles, in at least some way?

I am not 100% about the hooks and using `down` from the previous breakpoint and if that is just confusing. Open to changing them again if you find it confusing

packages/common/src/ui/components/buttons/Button.tsx
joshxg: Hmm yeah I liked the idea of it also and I actually wrote it similar to this as it's own component but it was just so minimally different I couldn't justify it 🤷 

I'd be happy to change if you think it's a good idea

packages/common/src/hooks/useIsSmallScreen.tsx
joshxg: Oh I also added stories for the hooks because 1) they were helping me visualise and 2) actually stories for hooks would be useful (as if I haven't added enough things for us to do 🙄 )

joshxg:
I ended up adding a bit, so rerequesting

packages/common/src/hooks/useIsSmallScreen.tsx
mark-prins: you've done heaps! thanks - I think the hook approach is much clearer than checking breakpoint levels and media sizes. that level of granularity is available if we really need it, I think that mostly we will only use small, medium etc. The use of hooks is consistent with other usages too.

Will update the RH drawer when this PR is merged in, and can fix up the tests to add in a test for smaller screen size too now :)

packages/common/src/ui/components/buttons/Button.tsx
mark-prins: I thought there were enough differences between the two - you would have a better idea than I would though, so I'd say stick with what you have.

packages/common/src/hooks/useIsScreen/useIsMediumScreen.test.ts
```diff
@@ -0,0 +1,26 @@
...
+    const { result } = renderHook(useIsMediumScreen, {
+      wrapper: TestingProvider,
+    });
+    const { current } = result;
+
+    expect(current).toBe(true);
+  });
+
+  it('Returns false when the screen is greater than 1440', () => {
+    setScreenSize_ONLY_FOR_TESTING(1440);
```
mark-prins, line 17: :nit this feel real pedantic sorry. The test says _greater than_ 1440 and the screen size _is_ 1440

packages/common/src/hooks/useIsScreen/useIsScreen.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+IsMediumScreen.args = { hook: useIsMediumScreen, type: 'medium' };
+
+export const IsLargeScreen = Template.bind({});
+IsLargeScreen.args = { hook: useIsLargeScreen, type: 'large' };
+
+export const IsExtraLargeScreen = Template.bind({});
+IsExtraLargeScreen.args = { hook: useIsExtraLargeScreen, type: 'xl' };
+
+export const IsSmallScreen = Template.bind({});
+IsSmallScreen.args = { hook: useIsSmallScreen, type: 'sm' };
```
mark-prins, line 35: another :nit - I like the screen sizes to go in logical order
```suggestion
export const IsExtraLargeScreen = Template.bind({});
IsExtraLargeScreen.args = { hook: useIsExtraLargeScreen, type: 'xl' };

export const IsLargeScreen = Template.bind({});
IsLargeScreen.args = { hook: useIsLargeScreen, type: 'large' };

export const IsMediumScreen = Template.bind({});
IsMediumScreen.args = { hook: useIsMediumScreen, type: 'medium' };

export const IsSmallScreen = Template.bind({});
IsSmallScreen.args = { hook: useIsSmallScreen, type: 'sm' };
```

mark-prins (approved):
that is so good!!
really tidies up the responsive behaviour and makes it much easier to implement

joshxg:
Thanks! Will try and sneak the `ShrinkableButton` into another PR I think you're right

packages/common/src/hooks/useIsScreen/useIsMediumScreen.test.ts
joshxg: 33dfd46 😁

Issue #49, opened 2021-08-16
Basic detail page RHS drawer


joshxg:
A generic solution for this drawer: 

![image](https://user-images.githubusercontent.com/35858975/129508345-84b920a1-c768-462c-8689-ec9a2a3e15ac.png)

- Not adding the specific details, yet, but a generic solution for adding a drawer to a detail page

PR #161, opened 2021-09-14
#49 basic detail page rhs drawer

mark-prins:
Partial fix for #49. 
Was going to continue with 
- tests
- stories
- actions ( bottom of panel )

and thought that this was already too large, which is why this PR.

Have gone with a state based approach rather than a portal because I wanted to keep some control of layout; the calling component provides the sections ( and actions ) to display and the detail panel lays them out.

I've also gone off script with regard to the layout of the main components - abandoning material `Drawer` as I find the fixed positioning problematic. Changing to a flex layout has meant that we don't need to keep the drawer `expanded` state globally, as flex can cope with the layout rather than using margins and widths.

The main motivation for doing that though, is that I didn't like the way the detail panel is shown in the mock ups as going over buttons and other page elements ( like the material temporary drawer does ). This means that the panel can't really be left open - in the mockup the OK button, bottom right, is covered by the detail panel. I think that given how much we are planning to put into the detail panel, it needs to be a little more accessible, so I've made it a part of the flex layout. If the screen is large enough, it can stay open all the time.

Can easily change things back if you prefer though - and have the detail panel show over the top of stuff on the page.


joshxg (approved):
Beautiful!

I really like the drawer not overlapping - I think it's heaps nicer. Just hoping there's plenty of room for columns!

I am not sure about the state v portal approach .. both seem to have good benefits. If the portal was where `props.children` was in `DetailPanel` (and the children were a "`Section`" component, is the layout the same?

PR #162, opened 2021-09-15
#49 rhs drawer tests

### LINKED ISSUE #49: Basic detail page RHS drawer

mark-prins:
Fixes #49 

Finishes this off - adding in tests and a story and includes the actions section.

Had a few problems getting storybook to work within the host app, was getting

```
Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.
```

and working on advice from [this issue](https://github.com/storybookjs/storybook/issues/9516) ended up making a big 'ol mess of the imports in order to get it working. Not entirely sure that it is worth it, but it is done, and the story is nice.


packages/common/src/styles/theme.ts
```diff
@@ -52,6 +58,11 @@ const themeOptions = {
       menu: '#f2f2f5',
       white: '#fff',
     },
+
+    form: {
+      field: '#555770',
```
mark-prins, line 63: not entirely convinced that this was a good idea

packages/common/src/ui/layout/forms/Field.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
```
mark-prins, line 1: the `Field` and `Label` seemed like a good idea. Not sure how useful they actually are.

packages/host/src/DetailPanel.tsx
```diff
@@ -1,22 +1,20 @@
...
-  ChevronDown,
-  useTheme,
-  useMediaQuery,
-} from '@openmsupply-client/common';
-import {
-  Accordion,
-  AccordionDetails,
-  AccordionSummary,
-} from '@material-ui/core';
+import { Box } from '@openmsupply-client/common/src/ui/layout';
```
mark-prins, line 2: this blows

packages/host/src/DetailPanel.tsx
```diff
@@ -72,16 +70,16 @@ const StyledAccordian = styled(Accordion)(({ theme }) => ({
...
   const theme = useTheme();
   const isSmallScreen = useMediaQuery(theme.breakpoints.down('lg'));
 
   const Sections = useCallback(
     () => (
       <>
-        {sections.map(section => (
-          <Box key={section.titleKey}>
+        {sections.map((section, index) => (
+          <Box key={`action.titleKey_${index}`}>
```
mark-prins, line 82: a little dodgy - entirely so that we can have duplicate section titles without having to manually specify a key

packages/host/src/DetailPanel.tsx
```diff
@@ -98,31 +96,68 @@ const DetailPanel: React.FC = () => {
     [sections]
   );
 
+  const Action = styled(Link)({
+    color: '#e63535',
```
mark-prins, line 100: should add to theme? 🤔

packages/host/src/DetailPanel.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  it('Does not render empty panel', () => {
+    const actions: Action[] = [];
+    const sections: Section[] = [];
+    const { queryByText, queryByTestId } = render(
+      <DetailPanelExample actions={actions} sections={sections} />
+    );
+
+    act(() => {
+      queryByText('More')?.click();
```
joshxg, line 38: 🤔  the screen by default is large enough that the drawer is automatically open. So the `queryByText`'s always can't find anything and the click doesn't happen - are they meant to be ensuring that if the test environment screen changes the test still passes?

Also the drawer is always rendered except when actions/sections aren't passed, it just has nothing in it so is hidden - `toBeVisible` doesn't seem to work .. best I came up with was `toHaveAttribute("aria-expanded, "false")`

packages/host/src/DetailPanel.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+    ];
+    const { queryAllByText, queryByText } = render(
+      <DetailPanelExample actions={actions} sections={sections} />
+    );
+
+    act(() => {
+      queryByText('More')?.click();
+    });
+
+    expect(queryAllByText('Comment')).toHaveLength(3);
```
joshxg, line 91: Maybe also useful to check that we're rendering the correct sections?

```
it('Does render the correct sections', () => {
    // ensuring the drawer is closed for the test
    setScreenSize_ONLY_FOR_TESTING(1000);
    const actions: Action[] = [];
    const sections: Section[] = [
      { titleKey: 'heading.comment', children: [<span key="0" />] },
      { titleKey: 'heading.additional-info', children: [<span key="0" />] },
    ];
    const { getByRole } = render(
      <DetailPanelExample actions={actions} sections={sections} />
    );

    act(() => {
      getByRole('button', { name: /more/i }).click();
    });

    const section1 = getByRole('button', { name: /comment/i });
    const section2 = getByRole('button', { name: /additional info/i });

    expect(section1).toBeInTheDocument();
    expect(section2).toBeInTheDocument();
  });
```

Also could check that when a section is expanded it's content is shown

```
it('renders the children of a section once expanded', () => {
    setScreenSize_ONLY_FOR_TESTING(1000);
    const actions: Action[] = [];
    const sections: Section[] = [
      {
        titleKey: 'heading.comment',
        children: [<span data-testid="child1" key="0" />],
      },
      {
        titleKey: 'heading.additional-info',
        children: [<span data-testid="child2" key="0" />],
      },
    ];
    const { getByRole } = render(
      <DetailPanelExample actions={actions} sections={sections} />
    );

    act(() => {
      getByRole('button', { name: /more/i }).click();
    });

    const section2 = getByRole('button', { name: /additional info/i });

    act(() => {
      section2.click();
    });

    expect(screen.getByTestId('child2')).toBeVisible();
  });
```

packages/host/src/DetailPanel.tsx
joshxg: sometimes you gotta use an index 🤷

packages/host/src/DetailPanel.tsx
joshxg: why an `a` over some sort of `button`?

packages/common/src/ui/layout/forms/Field.tsx
joshxg: Seem useful to me.. although I am confused by the `ui/layout` folder

packages/common/src/styles/theme.ts
joshxg: Not sure how much these (field/labels) will be used. You'd think a lot, so probably is!

packages/host/src/DetailPanel.tsx
joshxg: 😭

packages/transactions/src/OutboundShipment/DetailView/DetailView.tsx
```diff
@@ -65,22 +73,76 @@ export const OutboundShipmentDetailView: FC = () => {
...
+  const { OpenButton, setActions, setSections } = useDetailPanel();
+  const t = useTranslation();
+  const d = useFormatDate();
+  const { success, warning } = useNotification();
+  const entered = draft?.entered ? d(new Date(draft.entered)) : '-';
+
+  const copyToClipboard = () => {
+    navigator.clipboard.writeText(JSON.stringify(draft, null, 4) ?? '');
+    success('Copied to clipboard successfully')();
+  };
```
joshxg, line 85: ohhh nice!

joshxg (approved):
really coming together, badass! 🎉

packages/common/src/hooks/useDetailPanel.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+  });
+
+  it('Does not render an open button if open', () => {
+    const { queryByText } = render(<DetailPanelExample />);
+
+    act(() => {
+      queryByText('More')?.click();
+    });
+
+    expect(queryByText('More')).not.toBeInTheDocument();
```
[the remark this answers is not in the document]
mark-prins: ah. thanks - that's what I wanted to do; had only the button role before and that wasn't strict enough I thought.
d2bac08

packages/common/src/ui/layout/forms/Field.tsx
mark-prins: good point. I was working with layout only but the form is likely to have input components too. would it work for you if I lifted it up a level and had `ui/forms` instead?

joshxg: Sure. When I think of layout, I suppose I think of things like the `row` component or `grid`/`box`/`column` 🤔

packages/host/src/DetailPanel.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+    const { queryByText, queryByTestId } = render(
+      <DetailPanelExample actions={actions} sections={sections} />
+    );
+
+    act(() => {
+      queryByText('More')?.click();
+    });
+
+    expect(queryByTestId('detail-panel')).not.toBeInTheDocument();
+  });
```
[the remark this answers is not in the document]
mark-prins: done, thanks

packages/host/src/DetailPanel.test.tsx
mark-prins: very nice, ta. have added

mark-prins:
@joshxg I've re-requested the review, since there were a number of changes. 

Mostly they are changes that you suggested, there were just two things to note:
- the later designs ( not yet in zeplin ) have icons on the actions, so have added them in and converted to buttons rather than links
- the reason for not rendering the detail panel when not populated with actions or sections, is so that it only shows on pages when requested. could make this more explicit if you like but I thought in this case having it implicit would be ok. have hidden the 'More' button where there is no contents in the panel to prevent the weirdness you noted.

joshxg (approved):
Nice

After seeing the actions and the changes to the shape, it seems that it could all just be done with components.. we could easily make a `<DetailPanelAction/>` component and pass that. Similar for the `Section`.. a `DetailPanelSection`, `DetailPanelHeading`, `DetailPanelContent`..

Then it occurs to me that with that you may as well use a portal 🙃

Issue #157, opened 2021-09-13
Outbound shipment: Pagination footer


joshxg:
- Currently when there are too many rows, the pagination gets pushed off the page
- The pagination also looks a bit different to what we have showing

![image](https://user-images.githubusercontent.com/35858975/133153009-a34bce14-181c-499e-b7fa-fed6628ace3e.png)

- We are using a select for the number of rows - it's cool, but it's not in the design?
- Seems the pagination is stuck to the bottom and the number of rows is fixed. Should the table body be scrollable? (just the rows?) Or should we determine the number of rows for the users screen height

mark-prins:
thanks - I had a note to revisit the pagination design. was going to implement as per the mock ups.
the content area is scrollable - the table body isn't though. It could be, like what we did with universal codes: I found that problematic though, in that there is a very limited space for the table results. Willing to have another try..

so we could do this:
- fix the max height of the table, so that the whole thing (including pagination) always shows 
- make the body of the table scrollable

but we could match the design better by doing this:
- calculate the number of rows to show using the screen height
- don't allow the user to select rows per page

I'm up for it - what do you think?

joshxg:
For the list views I think non-scrollable and calculate the number of rows is good.

I am wondering if it would be annoying for a user when they're entering values and tabbing to the next cell that there was a small number of rows on a page. 

I'm not really sure the average number of rows on an invoice/requisition to know if this is practically an issue. For a prescription definitely not, but for store to store orders/invoices?

Could likely implement some sort of 'if tabbed on the last cell on a page go to the next page' thing, so it should be all good, though!

mark-prins:
yeah, good point! am wondering about not having pagination on the detail view, will have to run some queries to see what the limits are for number of items per transaction  & requisition

PR #164, opened 2021-09-15
#157 pagination footer 2

joshxg:
Another part of #157 

This is the hooks I ended up making.. made them somewhat generic because i'm sure they'll be useful later

- `useWindowDimensions`: Using this to determine the height of the viewport that will trigger a state change and re-render rather than just using `window.innerHeight`. Potentially overkill
- `useBoundingClientRect` : Using this to track the height of the `AppBar`. When I made the table flex, if I rendered too many rows, they would overlap the `AppBar`
- `useAppBarRect` : Meant to make tracking the `AppBar` height easier for peeps. I was undecided here if 1) it was needed 2) it should be in the `useHostContext` 3) if there should just bee an `AppBar` store where we put the other things the `AppBar` is responsible for there (the refs for portals). Another potential improvement/different approach would be to use a `Context` in `Host`, from which the `AppBar` can update the value once rendered and all children of `Content` (a sibling of `AppBar`) can read the context values.


packages/common/src/hooks/useAppBarRect/useAppBarRect.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+    };
+
+    expect(state).toEqual(expected);
+  });
+
+  it('Triggers a new state update when the window dimensions change', () => {
+    const { result } = renderHook(useAppBarRect);
+
+    expect(result.all.length).toBe(2);
+  });
```
joshxg, line 46: these tests are all pretty lame as they have to deal with the dom dimensions and there is no layout engine .. tried to make them offer *something* and not give false negatives, but not really offering much

mark-prins (approved):
that's a lot of hooks! 😄

PR #165, opened 2021-09-15
#157 pagination footer

### LINKED ISSUE #157: Outbound shipment: Pagination footer

joshxg:
Fixes #157 - sigh!

image

- Added flex to the table body, which meant I needed to add `display: 'flex'` to all the parents?
- Extracted pagination into a `PaginationRow` component
- After creating a hook to put the query params and things to help calculate the number of rows it seemed easier to just put all the query props in the one place


packages/common/src/styles/RTLProvider.tsx
```diff
@@ -4,5 +4,16 @@ import { useRtl } from '../intl/intlHelpers';
...
+      style={{
+        height: '100vh',
+        width: '100vw',
+        display: 'flex',
+        flexDirection: 'column',
+        direction: isRtl ? 'rtl' : 'ltr',
+      }}
+      {...props}
+    />
+  );
```
joshxg, line 18: I think this was silly. I forgot that chucking a `div` in here would mess things up and it would be easier to add the `direction` straight onto the `body`

packages/common/src/ui/layout/tables/RemoteDataTable.tsx
```diff
@@ -19,79 +17,43 @@ import {
...
   data = [],
-  initialSortBy,
   isLoading = false,
-  onFetchData,
+  onSortBy,
   onRowClick,
-  totalLength = 0,
+  pagination,
   tableApi,
+  onChangePage,
```
joshxg, line 41: I just pass the `sortBy` instead of `initialSortBy`. We were keeping the current `sortBy` in the `ListView` anyway. Then just needed a `onSortBy` to remove the `useSortBy` hook which doesn't really isn't doing anything..

Then .. (I actually implemented this by re-fetch the data by the number of rows to render rather than just slicing the data [ uhh.. 🙄]) I extracted all the query props into the one hook which could be a bit tidier but will do with the `useListData` hook later

packages/common/src/ui/layout/tables/RemoteDataTable.tsx
```diff
@@ -106,8 +68,17 @@ export const RemoteDataTable = <T extends Record<string, unknown>>({
...
+      sx={{
+        display: 'flex',
+        flexDirection: 'column',
+      }}
+    >
+      <MuiTable
+        stickyHeader
+        {...getTableProps()}
+        sx={{ flex: 1, display: 'flex', flexDirection: 'column' }}
+      >
```
joshxg, line 81: bunch of styles for flexing?

packages/common/src/ui/layout/tables/RemoteDataTable.tsx
```diff
@@ -138,17 +117,13 @@ export const RemoteDataTable = <T extends Record<string, unknown>>({
...
-                        : false
+                      sortingRule ? (sortingRule?.desc ? 'desc' : 'asc') : false
                     }
                   >
                     <TableSortLabel
                       hideSortIcon={column.id === 'selection'}
-                      active={column.isSorted}
-                      direction={column.isSortedDesc ? 'desc' : 'asc'}
+                      active={!!sortingRule}
+                      direction={sortingRule?.desc ? 'desc' : 'asc'}
```
joshxg, line 126: Could all be cleaner. Will do when extracting to a `HeaderRow`

packages/common/src/ui/layout/tables/RemoteDataTable.tsx
```diff
@@ -160,7 +135,15 @@ export const RemoteDataTable = <T extends Record<string, unknown>>({
...
         </TableHead>
-        <TableBody {...getTableBodyProps()}>
+        <TableBody
+          {...getTableBodyProps()}
+          sx={{
+            display: 'flex',
+            flexDirection: 'column',
+            flexGrow: 1,
+            overflow: 'hidden',
+          }}
```
joshxg, line 145: ends up working pretty well.. still wonder if overflow needs to be hidden 🤔

packages/common/src/ui/layout/tables/columns/PaginationRow/PaginationRow.tsx
```diff
@@ -0,0 +1,53 @@
...
+        <>
+          <Box display="flex" flexDirection="row">
+            <Typography sx={{ marginRight: '4px' }}>Showing</Typography>
+            <Typography sx={{ fontWeight: 'bold', marginRight: '4px' }}>
+              {xToY}
+            </Typography>
+            <Typography sx={{ marginRight: '4px' }}>of</Typography>
+            <Typography sx={{ fontWeight: 'bold', marginRight: '4px' }}>
+              {total}
+            </Typography>
```
joshxg, line 41: Wondering if theres a nicer way to do this

packages/common/src/ui/layout/tables/columns/PaginationRow/PaginationRow.tsx
```diff
@@ -0,0 +1,53 @@
...
+          </Box>
+          <Pagination
+            count={Math.ceil(total / first)}
+            onChange={(_, value) => {
+              onChange(value - 1);
```
joshxg, line 46: this is specifically passing back `0` for `page 1` - I think it was helpful but could be too confusing

packages/transactions/src/OutboundShipment/ListView/ListView.tsx
```diff
@@ -28,21 +27,61 @@ import {
...
-  const [queryProps, setQueryProps] = useState<QueryProps<Transaction>>({
-    first: 10,
-    offset: 0,
-  });
   const { data: response, isLoading } = useQuery(
-    ['transaction', 'list', queryProps],
-    () => listQueryFn(queryProps)
+    ['transaction', 'list', { first, offset, sortBy }],
+    () => listQueryFn({ first, offset, sortBy }),
+    { keepPreviousData: true }
```
joshxg, line 84: this is stopping `loading` become true and instead just replaces the `data` with the new data which can make it feel a lot nicer, but could feel a lot worse if not on a LAN!

mark-prins (approved):
taking boss level up another notch!! that is incredible

I was a little nervous that being so crafty might make it brittle; it's working fine for me though, and is certainly much nicer to use. totally worth it.

there's a couple of minors which you might want to ignore:
- if I have the table on a small screen and resize it on a larger one then the table rows end up too high, rather than the data refetching and re rendering the table with a different number of rows.
- the loading spinner is now showing top left rather than centred

and one that I cannot reproduce: I was on page 5, clicked the next button and the table showed me the results of page 6 but had highlighted page 1. Seems ok now, so safe to ignore I think.

joshxg:
Ah yes of course, small -> big! I've added a change to refetch in that case 
e74cd52

Also did the changes for the callback and test updates 11cf7c2

PR #174, opened 2021-09-17
#173 add items to mocks

### LINKED ISSUE #173: Add items to mock transactions
Was thinking to tidy up the outbound shipment page, and as part of that was going to revising using the data table in the detail view, which is to use local data rather than fetching the data.

First step was to add some items to the transaction in order to have something to show.

mark-prins:
Fixes #173 

Just a little mockery for the items

### DIFF
--- a/packages/common/src/types/index.ts
+++ b/packages/common/src/types/index.ts
@@ -1,3 +1,10 @@
+export type Item = {
+  id: string;
+  code: string;
+  name: string;
+  packSize?: number;
+  quantity?: number;
+};
 export type Transaction = {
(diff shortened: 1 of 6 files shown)


joshxg (approved):
awesome thanks! So this is like a trans line item?

Also is the demo server using the mock server? If it's not maybe we should think about removing it 🤔  nice to have something to run actual graphql queries against though

mark-prins:
> So this is like a trans line item?

yeah, that's what I was thinking.

> is the demo server using the mock server?

Yes, it is. will replace at some point..

PR #177, opened 2021-09-18
# 146 usePagination hook

joshxg:
So I have been working on `useListData` and it's not quite done but it's already 1000 lines. Thought I'd take some commits and PR them individually... 🙃 

I broke this all down into small hooks and then each one has a test and storybook which is why the big diff .. 🙄 

- `usePagination`
- `useSortBy`
- `useRowRenderCount`
- `useQueryParams` (Just taking all three and combining them. Will also have `useFilters`, probably, later)

Then `useListData` is a combo of `useQueryParams`, `useQuery` and `useMutation`


packages/common/src/hooks/useListViewData/usePagination.ts
```diff
@@ -0,0 +1,21 @@
...
+  const [first, onChangeFirst] = useState(initialFirst);
+  const [offset, onChangeOffset] = useState(0);
+  const [page, onChangePage] = useState(0);
+
+  useEffect(() => {
+    onChangeOffset(page * first);
+  }, [first, page]);
+
+  return { page, onChangePage, onChangeFirst, offset, first };
+};
```
joshxg, line 21: Played with using `useReducer` but seemed overkill with all the boilerplate you need.

packages/common/src/hooks/usePagination/usePagination.ts
```diff
@@ -0,0 +1,21 @@
...
+interface PaginationState {
+  page: number;
+  offset: number;
+  first: number;
+  onChangePage: (newPage: number) => void;
+  onChangeFirst: (newFirst: number) => void;
+}
+
+export const usePagination = (initialFirst = 20): PaginationState => {
+  const [first, onChangeFirst] = useState(initialFirst);
```
joshxg, line 12: Possibly need some defensive code to ensure `first` is > 0

joshxg:
Ugh not sure why these diffs are showing your work thats in main - sorry!

joshxg:
Ah nuts I just made it worse

joshxg:
OK-  better.

🥇

packages/common/src/hooks/usePagination/usePagination.ts
mark-prins: I can also change the `page` and `first` to negative numbers, I guess ideally we'd prevent some of that.

mark-prins (approved):
👍  great! 
I'm thinking that I should start calling you captain hook :) ☠️

packages/common/src/hooks/usePagination/usePagination.ts
joshxg: Ah yes! Do you think `if < valid number` throw an error or default to the minimum?

mark-prins: I was thinking to default to a minimum - but now that you say it, I rather like the idea of throwing an error. Otherwise we'll just carry a silent bug around when someone accidentally goes negative with one of those values.

Issue #146, opened 2021-09-11
Outbound shipment: useListData hook


joshxg:
- A hook `useListData`
- Handles what we need for `ListView` style pages - some actions and getting the 'list data'
- Because some actions will have loading states and error states as well as the query itself, and we can consolidate all that nicely in one hook also list views will all be using sorting/pagination which we can also encapsulate/shove in a single hook nicely/in a way where we can forget about it.
- Also a lot of the actions will be similar on all pages - i.e. `delete`. 

Unsure about:
- Actions which don't require server interaction: export to csv (I'm guessing we'll do this client side, but as an example) even though it is something that might be on a lot of pages.
- Means this is like `useListDataAndTalkToTheServerForMe` hook so there might be a better name

mark-prins:
great idea. 😄 
A suggestion, if we do need to have a local data version too: `useFetchListData` and `useListData`

PR #178, opened 2021-09-18
#146 useSortBy hook

joshxg:


packages/common/src/hooks/useListViewData/useSortBy.ts
```diff
@@ -0,0 +1,27 @@
...
+    setSortBy(({ key: prevSortKey, isDesc: prevIsDesc }) => {
+      const newIsDesc = prevSortKey === newSortKey ? !prevIsDesc : false;
+      return { key: newSortKey, isDesc: newIsDesc };
+    });
+  }, []);
+
+  return {
+    sortBy: { key, isDesc, direction: isDesc ? 'desc' : 'asc' },
+    onChangeSortBy,
+  };
```
joshxg, line 26: passing `direction` is because some mui stuff wants `'desc'` or `'asc'` so just did it here.. why not 🤷 

I change this hook a little later to accept just a `keyof T` as the `initialSortBy`. This was silly and I will revert it back to using this implementation of passing a `SortRule`. Also might need to change the `onChangeSortBy` to accept a `SortRule`? Though, it is nice to just be able to pass a string, so .. maybe both?

PR #180, opened 2021-09-18
#146 useListData hook

joshxg:
#146 

- Still some tests to go
- Still want to make a `useOptimisticMutation` hook: something similar to https://github.com/openmsupply/msupply-dashboard-app/blob/master/frontend/src/hooks/useOptimisticMutation.ts - that one got a bit crazy and wasn't a great abstraction so want to improve on it..
- Move all the hooks out of the same folder


packages/common/src/hooks/useListData/useListData.ts
```diff
@@ -0,0 +1,57 @@
...
+    offset,
+    sortBy,
+  }: {
+    first: number;
+    offset: number;
+    sortBy: SortRule<T>;
+  }) => () => Promise<{ data: T[]; totalLength: number }>;
+  onDelete: (toDelete: T[]) => Promise<void>;
+  onUpdate: (toUpdate: T) => Promise<T>;
+}
```
joshxg, line 17: Not a great name and this idea needs iterating on.. the idea is to have an object for each list view which has an `onQuery`, `onDelete` and `onUpdate`, possibly also an `onCreate`

packages/common/src/hooks/useListData/useListData.ts
```diff
@@ -0,0 +1,57 @@
...
+    api.onDelete,
+    { onSuccess: invalidation }
+  );
+
+  const { mutateAsync: onUpdate, isLoading: isUpdateLoading } = useMutation(
+    api.onDelete,
+    { onSuccess: invalidation }
+  );
+
+  return {
```
joshxg, line 45: is super simple right now but by just defining a few graphql queries in an `api` object, gives a lot of functionality.

- `useMutation` -> `useOptimisticMutation`, probably
- error handling
- Likely need to return some more fields from the mutations
- Trying to think of a way where we can also call `useListData` with the same QueryKey, in any random component and get all the same data like you already can with `useQuery`, but this might need a separate hook 🤔 

Also trying to be conscious that putting too much here is going to make it very confusing and hard to maintain.. that is why, I suppose, I have tried to break it down a lot..

packages/common/src/hooks/useListData/useQueryParams.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+      expect.objectContaining({
+        sortBy: { key: 'id', isDesc: false, direction: 'asc' },
+        pagination: expect.objectContaining({ page: 0, offset: 0, first: 50 }),
+        page: 0,
+        offset: 0,
+        first: 50,
+      })
+    );
+  });
+});
```
joshxg, line 89: The added functionality of this composition of hooks is that when `numberOfRows` changes, `first` changes, if number of rows is greater than first... so that's all I was testing, really 🤔

packages/common/src/hooks/useListData/useQueryParams.ts
```diff
@@ -0,0 +1,27 @@
...
+  const pagination = usePagination(numberOfRows);
+  const { sortBy, onChangeSortBy } = useSortBy(initialSortBy);
+
+  useEffect(() => {
+    pagination.onChangeFirst(numberOfRows);
+  }, [numberOfRows]);
+
+  const queryParams = { ...pagination, pagination, sortBy, onChangeSortBy };
+
+  return { ...pagination, pagination, sortBy, onChangeSortBy, queryParams };
```
joshxg, line 26: bit of a weird return.. I just really hate it when I can't destructure one value but also keep the whole thing as a chunk..!

packages/common/src/hooks/useListData/useSortBy.test.ts
```diff
@@ -53,9 +47,7 @@ describe('useSortBy', () => {
   });
 
   it('has the correct values after triggering a few sort bys in sequence', () => {
-    const { result } = renderHook(() =>
-      useSortBy<TestSortBy>({ key: 'id', isDesc: false })
-    );
+    const { result } = renderHook(() => useSortBy<TestSortBy>('id'));
```
joshxg, line 50: this is just the initial sort by so I thought that just passing the field was a good idea. I'm now realising it isn't, and that you will want to be able to specify descending or ascending differently for different columns

packages/common/src/hooks/useListData/useSortBy.ts
```diff
@@ -0,0 +1,39 @@
+import { useCallback, useState } from 'react';
```
joshxg, line 1: seems i'm messing this diff up. This has a change described above, where I thought it would be good to just pass a key. Will need to change this back to pass {key, isDesc}

packages/common/src/ui/layout/tables/hooks/useDataTableApi.ts
```diff
@@ -16,7 +16,7 @@ export const useSetupDataTableApi = <DataTableRowType extends object>(
       // Mapping into a plain row object rather than a react-table row object
       // which contains a bunch of irrelevant data.
       selectedRows: selectedFlatRows.map(
-        ({ values }) => values
+        ({ original }) => original
```
joshxg, line 19: oops

PR #183, opened 2021-09-19
#146 use row render count

joshxg:


packages/common/src/hooks/useRowRenderCount/useRowRenderCount.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+          ? theme.palette.primary.main
+          : theme.palette.secondary.main;
+        return (
+          <Box key={i} bgcolor={bg} height={table.dataRow.height}>
+            Row #{i}
+          </Box>
+        );
+      })}
+
+      <Box bgcolor={theme.palette.darkGrey} height={table.paginationRow.height}>
```
mark-prins, line 41: I'm getting half the footer cut off, due to the 2rems of padding in storybook.
not a biggie - it looks nicer if you reduce the available window height for storybook only by 40px and I've spent a while playing around with setting a ref on the container and using `clientHeight` only to come the decision that it isn't worth playing around with.

mark-prins (approved):
Nice work captain! ☠️ 

one minor recommendation is all

PR #184, opened 2021-09-19
#146 use list data hook

joshxg:
#146 

- Still some tests to go
- Still want to make a `useOptimisticMutation` hook: something similar to https://github.com/openmsupply/msupply-dashboard-app/blob/master/frontend/src/hooks/useOptimisticMutation.ts - that one got a bit crazy and wasn't a great abstraction so want to improve on it..
- Move all the hooks out of the same folder


packages/common/src/hooks/useListData/useListData.ts
```diff
@@ -0,0 +1,75 @@
...
+    offset,
+    sortBy,
+  }: {
+    first: number;
+    offset: number;
+    sortBy: SortBy<T>;
+  }) => () => Promise<{ data: T[]; totalLength: number }>;
+  onDelete: (toDelete: T[]) => Promise<void>;
+  onUpdate: (toUpdate: T) => Promise<T>;
+}
```
joshxg, line 19: Not a great name and this idea needs iterating on.. the idea is to have an object for each list view which has an onQuery, onDelete and onUpdate, possibly also an onCreate

packages/common/src/hooks/useListData/useListData.ts
```diff
@@ -0,0 +1,75 @@
...
+    api.onDelete,
+    { onSuccess: invalidation }
+  );
+
+  const { mutateAsync: onUpdate, isLoading: isUpdateLoading } = useMutation(
+    api.onUpdate,
+    { onSuccess: invalidation }
+  );
+
+  return {
```
joshxg, line 62: is super simple right now but by just defining a few graphql queries in an `api` object, gives a lot of functionality.

- `useMutation` -> `useOptimisticMutation`, probably
- error handling
- Likely need to return some more fields from the mutations
- Trying to think of a way where we can also call `useListData` with the same QueryKey, in any random component and get all the same data like you already can with `useQuery`, but this might need a separate hook 🤔 

Also trying to be conscious that putting too much here is going to make it very confusing and hard to maintain.. that is why, I suppose, I have tried to break it down a lot..

packages/common/src/ui/layout/tables/hooks/useDataTableApi.ts
```diff
@@ -16,7 +16,7 @@ export const useSetupDataTableApi = <DataTableRowType extends object>(
       // Mapping into a plain row object rather than a react-table row object
       // which contains a bunch of irrelevant data.
       selectedRows: selectedFlatRows.map(
-        ({ values }) => values
+        ({ original }) => original
```
joshxg, line 19: oops

packages/transactions/src/OutboundShipment/ListView/ListView.tsx
```diff
@@ -1,131 +1,57 @@
-import React, { FC, useEffect, useState } from 'react';
+import React, { FC } from 'react';
```
mark-prins, line 1: an extra thing in this file - can you make the first column (name) sortable?
currently has `sortable` as false, from when it was the id column.
there's also something up when sortable is false - the column has the sort hover, so it looks like you can sort it, but nothing happens.

mark-prins (approved):
👍  That's very neat! Pulls everything together really nicely :)

PR #185, opened 2021-09-19
#181 general tab table

### LINKED ISSUE #181: Outbound detail: add table
Add the data table to outbound detail view

Will start with just a simple read only view with mock data of random lines added. Not sure what zeplin design I should work from. Going to from this 

image

Which is just the same table component as the list view.

Unsure how editing interactions are working 🤔

joshxg:
Fixes #181 

Just a basic table into the detail general tab

image


packages/common/src/hooks/useListData/useListData.ts
```diff
@@ -1,10 +1,11 @@
+import { ObjectWithStringKeys } from './../../types/utility';
```
joshxg, line 1: bunch of these to work with strings as keys

packages/common/src/hooks/useListData/useListData.ts
```diff
@@ -1,10 +1,11 @@
...
 import { SortBy } from './../useSortBy/useSortBy';
 import { UseMutateFunction } from 'react-query';
 import { useQueryClient, useMutation, useQuery } from 'react-query';
 import { QueryParams, useQueryParams } from '../useQueryParams';
 import { SortRule } from '../useSortBy';
 
-export interface ListApi<T> {
+export interface ListApi<T extends ObjectWithStringKeys> {
```
joshxg, line 8: ive used T a bunch and I don't actually like it but I also wasn't sure if `SomeEntity` or some other name was more useful 🤷

packages/common/src/hooks/useSortBy/useSortBy.ts
```diff
@@ -1,40 +1,51 @@
...
+
+      newSortBy = {
+        key: newSortKey,
+        isDesc: newIsDesc,
+      };
+
+      return newSortBy;
     });
+
+    return { ...newSortBy, direction: getDirection(!!newSortBy?.isDesc) };
```
joshxg, line 44: This change is to return the calculated `SortBy` object when calling `onChangeSortBy`, which comes into play in `useSortedData`..

packages/common/src/hooks/useSortedData/useSortedData.ts
```diff
@@ -0,0 +1,56 @@
...
+    const sorter = getDataSorter(newSortBy.key, !!newSortBy.isDesc);
+    setSortedData(data.sort(sorter));
+  };
+
+  useEffect(() => {
+    const sorter = getDataSorter(sortBy.key, !!sortBy.isDesc);
+    setSortedData(data.sort(sorter));
+  }, []);
+
+  return { sortedData, sortBy, onChangeSortBy: wrapped };
```
joshxg, line 55: This is a bit lame.

I started with just using `onChangeSortBy`, which changes `sortBy`, and then putting the `sortBy` in the `useEffect`.. so whenever the `sortBy` changes, resort the data. This was easy and it was nice. But it causes two renders.. one to set the `sortBy` and another to change the `data`, so I went with wrapping the `onChangeSortBy` and doing both in the same callback (which react then batches)

packages/common/src/types/utility.ts
```diff
@@ -0,0 +1,6 @@
+// This type ensures that the resulting value of a `keyof` is
+// a string.
+// eslint-disable-next-line @typescript-eslint/ban-types
+export type KeyOf<T extends object> = Extract<keyof T, string>;
```
joshxg, line 4: doing `keyof T` where `T extends Record<string,unknown>` still wasn't giving me a string key

packages/common/src/ui/layout/tables/RemoteDataTable.tsx
```diff
@@ -99,7 +105,8 @@ export const RemoteDataTable = <T extends Record<string, unknown>>({
                   <TableCell
                     {...column.getHeaderProps()}
                     onClick={() =>
-                      !column.disableSortBy && onSortBy({ key: column.id })
+                      !column.disableSortBy &&
+                      onSortBy({ key: column.id as KeyOf<T> })
```
joshxg, line 109: the `column.id` is IdType` 🤷

packages/common/src/hooks/useSortedData/useSortedData.ts
joshxg: Also need to determine the actual data type from the column, to sort by. At the moment we're just inferring the sort, but that's not going to work for long.

I had a play around with adding a `defaultSortDirection` to the column definition, and then when a trigger is sorted, pass the whole column to the call back.

This worked well but was a bit too much to add into this PR

packages/common/src/hooks/usePagination/usePagination.ts
```diff
@@ -1,12 +1,15 @@
...
 
-export interface PaginationState {
+interface Pagination {
   page: number;
   offset: number;
   first: number;
   onChangePage: (newPage: number) => void;
   onChangeFirst: (newFirst: number) => void;
 }
+export interface PaginationState extends Pagination {
```
mark-prins, line 10: I'm not sure why the addition of the `pagination` object, as well as using the individual member properties.
There's a good reason there somewhere, I'm just not seeing it. can you help me out?

packages/common/src/hooks/useSortedData/useSortedData.ts
mark-prins: good spotting of the double render. I had thought to add the `defaultSortDirection` and something for the datatype to the column definition. yeah, next PR for that.

packages/transactions/src/OutboundShipment/DetailView/DetailView.test.tsx
```diff
@@ -26,37 +26,37 @@ describe('OutboundShipmentDetailView', () => {
...
+    await waitFor(() => {
+      const allTabPanels = queryAllByRole('tabpanel');
 
-    expect(allTabPanels.length).toEqual(1);
+      expect(allTabPanels.length).toEqual(1);
+    });
   });
 
-  it('Renders the transport details content once the price tab has been pressed', async () => {
+  it('renders the transport details content once the price tab has been pressed', async () => {
```
mark-prins, line 48:
```suggestion
  it('renders the transport details content once the transport tab has been pressed', async () => {
```

packages/transactions/src/OutboundShipment/DetailView/tabs/GeneralTab.tsx
```diff
@@ -0,0 +1,74 @@
...
+    maxWidth: 100,
+    align: 'left',
+  },
+  {
+    label: 'label.packSize',
+    key: 'packSize',
+    width: 100,
+    minWidth: 100,
+    maxWidth: 100,
+    align: 'left',
```
mark-prins, line 39: :nit I'd right align the numeric columns and make them a little less wide than the name column

mark-prins (approved):
badass! nice to see it all come together.
Perhaps time now to rename the `RemoteDataTable` to just a `DataTable` :)

mark-prins:
oh, and the empty `stories` that have been appearing. Should they remain? are they placeholders or unused relics?

packages/common/src/hooks/useSortedData/useSortedData.ts
joshxg: Though the double render is probably irrelevant 🤔

packages/common/src/hooks/usePagination/usePagination.ts
joshxg: Umm yeah it's a bit weird. It's so you can destructure one or many properties and also keep a 'group'.. somewhat nice  for composing with other hooks..

PR #170, opened 2021-09-17
#169 UI tweaks

mark-prins:
Fixes #169 
a minor tidy - 'new shipment' button now full width, and the width transition still kinda works.
added the option to specify colour of the drop down menu items. maybe that's OTT

PR #186, opened 2021-09-20
#139 create 404 page

### LINKED ISSUE #139: Shell: create a page not found page
The unhappy man maybe? We don't have design for a 404 yet, thought it might be useful to knock one up.

mark-prins:
Fixes #139 
I'm not even certain we want this.. I think so. I'd created an issue, so thought to get it done. Also you are missing out on PR reviewing action.


mark-prins:
I wonder if we need a separate catch on the detail pages too though. If you enter `/customers/customer-invoice/anyoldrubbish` you get a blank.

packages/common/src/hooks/useDetailPanel.tsx
```diff
@@ -49,6 +49,7 @@ export const useDetailPanel = (): DetailPanel => {
         icon={<MenuDots />}
         labelKey="button.more"
         onClick={() => open()}
+        shouldShrink
```
mark-prins, line 52: leftover issue to address, from [this comment](https://github.com/openmsupply/openmsupply-client/pull/170#pullrequestreview-757936443)

packages/common/src/intl/intlHelpers.ts
```diff
@@ -18,6 +18,20 @@ export const useTranslation = (): ((
     id ? intl.formatMessage({ id: id as string }, values) : '';
 };
 
+export const useTranslationWithFallback = (): ((
```
mark-prins, line 21: possibly useful elsewhere? In this case, just so that the breadcrumb doesn't look daft

packages/common/src/intl/locales/ar.json
```diff
@@ -30,7 +30,9 @@
   "button.print": "Print",
   "button.select-a-color": "Select a color",
   "button.try-again": "Try again",
-  "error.something-wrong": "Oops! Somethings gone wrong.",
+  "error.404": "Click the back arrow or use the menu to find your way again.",
```
mark-prins, line 33: oh. sorry. I sorted the message files 🙄

packages/common/src/ui/components/menus/DropdownMenu/DropdownMenu.tsx
```diff
@@ -34,10 +34,9 @@ export const DropdownMenuItem: FC<DropdownMenuItemProps> = ({
       ) : null}
       <ListItemText
         inset={props.inset}
-        primaryTypographyProps={{ sx: { fontSize: 'inherit' } }}
+        primaryTypographyProps={{ color, sx: { fontSize: 'inherit' } }}
```
mark-prins, line 37: leftover issue from [this comment](https://github.com/openmsupply/openmsupply-client/pull/170#discussion_r711423474)

joshxg (approved):
Niiiiiice!


I had a quick look at formatjs and found that they had a default message and that if we declare in the code we can use their cli to extract it out .. although if you're inlining all your messages it makes it harder to reuse and have consistency 🤔 

https://formatjs.io/docs/getting-started/message-extraction/


I am also unsure about the use of unhappy man (not really) so I tried some alternatives

c2

Sep-20-2021 13-17-32

willdance

joshxg:
Oh but genuinely throwing a `Grow` or `Zoom` over the image? https://mui.com/components/transitions/

mark-prins:
yeah, I did try working with the default message originally and couldn't get it working nicely.
will pop that on the backburner - I've added a transition though, because that was an excellent suggestion :)

Issue #171, opened 2021-09-17
Bug: Pagination ellipsis clickable


mark-prins:
Small issue with the pagination control - click on a page ( say, 2 ) then click on the ellipsis.
The (2) loses focus, and if you then click next page (>) you are put on page 0

PR #187, opened 2021-09-20
Add controlled state to PaginationRow

### LINKED ISSUE #172: Outbound: Pagination unmounting
- When there is a load, the condition unmounts the table - this is generally ok (but not ideal) but not for the pagination as it is currently the page state is uncontrolled
- So, the internal state of the pagination resets to `0`
- I removed the `keepPreviousData`, so this is more obvious as the loading state is set everytime theres a new query.

So:

1. Explicitly pass the pagination component the current page number, probably. This is so this doesn't happen again.
2. Probably best to not unmount the entire table during loads.. maybe we can throw up an overlay over the table body or the subtle pagination loading indicator or something.

joshxg:
Fixes #172 #171

I think!

Makes the page state controlled explicitly rather than internally by the pagination component state


packages/common/src/ui/layout/tables/RemoteDataTable.tsx
```diff
@@ -164,9 +164,10 @@ export const RemoteDataTable = <T extends Record<string, unknown>>({
         </TableBody>
       </MuiTable>
       <PaginationRow
+        page={pagination.page}
         offset={pagination.offset}
         first={pagination.first}
-        total={pagination.total}
+        total={pagination.total ?? 0}
```
mark-prins, line 170: so close to being able to just spread 'em: `{...pagination}`

I think that others might request that all these props be destructured before use. I think fine like it is.

packages/common/src/ui/layout/tables/types.ts
```diff
@@ -44,19 +44,13 @@ export interface DataTableApi<T> {
...
-}
-
 export interface TableProps<T extends Record<string, unknown>> {
   columns: Column<T>[];
   data?: T[];
   sortBy: SortBy<T>;
   isLoading?: boolean;
   onSortBy: (sortRule: SortRule<T>) => void;
-  pagination: Pagination;
+  pagination: { page: number; total?: number; offset: number; first: number };
```
mark-prins, line 53: to avoid confusion with the `Pagination` interface exported from `usePagination`?

mark-prins (approved):
👍  I can't break the pagination now!!!

packages/common/src/ui/layout/tables/types.ts
joshxg: hmm also that `Pagination` has callbacks, changed em a little here 
66b3994 and 52660ed

PR #204, opened 2021-09-22
Add selected row state

### LINKED ISSUE #156: Outbound shipment: More row selection fun!
- If you select a row, then sort, the selection is removed. This is because the `data` changes (comes from the server) and is the intended behaviour.
- If you add `autoResetSelectedRows: false` - then the rows are kept selected, however they're selected by index. This means if you select row 2, then change the sort order, row 2 continues to be sorted.

So...

- Is this OK? I would think that we should keep the rows selected 🤔 
- I also had a thought around selecting on one page then going to the next page - obviously we lose the data of the next page, but wondered if the user expectation is that you can select all rows on a page, then go to the next, select all the rows, and perform an action on both pages

joshxg:
Fixes #156 

This is what I ended up with. Not 100% happy and open to suggestions or scraping. Confused myself. Why do I do this!

### DIFF
--- a/packages/common/src/ui/layout/tables/context/TableContext.ts
+++ b/packages/common/src/ui/layout/tables/context/TableContext.ts
@@ -1,0 +1,91 @@
+import create, { UseStore } from 'zustand';
+import createContext from 'zustand/context';
+
+export interface RowState {
+  isSelected: boolean;
+}
+
+export interface TableStore {
+  rowState: Record<string, RowState>;
+  numberSelected: number;
+
+  toggleSelected: (id: string) => void;
+  toggleAll: () => void;
+  setActiveRows: (id: string[]) => void;
+}
+
+export const { Provider: TableProvider, useStore: useTableStore } =
+  createContext<TableStore>();
+
+export const createTableStore = (): UseStore<TableStore> =>
+  create<TableStore>(set => ({
+    rowState: {},
+    numberSelected: 0,
+
+    toggleAll: () => {
+      set(state => {
+        const rowIds = Object.keys(state.rowState);
+        const numberOfRows = rowIds.length;
+        const isSelected = state.numberSelected !== numberOfRows;
+        const numberSelected = isSelected ? numberOfRows : 0;
+
+        return {
+          ...state,
+          numberSelected,
+          rowState: Object.keys(state.rowState).reduce(
+            (newState, id) => ({
+              ...newState,
+              [id]: { isSelected },
+            }),
+            state.rowState
+          ),
+        };
+      });
+    },
+
+    setActiveRows: (ids: string[]) => {
+      set(state => {
+        const { rowState } = state;
+
+        // Create a new row state, which is setting any newly active rows to unselected.
+        const newRowState: Record<string, RowState> = ids.reduce(
+          (newRowState, id) => {
+            return {
+              ...newRowState,
+              [id]: { isSelected: rowState[id]?.isSelected ?? false },
+            };
+          },
+          {}
+        );
+
+        const numberSelected = Object.values(newRowState).filter(
+          ({ isSelected }) => isSelected
+        ).length;
+
+        return { ...state, numberSelected, rowState: newRowState };
+      });
+    },
+
+    toggleSelected: (id: string) => {
+      set(state => {
+        const { numberSelected, rowState } = state;
+
+        // How many rows in total are currently rendered to determine
+        // if all rows, some or none are selected.
+        const isSelected = !rowState[id]?.isSelected;
+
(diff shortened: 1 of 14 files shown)


packages/transactions/src/OutboundShipment/DetailView/DetailView.tsx
```diff
@@ -187,3 +189,11 @@ export const OutboundShipmentDetailView: FC = () => {
...
   ) : null;
 };
+
+export const OutboundShipmentDetailView: FC = () => {
+  return (
+    <TableProvider createStore={createTableStore}>
+      <OutboundShipmentDetailViewComponent />
+    </TableProvider>
+  );
+};
```
joshxg, line 199: Now I've added the `useTableStore` into the `RemoteDataTable` the context becomes mandatory` 🤔

packages/transactions/src/OutboundShipment/ListView/ListView.tsx
```diff
@@ -23,16 +22,65 @@ import {
...
+  createTableStore,
+  getCheckboxSelectionColumn,
+  ColumnDefinition,
+  useTableStore,
 } from '@openmsupply-client/common';
 
 import { OutboundShipmentListViewApi } from '../../api';
 
-export const OutboundShipmentListView: FC = () => {
+const ListViewToolBar: FC<{
```
joshxg, line 34: Just shifting it out so avoid massive re-rendering of the listview when a select happens which then slices the data again aaand..

Looking to extract this out anyway, but a bit nicer soon

packages/common/src/ui/layout/tables/context/TableContext.ts
```diff
@@ -0,0 +1,91 @@
...
+        // If this row is being toggled on, add one, otherwise reduce the number
+        // of rows selected.
+        const newNumberSelected = numberSelected + (isSelected ? 1 : -1);
+
+        return {
+          ...state,
+          numberSelected: newNumberSelected,
+          rowState: {
+            ...state.rowState,
+            [id]: { ...state.rowState[id], isSelected },
```
mark-prins, line 86: I'm not sure the point of spreading `state.rowState[id]` as `isSelected` is the only prop

packages/common/src/ui/layout/tables/types.ts
```diff
@@ -13,7 +13,7 @@ export enum ColumnFormat {
 }
 
 export interface ColumnDefinition<T> {
-  label: LocaleKey;
+  label?: LocaleKey;
```
mark-prins, line 16: hm? what did I miss?

packages/transactions/src/OutboundShipment/DetailView/DetailView.tsx
```diff
@@ -80,7 +82,7 @@ const useDraftOutbound = (id: string) => {
   return { draft, setDraft, save };
 };
 
-export const OutboundShipmentDetailView: FC = () => {
+export const OutboundShipmentDetailViewComponent: FC = () => {
```
mark-prins, line 85: doesn't need to exported

packages/transactions/src/OutboundShipment/DetailView/DetailView.tsx
mark-prins: have you tried with two tables on a page? that would work, right? having a different provider wrapped around each.

I'm concerned that any component which uses a data table has to remember to wrap it in a provider, though I think I'm worrying about nothing and it is probably fine

mark-prins (approved):
👍  I think its great - took me a little time to understand what the test descriptions were all about, in the end it's all clear enough.

packages/common/src/ui/layout/tables/types.ts
joshxg: The checkbox column doesn't need the label

packages/common/src/ui/layout/tables/context/TableContext.ts
joshxg: Just so if we add, say, a `isDisabled` this still works, I guess!

joshxg:
two pages: Yeah works, but need to wrap each one in it's own context.

Context concern in general: I have the same concern. Technically the wrapper is not needed now since extracting the `useTableStore` to the `ListViewToolbar` component. But the table in general does need to be wrapped in the context. I think I convinced myself by material-ui moving toward a `TabContext` if you want to use tabs and I likened this to having a `toolbar` (the dropdown which needs to know which rows are selected) and the table.

Another factor that i'm a little unsure about is why the row select state is in context, but the sort by state isn't? Or the pagination state? 🤔

PR #205, opened 2021-09-23
#spotlight

joshxg:
Bit of a prototype to see how you feel. Found plenty of problems and things to do better which is good.

The maintainer is trying to make this component basically as flexible as possible and putting it a good amount of work. The code base is nice and small enough to be very approachable.

image

Package issues:
- Bug when there is no results 🤔  https://kbar.vercel.app/ this is happening here, but just reports an error in the console. In dev, the oh noes! screen pops up.
- Types need some help but aren't too bad. For example the package that's being used for keyword matching can accept a string array for keywords, but the type is only a string.
- I have no idea how the shortcuts work 🤔 
- The actions are memoized, so if you're using an action where the perform is a closure, you'll get funny results. I fixed this by making the function a `useRef`, but it's not ideal. Ideally would be able to pass a `deps` array in to the `  useRegisterActions` hook.

Issues with what i've done:
- I think ideally the host wouldn't know about navigating to other actions. For example, I wanted the `CustomerContainer` to register all the `customer` based navigation actions. However, that's not loaded until you navigate to it. As I right this now, I realise that I could have put this in the `CustomerNavigation` component.. 🤔 
- Actions are registered all over the place. Pros and cons, I think. Pro: It's super easy to register some stuff where the actual code is that does the action. Cons: They're scattered all over the place!
- The UI of it is horrid. Would be nice to have icons, like a `+` next to Create actions and some sort of navigation action next to navigation actions. Also subtitles. These are both got issues for em. Also better styling.

In general I *love* ❤️  it and think it is awesome. But in saying that, I love macos spotlight and even bought Alfred to enhance it! So I am incredibly bias. Would also need a bit more structure to make sure it doesn't go crazy, I suspect

### DIFF
--- a/packages/host/src/Host.tsx
+++ b/packages/host/src/Host.tsx
@@ -51,6 +51,7 @@ const Heading: FC<{ locale: string }> = props => (
 );
 
 const CustomKBarSearch = styled(KBarSearch)(({ theme }) => ({
+  color: '#fff',
   width: 500,
   height: 50,
   fontSize: 20,
@@ -58,6 +59,10 @@ const CustomKBarSearch = styled(KBarSearch)(({ theme }) => ({
   borderRadius: '5px',
   ':focus-visible': {
     outline: 'none',
+  },
+  '::placeholder': {
+    color: '#fff',
+    opacity: 0.75,
   },
 }));
 
--- a/packages/host/src/Host.tsx
+++ b/packages/host/src/Host.tsx
@@ -12,9 +12,17 @@ import {
   RouteBuilder,
   ErrorBoundary,
   GenericErrorFallback,
+  KBarProvider,
+  KBarPortal,
+  KBarPositioner,
+  KBarAnimator,
+  KBarSearch,
+  KBarResults,
+  useDrawer,
+  styled,
 } from '@openmsupply-client/common';
 import { AppRoute } from '@openmsupply-client/config';
-import { BrowserRouter, Routes, Route } from 'react-router-dom';
+import { BrowserRouter, Routes, Route, useNavigate } from 'react-router-dom';
 import AppDrawer from './AppDrawer';
 import AppBar from './AppBar';
 import DetailPanel from './DetailPanel';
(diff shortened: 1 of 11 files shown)


packages/host/src/Host.tsx
```diff
@@ -42,6 +50,122 @@ const Heading: FC<{ locale: string }> = props => (
...
 
+const CustomKBarSearch = styled(KBarSearch)(({ theme }) => ({
+  width: 500,
+  height: 50,
+  fontSize: 20,
+  backgroundColor: theme.palette.primary.main,
+  borderRadius: '5px',
+  ':focus-visible': {
+    outline: 'none',
+  },
```
mark-prins, line 61: I found the grey text hard to read - what about this?
```suggestion
  color: '#fff',
  width: 500,
  height: 50,
  fontSize: 20,
  backgroundColor: theme.palette.primary.main,
  borderRadius: '5px',
  ':focus-visible': {
    outline: 'none',
  },
  '::placeholder': {
    color: '#fff',
    opacity: 0.75,
  },
```

packages/host/src/Host.tsx
```diff
@@ -51,74 +175,80 @@ const Host: FC = () => {
...
-                            <Route
-                              path={RouteBuilder.create(AppRoute.Messages)
-                                .addWildCard()
-                                .build()}
-                              element={
-                                <Heading locale={locale}>messages</Heading>
-                              }
-                            />
-                            <Route path="/" element={<DashboardService />} />
+              <CommandK>
```
mark-prins, line 178: more nesting than a bird sanctuary

mark-prins (approved):
🎉  Squeeeee!!! that is awesome

can see that being super helpful. I had a play around with the shortcuts, as the ones defined need a little refinement. I was playing with combinations too and that's also working well.

packages/host/src/Host.tsx
joshxg: 🤣   Heh, yeah.

I ended up going with a `DependencyProvider` in cold chain to handle it all..

mark-prins:
`go to ci 7` 🤯  we're not far away from a nice looking command line web app

PR #212, opened 2021-09-25
#194 shifting column definitions

### LINKED ISSUE #194: Outbound: Shift columns from ListView
- ListView has a bunch of columns in the middle of it!
- We need to decide where to put column definitions.
- Thinking- column definitions into `layout/table/columns` and then each component with a table has a `columns` file.
- Doesn't help with the 'customisation' problem - a'la mobile reasons, prices and vaccines oh my.
- Maybe need to use a builder pattern to construct a columns array which can specify a column
- Also need to be able to pass parameters to the column - for example the color column needed to have the `onUpdate` passed in

joshxg:
Fixes #194 

- What I wanted was a simple builder pattern and what I ended up with was a big PR
- I started working on the builder and I struggled a lot with the react table types. I was spending a lot of time fiddling and making sure types were matching but ended up having to do a lot of `as X`. 
- Decided that I had lost way to much time on react-table types now and in the past and decided to go with custom ones, though they are heavily inspired because I do like the general approach.

Ended up with something very similar:
- `ColumnDefinition<T>` is a 'subset' of a `Column<T>` that only needs a `key`. The intention is that this definition is passed into the `useColumns` function which can just fill the rest of the defaults and spits out a `Column` which is used in the table. This has all the fields we use without having to use all the optional plugin fields etc
- A `Column` has a `Cell` and `Header` fields which have to be react components which take `CellProps` and `HeaderProps`, which if not set, are assigned a `BasicCell` and `BasicHeader` and so we can just render those.

To build the columns:
- `ColumnSetBuilder` can build using `addColumn` which can take a `columnKey` (which maps to some default columns)` and some optional overrides
- Or, can take a full `ColumnDefinition`
- This is then built into an array which is ordered by the `order` field on a column and passed into `useColumns`
- Possibly should just pass an array of `columnKey` or `ColumnDefintion` into `useColumns` and `useColumns` uses the `ColumnSetBuilder`

Some things I was thinking about while doing this:
- Trying to have a solution which can allow for some sort of storing of a string in `localStorage` for the columns a user wants to see (assuming some sort of `hide columns` feature).
- Also potentially the order of columns. So, serialising something like `[{key: "status", order: 3}`, {key:"type", order: 2]` could work
- Also a little bit about plugins. Thinking something around the lines of being able to dynamically load a JS bundle with a `column` object, which has the `Cell`/`Header` and other things and be able to simply pass this object into the array and 'for it all to just work'. Though, I am not really sure and didn't spend too much time on this as there's a few options (maybe just a plugin registry and instead the columnkey can just be `plugin:2` and 'it just works!'


packages/common/src/types/index.ts
```diff
@@ -1,14 +1,21 @@
+import { ObjectWithStringKeys } from './utility';
+
 export * from './utility';
 
-export type Item = {
+type RecordWithId = { id: string };
+
+export interface DomainObject extends RecordWithId, ObjectWithStringKeys {}
```
joshxg, line 7: This is quite silly but having the assurance that there was a key `id` helped in a few places.

The name needs help and maybe just `RecordWithId` would be better

packages/common/src/ui/layout/tables/RemoteDataTable.tsx
```diff
@@ -76,39 +70,33 @@ export const RemoteDataTable = <T extends Record<string, unknown>>({
...
+                isSortable={!!column.sortable}
+                isSorted={column.key === sortBy.key}
+                align={column.align}
+                columnKey={column.key}
+                direction={sortBy.direction}
+              >
+                <column.Header column={column} />
+              </HeaderCell>
+            ))}
+          </HeaderRow>
```
joshxg, line 89: I took away the headergroup and now just rendering `column.Header` directly.

Looking at this now, maybe would be better to just pass the column object straight to the `Cell`, that's likely easier

packages/common/src/ui/layout/tables/RemoteDataTable.tsx
```diff
@@ -76,39 +70,33 @@ export const RemoteDataTable = <T extends Record<string, unknown>>({
...
             return (
               <DataRow<T>
-                cells={cells}
-                key={key}
+                columns={columns}
+                key={row.id}
                 onClick={onRowClick}
-                rowData={original}
+                rowData={row}
+                rowKey={String(idx)}
```
joshxg, line 99: passing the `columns` over `Cells` means we can memoise the data row now if we wanted and also just made typing things a lot easier for me..

packages/common/src/ui/layout/tables/columns/NameAndColorColumn.tsx
```diff
@@ -1,20 +1,22 @@
...
 import { Box } from '@mui/system';
+import { DomainObject } from '../../../..';
+import { ColumnDefinition } from '../columns/types';
 
-type Colorable = {
+interface DomainObjectWithRequiredFields extends DomainObject {
   color: string;
-} & Record<string, unknown>;
+  name: string;
+}
```
joshxg, line 11: this is cute, it means I can't use that column for the column set in the `GeneralTab`

packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
```diff
@@ -50,12 +50,17 @@ export const DataRow = <T extends Record<string, unknown>>({
...
+              width: column.width,
             }}
           >
-            {cell.render('Cell')}
+            <column.Cell
+              rowData={rowData}
+              columns={columns}
+              column={column}
+              rowKey={rowKey}
+            />
```
joshxg, line 63: just iterating over the columns rather than cells which is basically same same

packages/common/src/ui/layout/tables/utils/ColumnSetBuilder.ts
```diff
@@ -0,0 +1,161 @@
...
+  | 'confirmed'
+  | 'total'
+  | 'comment'
+  | 'selection'
+  | 'name'
+  | 'code'
+  | 'packSize'
+  | 'quantity';
+
+const getColumnLookup = <T extends DomainObject>(): Record<
```
joshxg, line 27: I tried to get the typings of this nicer. I'm sure it must be possible but i'll need to revisit unless you have some ideas.

I wanted to constrain the `columnKey` passed to `addColumn` to be only those in the set of `keyof T` also. I couldn't get this to work. If I had, then I would have also possibly had to figure out a solution for columns which are not of `keyof T`, but are legal, such as `selection` 🤔

packages/common/src/ui/layout/tables/utils/ColumnSetBuilder.ts
```diff
@@ -0,0 +1,161 @@
...
+  },
+  quantity: {
+    label: 'label.quantity',
+    key: 'quantity',
+    width: 20,
+    align: ColumnAlign.Right,
+  },
+});
+
+export class ColumnSetBuilder<T extends DomainObject> {
```
joshxg, line 95: In general I think this is OK, not great. 

I'm still not fully sure how we'd handle customised columns yet but I think this gets us close..

packages/transactions/src/OutboundShipment/ListView/ListView.tsx
```diff
@@ -101,81 +98,23 @@ export const OutboundShipmentListViewComponent: FC = () => {
...
+      .addColumn('type')
+      .addColumn('status')
+      .addColumn('invoiceNumber')
+      .addColumn('confirmed')
+      .addColumn('entered')
+      .addColumn('total')
+      .addColumn('comment')
+      .addColumn('selection')
+      .build()
+  );
```
joshxg, line 117: Again, passing `['type', 'status', 'invoiceNumber', 'confirmed', 'entered', 'total', 'comment', 'selection']` into `useColumns` (And, maybe read this array from `localStorage`, if set) then having `useColumns` handle the builder would be better.

packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
```diff
@@ -33,14 +36,11 @@ export const DataRow = <T extends Record<string, unknown>>({
       onClick={onRowClick}
       hover={hasOnClick}
     >
-      {cells.map(cell => {
-        const cellProps = cell.getCellProps();
-        const { key: cellKey } = cellProps;
-
+      {columns.map(column => {
```
mark-prins, line 39: no need for the return now

packages/common/src/types/utility.ts
```diff
@@ -1,6 +1,6 @@
 // This type ensures that the resulting value of a `keyof` is
 // a string.
 // eslint-disable-next-line @typescript-eslint/ban-types
-export type KeyOf<T extends object> = Extract<keyof T, string>;
+export type KeyOf<T extends object> = Extract<keyof T, string & keyof T>;
```
mark-prins, line 4: I couldn't see any usages of `KeyOf` now. could it be removed?

packages/common/src/ui/layout/tables/utils/ColumnSetBuilder.test.ts
```diff
@@ -0,0 +1,86 @@
...
+      .addColumn({ key: 'invoiceNumber' })
+      .build();
+
+    expect(columns).toEqual([
+      expect.objectContaining({ key: 'status' }),
+      expect.objectContaining({ key: 'invoiceNumber' }),
+    ]);
+  });
+
+  it('creates an array of column objects when passed a min of column definitions and keys', () => {
```
mark-prins, line 29: :nit 
```suggestion
  it('creates an array of column objects when passed a mix of column definitions and keys', () => {
```
?

packages/common/src/ui/layout/tables/utils/ColumnSetBuilder.ts
mark-prins: yeah - no great ideas sorry. I think we should merge this and revisit

packages/transactions/src/OutboundShipment/ListView/ListView.tsx
mark-prins: for sure - that would be great to do; both the passing of an array ( though the `addColumn` syntax is nice and clear anyway, probably more readable in fact ) and the support for column user preferences

mark-prins (approved):
👍  that is fantastic - nice wrangling of typescript and generics and a clear and reusable approach

mark-prins:
maybe change the target branch though? we might not need to remove kbar now, which would be good

packages/common/src/types/utility.ts
joshxg: will do!

joshxg:
@mark-prins spent a bit of time cleaning some commits to try and change the base from the remove kbar branch.

going to be a lot easier to just add it back afterward, sorry.

packages/common/src/ui/layout/tables/RemoteDataTable.tsx
joshxg: Will do this also

PR #213, opened 2021-09-26
#209 demo server issues

### LINKED ISSUE #209: Demo server issues
Some minor issues I've noticed in reviewing the current build
- the pagination is only showing page one ( due to line 96 in `packages/mock-server/src/schema/Transaction.js )
- The buttons in storybook are showing very wide ( had to wrap in a `Box` to constrain them )
- Same issue in the colour menu
- The DetailView and ListView have an error ( as part of this, will need to do something about kbar )
- Storybook won't build ( temporarily resolved by changing the ts target in a custom build of kbar )

joshxg:
Fixes #209 

- New release of kbar adding es5 compiling added. I've not added all the code back as I 1) want to check it is working for you 2) maybe want to work out a better strategy to adding actions that is better than the mvp of throwing the actions around so much
- Import paths made the pages break - annoying
- Boxes around the buttons
- Not sure what the pagination issue was it seems to be OK

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -19,6 +19,7 @@
     "graphql": "^15.5.1",
     "graphql-request": "^3.4.0",
     "history": "^5.0.0",
+    "kbar": "^0.1.0-beta.5",
     "notistack": "^1.0.6-next.3",
     "react-query": "^3.24.3",
     "react-router-dom": "^6.0.0-beta.0",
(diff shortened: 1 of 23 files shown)


mark-prins:
@joshxg the pagination problem is line 96 in `Transaction.js` I have a commit, I'll add it in to your branch

packages/common/package.json
```diff
@@ -19,6 +19,7 @@
     "graphql": "^15.5.1",
     "graphql-request": "^3.4.0",
     "history": "^5.0.0",
+    "kbar": "^0.1.0-beta.5",
```
mark-prins, line 22: that was good timing!

mark-prins (approved):
👍  works nicely!
those pesky imports.. I had forgotten about that little problem

packages/common/package.json
joshxg: I made an issue and it was sorted within a few hours 🤓  😁

Issue #192, opened 2021-09-20
Outbound: Editable cells


joshxg:
- Need to introduce an editable cell into the outbound detail page for the quantity!

mark-prins:
Along with the validation - to prevent issuing more stock than there is on hand for a particular batch, and how to handle display of warnings and validation errors?
I like the inline editing, would be nice to have editable cells if we can

richardmoizeau:
Here are all designs for "Add Item" function. We have chosen that there will be no in-line editing even for quantity. However we haven't decided on a way to edit an already created shipment line. You double-click the line and the Add Item window opens ?

https://app.zeplin.io/project/60e44fff6fa942168092f1ff/dashboard?tag=Add%20Item

PR #214, opened 2021-09-28
#192 editable cells

joshxg:
Starts #192!

- After trying many different iterations of 'get the callback to a cell' 
    - Closure around the column i.e. like the name/color column is now (For this, needed to still pass the prop from page -> tab, or pass the columns from page > tab > table) - I am really not sure that this approach is better. It might just be a 'cool idea' and I should revert back to passing the callback when creating the column]
    - Straight passing a callback from page -> tab -> table -> row -> cell
    - Context, having the cell do something like `const {callback} = useContext(rowKey, columnKey)`

I landed on this approach which is essentially, add onto the data, a way of updating that data also..

So, wrapping an item `{id, quantity}` with a function to update the quantity `{id, quantity, setQuantity}`, and then the cell can just use `rowData.setQuantity()`

I thought it has ended up quite a nice and easy. Though, I can see a negative in mixing data/actions

Some things other things i've been thinking:
- Probably better to have an `ItemRow` type etc - technically not quite right with the api return value [We should also have different types for `TransactionListRow` and `Transaction`, probably, to avoid the nullables]
- Maybe using mobx here would just be better. We could create classes for `Item` etc, and wrap cells in `observe`. Then each cell would just automatically update itself 🤔 
- might be better to just use an `update(column,row,value)` function instead of `setQuantity(row, value)`, then every 'editable' type cell can just extend some `UpdatableRow` interface


packages/common/src/hooks/useEffectAfterMounting/useEffectAfterMounting.test.ts
```diff
@@ -0,0 +1,48 @@
+import { useEffectAfterMounting } from './useEffectAfterMounting';
```
joshxg, line 1: There's two hooks here `useEffectAfterMounting` and `useMountedRef`.

They were from previous iterations (Was triggering the callback when `value` changed, but not on the first render) until I just added the callback to `onBlur` which was a bit easier.

I left these in here as they work and might get use 🤷

packages/common/src/hooks/useSortBy/useSortBy.ts
```diff
@@ -44,8 +46,5 @@ export const useSortBy = <T extends ObjectWithStringKeys>({
     return { ...newSortBy, direction: getDirection(!!newSortBy?.isDesc) };
   }, []);
 
-  return {
-    sortBy: { ...sortBy, direction: getDirection(!!sortBy.isDesc) },
-    onChangeSortBy,
-  };
+  return { sortBy, onChangeSortBy };
```
joshxg, line 49: I need to revist sorting. Decided to leave it here for this PR.

This was returning a new sortBy object everytime, causing annoying re-renders

packages/common/src/hooks/useSortedData/useSortedData.ts
```diff
@@ -41,16 +41,14 @@ export const useSortedData = <T extends Record<string, unknown>>(
...
     setSortedData(data.sort(sorter));
-  }, []);
+  }, [sortBy]);
+
+  useEffect(() => {
+    setSortedData(data);
+  }, [data]);
 
-  return { sortedData, sortBy, onChangeSortBy: wrapped };
+  return { sortedData, sortBy, onChangeSortBy };
```
joshxg, line 53: ....

Hm.

So when the `data` changed from a parent and flowed down, the `sortedData` was still the same in state unless a sort occurred. So changed values weren't showing in the UI.

So again, need to revisit sorting 🙄  but this is working ok for this PR..

packages/common/src/ui/layout/tables/columns/EditableQuantityColumn.tsx
```diff
@@ -0,0 +1,73 @@
...
+import { DomainObject } from '../../../../types';
+import { TextField, InputAdornment, Tooltip, Typography } from '@mui/material';
+import React, { useState, useEffect } from 'react';
+
+interface SomeQuantityEntity extends DomainObject {
+  quantity: number;
+  setQuantity: (rowKey: number, newQuantity: number) => void;
+}
+
+export const getEditableQuantityColumn = <
```
joshxg, line 11: didn't touch the styles, just wanted to get the pattern for editable cells in place

packages/common/src/ui/layout/tables/columns/EditableQuantityColumn.tsx
```diff
@@ -0,0 +1,73 @@
+import { ColumnDefinition } from './types';
+import { DomainObject } from '../../../../types';
+import { TextField, InputAdornment, Tooltip, Typography } from '@mui/material';
+import React, { useState, useEffect } from 'react';
+
+interface SomeQuantityEntity extends DomainObject {
+  quantity: number;
+  setQuantity: (rowKey: number, newQuantity: number) => void;
+}
```
joshxg, line 9: this is pretty cool and makes sure the row data has the required fields if you want to use this column

packages/common/src/ui/layout/tables/columns/EditableQuantityColumn.tsx
```diff
@@ -0,0 +1,73 @@
...
+export const getEditableQuantityColumn = <
+  T extends SomeQuantityEntity
+>(): ColumnDefinition<T> => ({
+  key: 'quantity',
+  width: 100,
+  Cell: ({ rowData, rowKey }) => {
+    const { quantity } = rowData;
+    const [buffer, setBuffer] = useState(quantity);
+    const [value, setValue] = useState(quantity);
+    const [error, setError] = useState(false);
```
joshxg, line 20: all this really depends on what we're doing with the cells and the desired behaviour of what values should / shouldn't be allowed to enter and how restrictive we want to be with actual input.

Ideally I would want to try and avoid excessive constraints (issues with mobile not being able to enter floats because you can enter a `.` for example) and provide feedback instead

packages/common/src/ui/layout/tables/columns/EditableQuantityColumn.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+      if (isValid) {
+        setValue(asNumber);
+        setError(false);
+      } else {
+        setError(true);
+      }
+
+      setBuffer(asNumber);
+    };
```
joshxg, line 40: Just a basic implementation - has a buffer for invalid values and tries to update a valid, value.. value.

packages/common/src/ui/layout/tables/columns/EditableQuantityColumn.tsx
```diff
@@ -0,0 +1,73 @@
...
+    return (
+      <TextField
+        sx={{ maxHeight: 40 }}
+        error={error}
+        variant="filled"
+        size="small"
+        helperText="Incorrect value"
+        hiddenLabel
+        value={buffer}
+        onBlur={() => rowData.setQuantity(Number(rowKey), value)}
```
joshxg, line 57: on leaving the cell, then update the actual draft. A bit nicer for renders, though you start to run into race conditions with pressing buttons which might be disabled etc with this approach

packages/common/src/ui/layout/tables/hooks/useColumns.tsx
```diff
@@ -24,23 +25,27 @@ export const useColumns = <T extends DomainObject>(
...
+          align: ColumnAlign.Left,
+          ...getColumnWidths(column),
+        };
 
-    return { ...defaults, ...column };
-  });
+        return { ...defaults, ...column };
+      }),
+    []
+  );
```
joshxg, line 48: memoising otherwise the cells all remount when rendering

packages/transactions/src/OutboundShipment/DetailView/DetailView.tsx
```diff
@@ -48,7 +54,44 @@ const placeholderTransaction: Transaction = {
...
+
+        return newState;
+      });
+    },
+  }));
+
+  const newDraft = { ...original, items: mappedLines };
+
+  return newDraft;
+};
```
joshxg, line 92: So, added to `DraftStore` zustand store, the method `createDraft`. This is required when you create the store (so, need to create a method like this for any type of store, transaction, location etc) and is called after we receive a response from a request for data. This function is passed the zustand state setter, and injects each row with the `setQuantity` function.

In general I really like this and think it's nice and simple way to get the updater to a cell and even becomes type safe.

I wonder though if there are better ways to be doing this (MobX, valtio?, plain classes?, MobX State Tree?). I've not seen this approach before and I suspect it's because it seems very similar to a mobx style approach (Where, we wrap the response in some mobx class, and pass that down the tree) but instead of wrapping each cell in an `observe` HOC, we still trigger the normal render

packages/transactions/src/OutboundShipment/ListView/ListView.test.tsx
```diff
@@ -8,6 +8,8 @@ import AppBar from '@openmsupply-client/host/src/AppBar';
 
 import { OutboundShipmentListView } from './ListView';
 
+jest.setTimeout(10000);
```
joshxg, line 11: not great but was sometimes getting a timeout

packages/transactions/src/OutboundShipment/DetailView/DetailView.tsx
joshxg: could also make this simpler with https://immerjs.github.io/immer/produce maybe.

It's been awhile since I'd done this immutable style updating. I really hate it 🤣

packages/mocks/src/handlers.ts
```diff
@@ -11,7 +11,7 @@ const choose = (options: unknown[]) => {
 };
 
 const getItems = () =>
-  Array.from({ length: Math.random() * 10 }).map(() => ({
+  Array.from({ length: Math.random() * 10 + 1 }).map(() => ({
```
mark-prins, line 14: that was annoying, huh? wanted to have the option when demoing, to show what no items would look like. not needed now I think

mark-prins (approved):
🎉 

well that all works nicely. I also like the passing of the update function and adding it to the row object - there are so many different ways to do the same thing! Pros and cons with each. 

Would be good to rename the `rowData` now that it has an action though

It feels fairly specific at this stage, with everything tied to the 'quantity' field when this could all apply to any editable, positive, non-zero numeral. That's minor stuff to be tidied later though.

I am curious about mobx and possibly other options, which could give much the same; so possibly it's worth looking into further. this approach seems fine though and I'm not feeling much need to look around at alternatives.

packages/common/src/ui/layout/tables/columns/EditableQuantityColumn.tsx
joshxg: Yup. I also wasn't exactly sure what to do with the styles as there's no examples other than the add item dialog, but the styling on that is also a slightly different

packages/common/src/ui/layout/tables/columns/EditableQuantityColumn.tsx
```diff
@@ -0,0 +1,73 @@
...
+      const isValid = Number.isInteger(asNumber) && asNumber >= 0;
+
+      if (isValid) {
+        setValue(asNumber);
+        setError(false);
+      } else {
+        setError(true);
+      }
+
+      setBuffer(asNumber);
```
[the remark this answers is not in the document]
joshxg: made it a string here f591523

PR #215, opened 2021-09-28
Add status bar and footer

### LINKED ISSUE #150: Framework: Status
Unsure which was style was chosen:

image

image
image

mark-prins:
Fixes #150 

Add a footer which shows status of current user

### DIFF
--- a/packages/common/src/hooks/useHostContext.ts
+++ b/packages/common/src/hooks/useHostContext.ts
@@ -3,6 +3,7 @@ import create from 'zustand';
 import { LocalStorage } from '../localStorage';
 
 import { SupportedLocales } from '../intl/intlHelpers';
+import { Store, User } from '../types';
 
 type HostContext = {
   setAppBarContentRef: (ref: React.MutableRefObject<null> | null) => void;
@@ -10,6 +11,10 @@ type HostContext = {
   appBarContentRef: React.MutableRefObject<null> | null;
   locale: SupportedLocales;
   setLocale: (locale: SupportedLocales) => void;
+  setStore: (store: Store) => void;
+  store: Store;
+  setUser: (user: User) => void;
+  user: User;
 };
 
 export const useHostContext = create<HostContext>(set => ({
@@ -18,7 +23,11 @@ export const useHostContext = create<HostContext>(set => ({
   locale: LocalStorage.getItem('/localisation/locale') ?? 'en',
   appBarButtonsRef: createRef(),
   appBarContentRef: null,
+  store: { id: '4321dcba', name: 'Central Warehouse' },
+  user: { id: 'abcd1234', name: 'Administrator' },
   setLocale: locale => set(state => ({ ...state, locale })),
+  setUser: user => set(state => ({ ...state, user })),
+  setStore: store => set(state => ({ ...state, store })),
 }));
 
 useHostContext.subscribe(({ locale }) => {
--- a/packages/common/src/hooks/useRowRenderCount/useRowRenderCount.ts
+++ b/packages/common/src/hooks/useRowRenderCount/useRowRenderCount.ts
@@ -11,9 +11,14 @@ export const useRowRenderCount = (): number => {
   const dataRowHeight = mixins.table.dataRow.height;
   const headerRowHeight = mixins.table.headerRow.height;
   const paginationRowHeight = mixins.table.paginationRow.height;
-
+  const pageFooterHeight = mixins.footer.height;
+  console.info('windowHeight', windowHeight);
   const numberOfRowsToRender = Math.floor(
-    (windowHeight - (height ?? 0) - headerRowHeight - paginationRowHeight) /
+    (windowHeight -
+      (height ?? 0) -
+      headerRowHeight -
+      paginationRowHeight -
+      pageFooterHeight) /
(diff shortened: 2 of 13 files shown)


packages/common/src/hooks/useHostContext.ts
```diff
@@ -3,13 +3,18 @@ import create from 'zustand';
...
 import { SupportedLocales } from '../intl/intlHelpers';
+import { Store, User } from '../types';
 
 type HostContext = {
   setAppBarContentRef: (ref: React.MutableRefObject<null> | null) => void;
   appBarButtonsRef: React.MutableRefObject<null> | null;
   appBarContentRef: React.MutableRefObject<null> | null;
   locale: SupportedLocales;
   setLocale: (locale: SupportedLocales) => void;
+  setStore: (store: Store) => void;
```
mark-prins, line 14: probably not the best place for it - just wanted to source the values from somewhere stateful

packages/common/src/hooks/useHostContext.ts
```diff
@@ -18,7 +23,11 @@ export const useHostContext = create<HostContext>(set => ({
   locale: LocalStorage.getItem('/localisation/locale') ?? 'en',
   appBarButtonsRef: createRef(),
   appBarContentRef: null,
+  store: { id: '4321dcba', name: 'Central Warehouse' },
```
mark-prins, line 26: and this will need to change when we get to auth and sessions

packages/common/src/hooks/useRowRenderCount/useRowRenderCount.ts
```diff
@@ -11,9 +11,14 @@ export const useRowRenderCount = (): number => {
   const dataRowHeight = mixins.table.dataRow.height;
   const headerRowHeight = mixins.table.headerRow.height;
   const paginationRowHeight = mixins.table.paginationRow.height;
-
+  const pageFooterHeight = mixins.footer.height;
```
mark-prins, line 14: it does reduce the rows to display unfortunately. 
the problem is just that the table squashes the footer so that the items don't vertically align

packages/common/src/ui/icons/User.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import SvgIcon, { SvgIconProps } from '@mui/material/SvgIcon';
+
+export const UserIcon = (props: SvgIconProps): JSX.Element => {
```
mark-prins, line 4: and here is where I regret calling icons 'Something' and not 'SomethingIcon' 🙄

packages/host/src/Footer.stories.tsx
```diff
@@ -0,0 +1,33 @@
+import React, { useEffect } from 'react';
```
mark-prins, line 1: Daddy! can I have a story? Make it a really boring one!

packages/host/src/Footer.tsx
```diff
@@ -0,0 +1,41 @@
...
+  styled,
+  Typography,
+  useHostContext,
+  UserIcon,
+} from '@openmsupply-client/common';
+
+export const Footer: React.FC = () => {
+  const { user, store } = useHostContext();
+  const PaddedCell = styled(Box)({ display: 'flex', padding: '0 8px' });
+  const iconStyles = { color: 'midGrey', height: '16px', width: '16px' };
```
mark-prins, line 14: ok, so i'm really mixing the style of styles with this. I wanted to use the access to `theme` that `sx` provides and also didn't want to have things duplicated.

packages/common/src/ui/icons/User.tsx
joshxg: haha yes. I think it's not too late and probably a good idea

packages/host/src/Footer.tsx
joshxg: I think it's fine 🤷   it's unlikely one of these change and not the other?

packages/host/src/Footer.tsx
```diff
@@ -0,0 +1,41 @@
...
+  Home,
+  styled,
+  Typography,
+  useHostContext,
+  UserIcon,
+} from '@openmsupply-client/common';
+
+export const Footer: React.FC = () => {
+  const { user, store } = useHostContext();
+  const PaddedCell = styled(Box)({ display: 'flex', padding: '0 8px' });
```
joshxg, line 13: Can this go outside of the component? It's a bit weird here and is also going to remount rather than re-render  each time, unnecessarily

packages/host/src/Footer.stories.tsx
joshxg: I am hesitant to say as I am unsure of the value, but also a test to just make sure that there is text equal to the currently logged in user and store I *think* might one day help someone

joshxg (approved):
Awesome. I think it makes the pages look more complete. Getting there 🤣 

- The dashboard service has a `100vh` on the `main` tag so the footer isn't showing there
- Curious if clicking these in the future is how we switch stores/user? I was somewhat expecting the username/store to be around some sort of 'profile' thing in the top right which has a logout but I think that's just because I see browser

mark-prins:
thanks - added tests and fixed dashboard

PR #216, opened 2021-10-01
Add items to api

joshxg:
I was trying to allow  the transactions to have the same items and also add an update to items


packages/mock-server/src/schema/Transaction.js
```diff
@@ -101,7 +69,7 @@ const getItems = () =>
   Array.from({ length: Math.random() * 10 }).map(() => ({
     id: `${faker.datatype.uuid()}`,
     code: `${faker.random.alpha({ count: 6 })}`,
-    name: `${faker.commerce.productName()}`,
+    name: `${faker.commerce.productName()}`,1
```
mark-prins, line 72:
```suggestion
    name: `${faker.commerce.productName()}`,
```

mark-prins (approved):
🎉 
Sweet! thanks - that's really helpful. There's a mis-type which affected compilation - otherwise, all good!

PR #220, opened 2021-10-04
#218 order app bar buttons

### LINKED ISSUE #218: Framework: Order of AppBar buttons
I think we need to have a way to order the buttons in the AppBar, beyond what we can do currently.

The `Add item` button needs to be on the left, as that is an important / often used button. The `More` has to be on the right, so that it is linked with the detail panel. Currently both of those are added by the page and therefore can't be located at both ends, the buttons which are shown on every page (Docs & Language) are shown on the left.

Wondering about just adding an order property - or an alignment prop, similar to flex: end, center & start; or left, middle & right. 

The comment has been made that Docs isn't required in the top bar, with the argument being that it is only helpful for new users. I disagree, thinking that a 'help' option has to always be readily available, so would argue to keep it on every page.

Might also be time to move `Language` out - that can possibly be added to the settings section for now?

joshxg:
Fixes #218 

- I went with the really bad solution.
- I started by copying the content portal component where I was going to use `React.Childre.toArray.sort(a,b => a.props.order - a.props.order)`
- Then I shifted the `LanguageButton` and realised that all that was left was the docs button, and that maybe we should be linking to the specific page. So.. instead of creating some new way of passing a URL to the docs button from any page and a way to order the buttons, I could just put a new docs button in each page 🤷 
- I also realised that adding a prop to order the components gets a bit messy if you're rendering buttons from different places, because to order the buttons you need to wrap the buttons and so you end up with a dom hierarchy like (hope this makes sense, noting that the actual `portal` is essentially inaccessible)
```
<container>
    <component which sorts children/>
    <component which sorts children/>
</container>
```
- Another option I just thought of now would be to put the `docs` button/default buttons into the `AppBarPortals`, and maybe make an `AppBarButton` component with an order. I think that would do it. This would still be rendering all the buttons in the same component so we'd be able to properly sort. Damn I think I should do this. Though, this would still mean you need to add an `AppBarPortals` component on every page 🤔


packages/common/src/hooks/useQueryParams/useQueryParams.test.tsx
```diff
@@ -14,6 +14,7 @@ type ThemeChangererProps = {
   paginationRowHeight: number;
   dataRowHeight: number;
   headerRowHeight: number;
+  footerHeight: number;
```
joshxg, line 17: offroad test fixes

packages/common/src/hooks/useRowRenderCount/useRowRenderCount.test.tsx
```diff
@@ -12,6 +12,7 @@ type ThemeChangererProps = {
   paginationRowHeight: number;
   dataRowHeight: number;
   headerRowHeight: number;
+  footerHeight: number;
```
joshxg, line 15: offroad test fixes

packages/common/src/ui/components/portals/AppBarButtons/AppBarButtons.test.tsx
```diff
@@ -0,0 +1,60 @@
+import React, { FC } from 'react';
+import { act, render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { AppBarButtons, AppBarButtonsPortal } from './AppBarButtons';
+
+describe('AppBarContent', () => {
```
mark-prins, line 6: `AppBarButtons` ...of course, I'd never have the wrong name on a test suite 🙄

mark-prins (approved):
👍  You're too hard on yourself - I don't think it's a really bad solution. it does everything it needs to and is neater to use, having the `AppBarButtons` portal. nice to have customised docs buttons.

I vote for keeping the ordering of buttons on the backburner - it's a good idea, just not sure if we'll need to have general buttons now

packages/common/src/ui/components/portals/AppBarButtons/AppBarButtons.test.tsx
joshxg: 🙈 

Hahha!

I'm being lazy and fixing this down the line

joshxg:
Thanks, I was more just hoping it would be a little easier to order them!

PR #222, opened 2021-10-04
#221 Opening drawer lag

### LINKED ISSUE #221: Framework: Too many re-renders when opening the drawer
When opening the drawer, the state is set to much for the dimensions of the app bar (I think), making the drawer open/close far to laggy

joshxg:
Fixes #221

There's still an issue with resizing the window trigger too many window dimension state changes which is annoying me. I thought to do that in a new issue, though.

### DIFF
--- a/packages/common/src/hooks/useDebouncedValue/useDebouncedValue.ts
+++ b/packages/common/src/hooks/useDebouncedValue/useDebouncedValue.ts
@@ -1,0 +1,15 @@
+import { useState, useEffect } from 'react';
+
+export const useDebouncedValue = <T>(value: T, delay?: number): T => {
+  const [debouncedValue, setDebouncedValue] = useState<T>(value);
+
+  useEffect(() => {
+    const timer = setTimeout(() => setDebouncedValue(value), delay || 500);
+
+    return () => {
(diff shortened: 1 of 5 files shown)


packages/common/src/hooks/useDebouncedValue/useDebouncedValue.ts
```diff
@@ -0,0 +1,15 @@
...
+  useEffect(() => {
+    const timer = setTimeout(() => setDebouncedValue(value), delay || 500);
+
+    return () => {
+      clearTimeout(timer);
+    };
+  }, [value, delay]);
+
+  return debouncedValue;
+};
```
joshxg, line 15: Very tempted to add a hooks utility library like https://react-hooks.org/ (10kb gzipped!). Maybe if we use a few more..

mark-prins: shame about the size! it's a neat library. 
We could import individual parts to keep the size down `@rooks/useDebouncedValue` but that feels like too much overhead! I'd just import the whole thing

mark-prins (approved):
👍  nice!

Issue #225, opened 2021-10-05
Outbound: Use a reducer for state


joshxg:
Looking to use a reducer for the state in the `useDocumentDraft` hook.

Think this in general just needs a bit of a change up but the general idea:

```
const {data} = useQuery(key, api);
const [state, dispatch] = useReducer(reducer(data));

return { merge(data, state) }
```

I'm not sure that merging every state change is a good idea and might just be better to merge on `data` change.

Would also like to be greedy and attach the dispatch to the state (at the top level and for each nested entity i.e. lines) so we don't need another mechanism to pass down callbacks.

PR #226, opened 2021-10-06
#225 reducer in use draft

joshxg:
Starts #225

- This is enough for this PR!
- I moved the `useDraftDocument` into a folder in `common/hooks`
- I added a reducer and attached an `updateQuantity` func to the row in a little neater way than what was previously happening.
- Think this is a lot tidier and simpler to understand, although the boiler plate typescript is a bit annoying

What we end up with:
- a hook `useDraftDocument` which takes a query key, an `Api` object (with `onRead` which returns some entity and `onUpdate` methods which takes the whole state and returns the same entity as `onRead`) and a reducer. Simply: `useDocumentDraft(['transaction', id], api, reducer)`. The reducer handles the state and the api handles all the.. api stuff. 
- Can do a lot with this which I'll continue on with. At the moment i'm wrapping entities in the document with a `updateQuantity` callback which is a dispatching an action. I'm not 100% sure about this but it makes things easier so running with it for now. I originally was going to just add the `dispatch` - the cell in this case decides which action to dispatch and was thinking that it might get messy? I could also see benefits to it, though (essentially, cells just emitting events for things which the reducer can start reading, seems pretty OK, actually!).
- Think there can be some simplifications with also doing the sorting in the reducer similar to how the cells are working.


packages/transactions/src/OutboundShipment/DetailView/reducer.ts
```diff
@@ -0,0 +1,72 @@
...
+        const row = state.items[rowIdx];
+
+        if (row) {
+          const newRow = { ...row, quantity };
+          const newItems = [...state.items];
+          newItems[rowIdx] = newRow;
+          const newState = { ...state, items: newItems };
+
+          return newState;
+        }
```
joshxg, line 67: immer 🙏

packages/transactions/src/OutboundShipment/DetailView/types.ts
```diff
@@ -0,0 +1,24 @@
+import {
+  Item,
+  Transaction,
+  DefaultDraftAction,
+} from '@openmsupply-client/common';
+
+export interface ItemRow extends Item {
+  updateQuantity: (quantity: number) => void;
```
mark-prins, line 8: nice!

mark-prins (approved):
I like that a lot! have also been thinking about handling of the draft document, this is a very nice way to handle it

packages/common/src/ui/layout/tables/columns/EditableQuantityColumn.tsx
```diff
@@ -54,7 +54,7 @@ export const getEditableQuantityColumn = <
         helperText="Incorrect value"
         hiddenLabel
         value={buffer}
-        onBlur={() => rowData.setQuantity(Number(rowKey), value)}
+        onBlur={() => rowData.updateQuantity(value)}
```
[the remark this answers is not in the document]
joshxg: Yes I am wondering if something like

```
interface ItemRow {
    updateQuantity: ..
    values: { }
}
```

might be more appropriate..

PR #230, opened 2021-10-08
#121 add item

### LINKED ISSUE #121: Outbound: Add Item
- Add the ability to add a row to an outbound

image

- Using the `Add Item` button?
- Did this idea get scapped?
image

mark-prins:
Fixes #121 

This has ended up larger than I wanted - though I've renamed a bit and added translations.. so it's not as bad as it looks!
Only the first part though, I'd like to get your opinion on this approach before adding stories for the new UI components, and then adding in the batch selection part of the modal form.

There's still an issue with kbar when entering the quantity, and the validation isn't working as I would like.
Otherwise, it's generally working.

Have added the `react-hook-form` package as I was creating those hooks myself, and in the end thought that the 2kb hit was worth it. happy to handle these sorts of forms another way if you prefer


packages/common/src/hooks/useFormContext/useFormContext.ts
```diff
@@ -0,0 +1,18 @@
+import create from 'zustand';
+import { Item } from '../..';
+
+type FormContext = {
+  item: Item;
+  setItem: (item: Item) => void;
+};
+
+export const useFormContext = create<FormContext>(set => ({
+  item: {
```
mark-prins, line 10: you had used a format for defining defaults for objects that I liked, and realise now that I forgot to do for this one!

packages/transactions/src/OutboundShipment/DetailView/modals/ItemDetails.tsx
```diff
@@ -0,0 +1,120 @@
...
+  const selectItem = (
+    _event: SyntheticEvent<Element, Event>,
+    value: Item | null
+  ) => {
+    setValue('id', value?.id || '');
+    setValue('code', value?.code || '');
+    setValue('name', value?.name || '');
+  };
+
+  register('id', { required: true });
```
mark-prins, line 85: being required, this should be affecting `isValid` but isn't doing what it should

packages/transactions/src/OutboundShipment/DetailView/DetailView.tsx
```diff
@@ -53,13 +56,24 @@ export const OutboundShipmentDetailViewComponent: FC = () => {
...
+  } = useForm({ mode: 'onBlur' });
+  const addItemClose = (item: Item) => {
+    addItem(item);
     hideDialog();
   };
-  const addAnotherItem = () => {
-    console.info('item added 😉');
+  const addItem = (item: Item) => {
+    // TODO: add to dataset and have the reducer add the fn
+    draft?.items?.push({ ...item, updateQuantity: () => {} });
```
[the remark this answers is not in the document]
joshxg: Cool. Sorry, have a few more thoughts on this one :)

At the moment, I'm thinking that we should make each `ItemRow`/`line` in the table/document similar to `TransactionItem` in mobile, where we group the batches by `Item` so this becomes more of a shape like:


```
interface Invoice {
 id: string;
 lines: InvoiceLine[];
}

interface InvoiceLine {
  batches: InvoiceBatch[];
  id: string [client assigned]
  totalQuantity: number
  setTotalQuantity: () => {} // assigns FEFO to batches
}

interface InvoiceBatch {
    quantity: number;
    expiry: string;
    setQuantity: () => {} // assigns quantity, needs to also update totalQuantity or not surpass totalQuantity, unsure what the desired editing behaviour is, but some logic
}
```

We could then pass in a copy of an `InvoiceLine` into the modal where editing the top level `quantity` field can adjust the batches below, and editing batches below, does.. something? Then `onSubmit`, we just pass the coped `InvoiceLine` back and merge that back into the document?

This is where i'm unclear on the editing behaviour. If we edit in-row, then we would want the `setTotalQuantity` to dispatch and action to the reducer, which will handle the FEFO logic. But, we would want slightly different behaviour for `setTotalQuantity` if we did this in the modal, where we wouldn't want the `draftDocument` to be updated, yet.

Another option I thought would be take a snapshot of the document when opening the modal, then just update the draft while editing in the modal and if cancelled, "just" apply the snapshot patch as a rollback.. I think this gives the 'maximum' code reuse, but I think makes any autosaving if needed in the future harder.

packages/transactions/src/OutboundShipment/DetailView/modals/ItemDetails.tsx
```diff
@@ -0,0 +1,120 @@
...
+  const renderOption = (
+    props: React.HTMLAttributes<HTMLLIElement>,
+    item: Item
+  ) => (
+    <ItemOption {...props} key={item.code}>
+      <span style={{ width: 100 }}>{item.code}</span>
+      <span style={{ width: 500 }}>{item.name}</span>
+      <span>0</span>
+    </ItemOption>
+  );
```
joshxg, line 74: the `filterOptions` and `renderOption` don't need to be in the component and can be defined outside of the scope

packages/transactions/src/OutboundShipment/DetailView/modals/ItemDetails.tsx
```diff
@@ -0,0 +1,120 @@
...
+        }
+      `
+    );
+    return items;
+  };
+
+  const t = useTranslation();
+  const { data, isLoading } = useQuery(['item', 'list'], listQueryFn);
+  const options =
+    data?.slice(0, 100).map(item => ({ label: item.name, ...item })) || [];
```
joshxg, line 59: Wow.. yup.. thats uhh... not ideal performance, huh.

I added a `limit: 100` into the `createFilterOptions` rather than the slice and it works just as well, but is using all of the options, rather than just the first `100`. I think this is likely fine and we could put a `Showing first 100/X results`? I don't really think having more than a couple hundred results is helpful to anyone, anyway.

packages/common/src/ui/forms/Modal/ModalAutocomplete.tsx
```diff
@@ -0,0 +1,78 @@
...
+  filterOptions,
+  loading,
+  loadingText,
+  noOptionsText,
+  onChange,
+  options,
+  renderInput,
+  renderOption,
+  width,
+}: PropsWithChildren<ModalAutocompleteProps<T>>): JSX.Element {
```
joshxg, line 55: Could be nice to expose the filter options as props `filterStringifiererer` etc and having the component construct the options rather than consumers having to import the filter factory also?

joshxg (approved):
Awesome.

`react-hook-form` would have been my pick for a form library (which we should be using as it's so annoying!) also.

I think it's great and my comments are really just some annoying brain dumps and thoughts on the future and moving forward in creating our component structure rather than anything really needing to be addressed here.

joshxg:
Re: KBar - I am wondering if it might just be better to remove it and add it back in the future, if wanted.. already been a bit of a time sink and I'm not sure if it's worthwhile (even though I love it)

packages/common/src/hooks/useFormContext/useFormContext.ts
mark-prins: ah. part of the version prior to `react-hook-form` removed in 98bd632

packages/transactions/src/OutboundShipment/DetailView/DetailView.tsx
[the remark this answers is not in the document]
mark-prins: the `InvoiceBatch` shape is good - that is what I wanted to do, having the item / line level and batches beneath it. that makes logical sense to me.

Is the problem with editing the support of `Cancel` / 'revert changes'?
We still want to assign batches base on FEFO in the modal: given a total quantity, assign quantities to batches as required. The user then adjusts as needed.

packages/transactions/src/OutboundShipment/DetailView/modals/ItemDetails.tsx
```diff
@@ -0,0 +1,120 @@
+import React, { SyntheticEvent } from 'react';
```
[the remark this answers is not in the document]
mark-prins: oof. good catch thanks.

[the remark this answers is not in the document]
mark-prins: 8dd3a25

packages/common/src/ui/forms/Modal/ModalInput.tsx
```diff
@@ -0,0 +1,32 @@
...
+  padding: '4px 8px',
+  width: '240px',
+};
+
+export const ModalInput: React.FC<ModalInputProps> = ({
+  defaultValue,
+  inputProps,
+}) => {
+  return (
+    <Grid item xs={10} sx={{ marginBottom: '4px' }}>
```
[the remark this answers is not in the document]
mark-prins: nice. have kept the mui default - `2` is pretty small for a gap I think, but you can always correct this later :)

packages/transactions/src/OutboundShipment/DetailView/DetailView.tsx
```diff
@@ -53,13 +56,24 @@ export const OutboundShipmentDetailViewComponent: FC = () => {
...
   const { success, warning } = useNotification();
-  const addItem = () => {
-    console.info('item added 😉');
+  const {
+    register,
+    reset,
+    handleSubmit,
+    setValue,
+    formState: { isDirty, isValid },
+  } = useForm({ mode: 'onBlur' });
```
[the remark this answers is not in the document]
mark-prins: yep! really need to slim down that `DetailView` and was expecting to add a `BatchesTable` to this form

packages/transactions/src/OutboundShipment/DetailView/modals/ItemDetails.tsx
mark-prins: ah yep.

packages/common/src/ui/forms/Modal/ModalAutocomplete.tsx
mark-prins: yeah, much better. will do

mark-prins:
@joshxg re kbar - found that this is only when I click the chrome suggestion for the input field value, based on history. There's a simple null check fix needed, have raised an issue. bit of an edge case, I think, so not enough reason to dump it I think.

PR #231, opened 2021-10-08
Add mock server refactor

joshxg:
A start. More than I wanted to do in a single PR but decided to just get it done. Sorry. Hopefully it's consistent enough to be easy enough to grok.

This is just the graphql api, plan to add the `mock` handlers into here so that we can reuse the `Resolver/MutationService` things


packages/mock-server/src/data/data.js
```diff
@@ -0,0 +1,132 @@
...
+        itemId,
+      };
+
+      quantityToUse = quantityToUse - availableNumberOfPacks;
+
+      stockLines.push(stockLine);
+    }
+  });
+
+  return stockLines.flat();
```
mark-prins, line 50: what was the reason for flattening here? `stockLines` is only a single level

mark-prins (approved):
👍  awesome!

PR #232, opened 2021-10-10
#batches to mock server 2

joshxg:
Sigh

I have a problem. Sorry.

- All the mocks are in `mock-server`. Maybe `mocks` was a better name.
- I was having issues using the mocks in typescript as it was in JS, so I just thought i'd make it all typescript. 🙄  I found lots of bugs.. double 🙄
- Made the app work with the mock service worker, which is now using all the same resolvers as the mock server.
-


mark-prins (approved):
that's a lot of changes! thanks again, typescript 😃 
the mock server is looking impressive! nice to have the service worker on board now too!

PR #233, opened 2021-10-10
#batches to mock server 3

### LINKED ISSUE #229: Update mock server to handle batches
- Update the mock server to handle batches
- Also would be good to try and share the msw and mock implementations so we don't need to keep updating both
- Would be good if this was a little neater so making changes in the future was easier
- If not too much work, could add names and also rename all `transaction` things to `invoice` as `transaction` is not a thing anymore.

joshxg:
Fixes #229 

- Adds mutations for invoice lines which will adjust the underlying stockline quantity
- Other bug fixes found from the other PRs which I should have done in those ..


packages/mock-server/src/data/data.ts
```diff
@@ -104,7 +106,7 @@ const createInvoiceLines = (
 };
 
 const createItems = (
-  numberToCreate = Math.ceil(Math.random() * 10)
+  numberToCreate = Math.ceil(Math.random() * 10) + 10
```
mark-prins, line 109: `utils.randomInteger` perhaps?

packages/mock-server/src/schema/invoiceLine.ts
```diff
@@ -9,6 +13,76 @@ const Types = `
...
+const Queries = `
+    invoiceLines(first: Int, offset: Int, sort: String, desc: Boolean): InvoiceLineResponse
+    invoiceLine(id: String!): InvoiceLine
+`;
+
+const Mutations = `
+    updateInvoiceLine(invoiceLine: InvoiceLinePatch): InvoiceLine
+    insertInvoiceLine(invoiceLine: InvoiceLinePatch): InvoiceLine
+    deleteInvoiceLine(invoiceLine: InvoiceLinePatch): InvoiceLine
+    deleteInvoiceLines(invoiceLine: [InvoiceLinePatch]): [InvoiceLine]
```
mark-prins, line 32: :nit I'd prefer the parameter to be `invoiceLines`

mark-prins (approved):
👍  looks good!

Issue #235, opened 2021-10-10
Creating a new outbound flow


joshxg:
- Add a new outbound shipment via selecting a customer first as per this image:

![image](https://user-images.githubusercontent.com/35858975/136714994-3e416696-fda8-4717-b1d3-534e3462bfe7.png)

- And the flow described here, under "Create an ad hoc OS":

https://docs.google.com/spreadsheets/d/1H0Rv1DEOgW0cE07wpfc_KivLf8TcNhYDqlwRp92x-To/edit#gid=890672777

@mark-prins 

Technical questions:
- We've had this discussion before, but should the invoice be created before navigating? The discussion previously I don't remember there being a modal to select a customer from prior to creating the invoice, so with the modal, it makes more sense that the invoice is created?
- I think it is just a lot easier to just send a request to create and navigate to the correct invoice number

mark-prins:
somehow I'd blanked out the suggestion of a modal here - I had in mind a blank customer field instead.
This way allows for creation of a customer, which I think is the motivation. ( the mock up really needs a title 'select a customer'! as well as alignment between 'inbound' and 'customer' .. this should be 'new outbound shipment' )

Sure - if we are using a modal at this stage, then create and navigate to the invoice id

joshxg:
Cool.

Have had a quick look and I think just using `<AutoComplete open={true}/>` will do this just fine.. I think there might often be too many visible customers for some people which means we might need to do server side filtering, I think an initial version using client side filtering is probably OK?

This sandbox has a randomly generated list of 10,000 items working very quickly, but requires ~10kb of virtualisation packages .. https://codesandbox.io/s/virtualize-material-demo-forked-le8vz?file=/demo.js

This is all working with the autocomplete which I think you might be touching that area at the moment? I'll start with adding names to the api, first

PR #242, opened 2021-10-11
#235 new outbound flow

josh-griffin:
Starts #235!

- Just adds `Name` to the mock api and a query from the app to get the names list


mark-prins (approved):
🎉  nice!

PR #244, opened 2021-10-12
#235 new outbound flow 3

### LINKED ISSUE #235: Creating a new outbound flow

joshxg:
Fixes #235 

- Adds a `CustomerSearch` component wrapper around `ListSearch` - unsure where to put it 🤔 
- Hooks up the customer search to be how we create an invoice now

### DIFF
--- a/packages/transactions/src/OutboundShipment/ListView/CustomerSearch.tsx
+++ b/packages/transactions/src/OutboundShipment/ListView/CustomerSearch.tsx
@@ -1,0 +1,34 @@
+import React, { FC } from 'react';
+import { ListSearch, Name, useQuery } from '@openmsupply-client/common';
+import { nameListQueryFn } from '../../api';
+
+interface CustomerSearchProps {
+  open: boolean;
+  onClose: () => void;
+  onChange: (name: Name) => void;
+}
+
+// TODO: Would be better to disable this query until the button to open the modal
+// has been hovered, so we could still pre-fetch the data but not fetch if the user
+// is looking at existing invoices.
+export const CustomerSearch: FC<CustomerSearchProps> = ({
+  open,
+  onClose,
+  onChange,
+}) => {
+  const { data, isLoading } = useQuery(['names', 'list'], nameListQueryFn);
+
+  return (
+    <ListSearch
+      loading={isLoading}
+      open={open}
+      options={data?.data ?? []}
+      onClose={onClose}
+      title="label.customer"
+      optionKey="name"
+      onChange={(_, name: Name | null) => {
+        if (name) onChange(name);
+      }}
+    />
+  );
+};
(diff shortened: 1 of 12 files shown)


packages/transactions/src/OutboundShipment/ListView/CustomerSearch.tsx
```diff
@@ -0,0 +1,34 @@
...
+      options={data?.data ?? []}
+      onClose={onClose}
+      title="label.customer"
+      optionKey="name"
+      onChange={(_, name: Name | null) => {
+        if (name) onChange(name);
+      }}
+    />
+  );
+};
```
joshxg, line 34: Was thinking of going further and doing a `const {CustomerSearch, onOpen } = useCustomerSearch()` but wasn't sure if it was going to far heh

packages/transactions/src/OutboundShipment/ListView/ListView.tsx
```diff
@@ -117,8 +120,31 @@ export const OutboundShipmentListViewComponent: FC = () => {
...
+
+            const result = await onCreate(invoice);
+
+            invalidate();
+            navigate(`/customers/customer-invoice/${result.invoiceNumber}`);
+          };
+
+          createInvoice();
+        }}
+      />
```
joshxg, line 147: With throttling it feels as if there needs to be something to give feedback. Maybe a blocking modal (which we might need for finalising, anyway). Thought to do it later, but something to consider?

packages/transactions/src/OutboundShipment/ListView/CustomerSearch.tsx
joshxg: Hm yes. Likely a lot of things. Are you suggesting to do the hook or another approach?

Few ways .. I had also thought that `ListSearch` could be used further into a `AsyncListSearch` and to pass a query function and query key 🤷

joshxg:
It does seem quite ugly without any lines as of now. Thought rendering the table with the headers would be good. Thought rendering empty lines like mSupply does now would look OK, but then remembered we don't use alternating background colours, so maybe not!

Maybe a list with headers and a extra redundant button: "add a line"

mark-prins:
and if possible, the background colour of the text on hover

PR #245, opened 2021-10-13
#219 second level left nav

### LINKED ISSUE #219: Framework: left nav for second level items
Clarified with Manu the intention for second level menu items - turns out that I had that incorrect.
What he was intending was:
If the drawer is collapsed, hovering over a top level menu item which has lower level items under it would expand the drawer, showing the lower levels.

The complication is that the drawer would have to remain open only while that menu set is being interacted with and collapse again when a top level item (with no children) is hovered over.

Starting with:
Screen Shot 2021-10-04 at 9 24 09 AM

Hover over `Customers`:
Screen Shot 2021-10-04 at 9 24 22 AM
 
The move the mouse and hover over `Invoices`
Screen Shot 2021-10-04 at 9 24 34 AM

Then move down and hover over `Suppliers`
Screen Shot 2021-10-04 at 9 24 49 AM

mark-prins:
Fixes #219
Am using the hover event fired by the tooltip to expand and collapse the drawer. I've removed the standard navigation behaviour of top level menu items, so that the `Customers` menu item now only expands the menu on click, rather than navigating to a blank page. Having trouble testing this on mobile, so will test the theory when on the demo site.

Have debounced the drawer close, so that it doesn't close when you hop around to the lower level menu items of an expanding section.

### DIFF
--- a/packages/common/src/ui/components/navigation/NavLink/NavLink.test.tsx
+++ b/packages/common/src/ui/components/navigation/NavLink/NavLink.test.tsx
@@ -74,4 +74,32 @@ describe('NavLink', () => {
     expect(node).toBeInTheDocument();
     expect(node).toHaveAttribute('href', '/customer-invoice');
   });
+
+  it('Correctly renders a span in place of a link for top level hover items', () => {
+    const { getByTestId } = render(
+      <TestingProvider>
+        <TestingRouter initialEntries={['/customers']}>
+          <Route
+            path="*"
+            element={
+              <Box>
+                <Wrapper collapsed />
+                <NavLink
+                  to="/customer-invoice"
+                  icon={<Customers />}
+                  text="Customers"
+                  end={false}
+                  expandOnHover
+                />
+              </Box>
+            }
+          />
+        </TestingRouter>
+      </TestingProvider>
+    );
+
+    const node = getByTestId('/customer-invoice_hover');
+
+    expect(node).toBeInTheDocument();
+  });
 });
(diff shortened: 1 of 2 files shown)


packages/common/src/hooks/useDrawer.ts
```diff
@@ -2,20 +2,32 @@ import create from 'zustand';
 import LocalStorage from '../localStorage/LocalStorage';
 
 type DrawerController = {
+  hoverActive: Record<string, boolean>;
```
mark-prins, line 5: a little clumsy, but doesn't need interacting with so I thought I could get away with it?

packages/common/src/ui/components/navigation/NavLink/NavLink.test.tsx
```diff
@@ -74,4 +74,32 @@ describe('NavLink', () => {
     expect(node).toBeInTheDocument();
     expect(node).toHaveAttribute('href', '/customer-invoice');
   });
+
+  it('Correctly renders a span in place of a link for top level hover items', () => {
```
mark-prins, line 78: very lame test, I was going to get all fancy and add hover checking.. but that was taking a while and I wasn't convinced of the value.

packages/common/src/ui/components/navigation/NavLink/NavLink.tsx
```diff
@@ -48,25 +49,82 @@ const StyledListItem = styled<
...
   const drawer = useDrawer();
   const selected = useSelectedNavMenuItem(to, !!end);
 
   const CustomLink = React.useMemo(
     () =>
-      React.forwardRef<HTMLAnchorElement>((linkProps, ref) => (
-        <Link {...linkProps} ref={ref} to={to} role="link" aria-label={text} />
-      )),
+      React.forwardRef<HTMLAnchorElement>((linkProps, ref) =>
+        expandOnHover && !end ? (
```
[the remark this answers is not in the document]
joshxg: Will there be top level nav items here with no sub menu? Maybe the dashboard? It would be even weirder if everything looked the same, and even the placement couldn't inform you if something was clickable

PR #253, opened 2021-10-14
#248 add item validation

### LINKED ISSUE #248: Add item: validation bug
The validation on the add item modal is misbehaving and won't enable the OK button when values are selected.

mark-prins:
Fixes #248 

Fixed the validation 
- have added a `required` prop to the `quantity` 
- added a trigger for the `id` validation - this was required but was not updating properly after item selection

Went a little further and added in a simple error message to show how that might look.

Couple of offroad fixes - noticed that the console was getting a little busy with warnings:

#### #1
```
react_devtools_backend.js:2526 Warning: Failed prop type: The prop `xs` of `Grid` can only be used together with the `item` prop.
at Grid (webpack://@openmsupply-client/host/../../node_modules/@mui/material/Grid/Grid.js?:271:85)
    at ModalLabel (webpack://@openmsupply-client/host/../common/src/ui/forms/Modal/ModalLabel.tsx?:23:26)
```

#### #2
```
Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?

Check the render method of `ModalInput`.
    at BasicTextInput (webpack://@openmsupply-client/host/../common/src/ui/components/inputs/TextInput/BasicTextInput.tsx?:91:21)
    at div
```


packages/common/src/ui/forms/Modal/ModalErrorMessage.tsx
```diff
@@ -0,0 +1,28 @@
+import * as React from 'react';
```
joshxg, line 1: This is really similar to just using the `helperText` prop with a few style changes. Maybe we just use that in `ModalInput`?

https://codesandbox.io/s/validationtextfields-material-demo-forked-inzd9?file=/demo.js

packages/common/src/ui/forms/Modal/ModalErrorMessage.tsx
```diff
@@ -0,0 +1,28 @@
...
+  errors?: FieldErrors;
+  name: string;
+  message?: Message;
+};
+export const ModalErrorMessage: React.FC<ModalErrorMessageProps> = ({
+  errors,
+  name,
+  message,
+}) => {
+  const methods = useFormContext();
```
joshxg, line 15: Think I'm missing a trick here.

What *I* think is happening, can you correct where I'm wrong:

You use `useFormState` in the parent, which I think is equal to `useFormContext().formState` and pass down `errors` which is equal to `useFormContext.formState.errors` and then do the long version and pass both values which are equal into `get`?

packages/transactions/src/OutboundShipment/DetailView/modals/ItemDetails.tsx
```diff
@@ -107,7 +100,10 @@ export const ItemDetails: React.FC<ItemDetailsProps> = ({
           />
         </ModalRow>
         <ModalInputRow
-          inputProps={register('quantity')}
+          inputProps={register('quantity', {
+            required: true,
+            min: { value: 1, message: t('error.greater-than-zero-required') },
```
joshxg, line 105: this is valid for `a` etc. Making the input a number type I suppose? Or regex or something for validation?

mark-prins: sure. have added a pattern attribute to validate numeric input

PR #256, opened 2021-10-14
#250 deleting invoices doesnt work

### LINKED ISSUE #250: Bug: deleting invoices doesn't work
I changed the api to be more like what I expect from the backend - a single `deleteInvoice` mutation rather than what we had been using: `deleteInvoices`

I forgot and this broke deleting invoices in the list view.

Fortunately a PR I was waiting for in graphql-request was merged and a release cut 7 hours ago which will make this nice and easy.

https://github.com/prisma-labs/graphql-request/releases

Which means that after upgrading, instead of some cumbersome string creation, we will be able to simple do something like:

```
const data = await batchRequests(endpoint, [
    { document, variables },
    { document, variables },
    { document, variables },
])
```

josh-griffin:
Fixes #250


packages/mock-server/src/api/mutations.ts
```diff
@@ -76,18 +77,24 @@ export const update = {
...
     });
 
-    return db.remove.invoice(invoice);
+    return db.remove.invoice(invoiceId);
   },
-  invoiceLine: (invoiceLine: InvoiceLine): InvoiceLine => {
+  invoiceLine: (invoiceLineId: string): string => {
+    const invoiceLine = ResolverService.byId.invoiceLine(invoiceLineId);
+
+    if (!invoiceLine.stockLineId) {
```
mark-prins, line 92: Not required?

packages/mock-server/src/worker/handlers.ts
```diff
@@ -28,16 +28,14 @@ const insertInvoice = graphql.mutation(
 );
 
 const deleteInvoices = graphql.mutation(
```
mark-prins, line 30: :nit this should now be `deleteInvoice`

mark-prins (approved):
LGTM!

It looks all good - I'm having trouble running this branch, the `mock-server` is complaining about the `window` global in config which seems strange. I thought we could merge and I'll revisit later if that's still a problem

PR #227, opened 2021-10-07
Add test fixes

joshxg:
Crap, I forgot to push before merging

Issue #237, opened 2021-10-10
Outbound: customer input


joshxg:
- Add an input to change the customer

![image](https://user-images.githubusercontent.com/35858975/136716287-b56373eb-1452-494c-9d44-318c8817c9fe.png)

- Unsure if the pencil makes the input editable and then an autocomplete dropdown while you're typing?
- Shouldn't be able to be set as an invalid customer

PR #257, opened 2021-10-15
#237 customer input

joshxg:
Fixes #227 

image

Not sure if the styles are quite right here but thought to stop here. Will do something very similar for the ref input, so figured heres a good place to stop and get feedback before moving forward.


packages/common/src/ui/components/inputs/TextInput/InputWithLabelRow.tsx
```diff
@@ -0,0 +1,18 @@
...
+interface InputWithLabelRowProps {
+  Input: ReactNode;
+}
+
+export const InputWithLabelRow: FC<InputWithLabelRowProps> = ({
+  Input = <BasicTextInput />,
+}) => {
+  return (
+    <Grid display="flex" alignItems="center" gap={1}>
+      <FormLabel sx={{ fontWeight: 'bold' }}>Customer name:</FormLabel>
```
mark-prins, line 14: was going to point out that it needs a translation, but then noticed this is a shared component.. so will need the label to be passed in

packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
```diff
@@ -210,12 +212,34 @@ export const OutboundShipmentDetailViewComponent: FC = () => {
...
+        </Box>
+
+        <Box display="flex" flex={1} justifyContent="center">
+          <TabList value={currentTab} onChange={onChangeTab}>
+            <Tab value="general" label={t('label.general')} />
+            <Tab value="transport" label={t('label.transport')} />
+          </TabList>
+        </Box>
+
+        <Box display="flex" flex={1} justifyContent="flex-end" />
```
mark-prins, line 242: Is it time to put a save and cancel button on the bottom of this component 🤔

packages/mock-server/src/worker/handlers.ts
```diff
@@ -108,7 +121,7 @@ export const serverError = graphql.query('error500', (_, response, context) =>
  * handler.
  */
 const batchMutationHandler = rest.post(
-  Environment.API_URL,
+  'http://localhost:4000',
```
mark-prins, line 124: are you having issues with the config too?

packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
joshxg: Ha yes I think so!

joshxg:
Seems what I've added in the MSW to handle batched mutations means the items query wasn't falling through and hitting the server. So i've added an item handler in the worker for now and will look at doing a better solution for the mock worker.

Also fixed up the nits in the other PR in this one

Issue #252, opened 2021-10-13
Add item: Batch table


josh-griffin:
Batch table for the add item modal as per:

![image](https://user-images.githubusercontent.com/35858975/137222531-a8fb3270-fd49-4266-92bb-1350a3ca1a52.png)

Would really have preferred to use the same table component. The only real difference seems to be the row separation line 🤷 

Tasks that I can see:

- Add more to the mock server: a way to query invoice lines for an item. Oops should have added. Maybe add cost/sell prices and additional info to show in the `i`
- Add an `InvoiceItem` interface, where `InvoiceLine` are grouped by into the parent `InvoiceItem`. `InvoiceItem` has an aggregated quantity of each of the `InvoiceLine`s, and has functionality to FEFO dispense to each of it's children. Children will need to update the parent with new quantities, also. 
- Add the popover info column

PR #258, opened 2021-10-15
#252 batch table

mark-prins:
Start of #252 

This part enhances the mock server a little so that the shape of the `items` query is the same as [in this API doc]( https://dev.msupply.foundation/docs/api/queries/#items)

Then displays the StockLines in a table, like so:

Screen Shot 2021-10-15 at 3 26 09 PM

The allocation of a requested quantity across the lines hasn't been done, nor the info popover, or the gathering up of the entered information into an invoice line that can be saved.


packages/invoices/src/OutboundShipment/DetailView/modals/ItemBatches.tsx
```diff
@@ -0,0 +1,143 @@
...
+      padding: '0 8px',
+      whiteSpace: 'nowrap',
+      ...sx,
+    }}
+  />
+);
+
+export const ItemBatches: React.FC<ItemBatchesProps> = ({ item }) => {
+  if (!item) return null;
+
```
joshxg, line 94: Would be a good place to put a query later. If there are a lot of items we likely don't want to be getting all the batches also, so we could have a query where the `item.id` is a part of the query key

joshxg (approved):
Wow it really is starting to look like a real app!

It's super super cool.

Time for me to go and see what the hell a placeholder really is in mSupply.. 🤣

PR #263, opened 2021-10-16
Add DetailViewToolbar component and added their ref input

### LINKED ISSUE #236: Outbound: their ref
Add an input for `their ref` on the outbound shipment detail view

- image
image

Unsure if the pencil icon is supposed to make the input editable

joshxg:
Fixes #236 

Still a bit messy but started to take some things out of the `DetailView`

### DIFF
--- a/packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
+++ b/packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
@@ -2,7 +2,6 @@ import React, { FC, useEffect } from 'react';
 import { useParams } from 'react-router';
 import {
   AppBarButtonsPortal,
-  AppBarContentPortal,
   Book,
   Box,
   Button,
@@ -19,9 +18,7 @@ import {
   PanelRow,
   PlusCircle,
   Rewind,
-  Tab,
   TabContext,
-  TabList,
   TabPanel,
   TableProvider,
   Typography,
@@ -36,7 +33,6 @@ import {
   useNotification,
   useTabs,
   useTranslation,
-  InputWithLabelRow,
 } from '@openmsupply-client/common';
 import { reducer, OutboundAction } from './reducer';
 import { getOutboundShipmentDetailViewApi } from '../../api';
@@ -44,7 +40,7 @@ import { GeneralTab } from './tabs/GeneralTab';
 import { ItemDetails } from './modals/ItemDetails';
 import { ExternalURL } from '@openmsupply-client/config';
 import { ItemRow } from './types';
-import { CustomerSearchInput } from '../CustomerSearchInput';
+import { OutboundShipmentToolbar } from './DetailViewToolbar';
 
 const useDraftOutbound = () => {
   const { id } = useParams();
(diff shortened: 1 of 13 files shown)


packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
```diff
@@ -211,37 +207,11 @@ export const OutboundShipmentDetailViewComponent: FC = () => {
...
-            <Tab value="general" label={t('label.general')} />
-            <Tab value="transport" label={t('label.transport')} />
-          </TabList>
-        </Box>
-
-        <Box display="flex" flex={1} justifyContent="flex-end" />
-      </AppBarContentPortal>
+      <OutboundShipmentToolbar
+        draft={draft}
+        onChangeTab={(val: string) => onChangeTab(val)}
```
mark-prins, line 212: I don't see why the need to keep `onChangeTab` in the `DetailView` - it is only used in the toolbar. Are you planning for something later?

packages/invoices/src/OutboundShipment/DetailView/DetailViewToolbar.tsx
```diff
@@ -0,0 +1,83 @@
...
+import { CustomerSearchInput } from '../CustomerSearchInput';
+import { OutboundShipment } from './types';
+
+interface OutboundShipmentToolbarProps {
+  draft: OutboundShipment;
+  currentTab: string;
+  onChangeTab: (newTab: string) => void;
+}
+
+export const OutboundShipmentToolbar: FC<OutboundShipmentToolbarProps> = ({
```
mark-prins, line 20: The name of the component and the filename don't align - and I think `OutboundShipmentToolbar` isn't specific enough for the detail view toolbar. It won't be used on the list view and I can't think where else it will be needed.

packages/invoices/src/OutboundShipment/DetailView/DetailViewToolbar.tsx
```diff
@@ -0,0 +1,83 @@
...
+  return (
+    <AppBarContentPortal
+      sx={{
+        display: 'flex',
+        flex: 1,
+        justifyContent: 'center',
+        paddingLeft: '25px',
+      }}
+    >
+      <Box
```
mark-prins, line 36: In the zeplin design, the inputs are above the tab controls, which allows them to spread horizontally and on smaller screens (like my laptop screen) the 'customer name' label shows on one line.

With the three box layout you have, the inputs and their labels are squished up, and the label wraps to two lines. Could you instead run a full width box for the inputs, then three boxes under that for the 'columns' and 'tab' controls? 

my screen:
Screen Shot 2021-10-18 at 9 28 36 AM

zeplin:
Screen Shot 2021-10-18 at 9 28 25 AM

mark-prins (approved):
Nice!
I have a preference for the layout which I think will look better. Great to have the invoice coming together!

packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
joshxg: The current tab is used for the tab context which is in the detail view. Could put the context into the toolbar and wrap all the content in it?

packages/invoices/src/OutboundShipment/DetailView/DetailViewToolbar.tsx
joshxg: Uhh.. soo.. `OutboundShipmentDetailViewToolbar` ? 😁

packages/invoices/src/OutboundShipment/DetailView/DetailViewToolbar.tsx
joshxg: Oh right ! 

image

Hope this is a bit better. Need to add the filters and columns buttons?

mark-prins: another time.. I suspect that could be a big job

packages/invoices/src/OutboundShipment/DetailView/DetailViewToolbar.tsx
mark-prins: sigh.

packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
mark-prins: ah, yeah. missed that. Could do, though it doesn't help too much. just wanting to slim down the `DetailView`

Issue #262, opened 2021-10-15
Outbound: Add save buttons


joshxg:
Need to add the save buttons to the outbound page

![image](https://user-images.githubusercontent.com/35858975/137551612-a424b611-6801-446d-8b7f-9dd4657664aa.png)

PR #265, opened 2021-10-17
#262 save buttons

joshxg:
#262 

... 🙄 

I went to start adding the save buttons and then nek minit here we are. Started to use the material variants `outlined` v `contained` to handle the full color/not full colour variants. 

- I moved all the oval style buttons into a standard folder and tried to base them all off `BaseButton` which handles the styling.. then the ShrinkableButton is based off that, and the `ButtonWithIcon` from that and a `DialogButton` from that.. 
- The icons weren't playing well - when using the colour on the button if using a default color for the icon then the color doesn't change - so I changed a few to not use a default. Didn't want to go through them all will do so in a follow up PR. I think this is wrestling less with mui?
- Added a story to show off all the different buttons


packages/common/src/hooks/useDebounce/useDebouncedValue.stories.tsx
```diff
@@ -26,18 +27,12 @@ const Value: Story = () => {
...
-        </Button>
-        <Button
-          variant="contained"
-          onClick={() => setNumber(state => (state -= 1))}
-        >
+        </BaseButton>
+        <BaseButton onClick={() => setNumber(state => (state -= 1))}>
           Decrement
-        </Button>
+        </BaseButton>
```
joshxg, line 35: Can just use our `BaseButton` now and it's styled as the contained, primary variant (maybe the outlined variant is better 🤷 )

packages/common/src/hooks/useDocument/useDocument.ts
```diff
@@ -62,11 +61,7 @@ export const useDocument = <
   // document state and manage the communication with the server from that.
   const { mutateAsync } = useMutation(api.onUpdate, {
     // TODO: onError: should dispatch some DefaultDocumentAction with errors for the reducer to handle.
-    onSuccess: (data, variables) => {
-      if (variables.id) {
-        navigate({ pathname: `../${data.id}` }, { replace: true });
-      }
-
+    onSuccess: () => {
```
joshxg, line 64: We're creating the invoice before navigating now so we don't need this

packages/common/src/hooks/useQueryParams/useQueryParams.test.tsx
```diff
@@ -15,6 +15,7 @@ type ThemeChangererProps = {
   dataRowHeight: number;
   headerRowHeight: number;
   footerHeight: number;
+  saveButtonRowHeight: number;
```
joshxg, line 18: I think this is a bit of a problem .. at the moment i've not adjusted further for the difference between list and detail views (not have the save button row).

I also find managing these rows to just generally be an annoying to handle, constantly update and the weird ui jankiness it causes is a bit horrible. Maybe there's a better solution for this,  but I am not sure anymore what we are really gaining by not just scrolling the rows in the table 🤷

packages/common/src/hooks/useQueryParams/useQueryParams.test.tsx
```diff
@@ -43,7 +46,8 @@ describe('useQueryParams', () => {
       dataRowHeight = 10,
       headerRowHeight = 0,
       paginationRowHeight = 0,
-      footerHeight = 0
+      footerHeight = 0,
+      saveButtonRowHeight = 0
```
joshxg, line 50: test updates

packages/common/src/hooks/useRowRenderCount/useRowRenderCount.stories.tsx
```diff
@@ -19,7 +19,10 @@ const Example = () => {
 
   return (
     <Box>
-      <Box bgcolor={theme.palette.darkGrey} height={table.headerRow.height}>
+      <Box
+        bgcolor={theme.palette.darkGrey.main}
+        height={table.headerRow.height}
+      >
```
joshxg, line 25: testupdates

packages/common/src/hooks/useRowRenderCount/useRowRenderCount.test.tsx
```diff
@@ -13,6 +13,7 @@ type ThemeChangererProps = {
   dataRowHeight: number;
   headerRowHeight: number;
   footerHeight: number;
+  saveButtonRowHeight: number;
```
joshxg, line 16: testu pdates

packages/common/src/hooks/useRowRenderCount/useRowRenderCount.ts
```diff
@@ -12,13 +12,15 @@ export const useRowRenderCount = (): number => {
   const headerRowHeight = mixins.table.headerRow.height;
   const paginationRowHeight = mixins.table.paginationRow.height;
   const pageFooterHeight = mixins.footer.height;
+  const detailViewFooter = mixins.saveButtonRow.height;
```
joshxg, line 15: test updates

packages/common/src/hooks/useSortBy/useSortBy.stories.tsx
```diff
@@ -19,12 +20,12 @@ const Template: Story = () => {
...
-        <button onClick={() => onChangeSortBy({ key: 'id' })}>
+        <BaseButton onClick={() => onChangeSortBy({ key: 'id' })}>
           Sort by ID!
-        </button>
-        <button onClick={() => onChangeSortBy({ key: 'quantity' })}>
+        </BaseButton>
+        <BaseButton onClick={() => onChangeSortBy({ key: 'quantity' })}>
           Sort by Quantity!
-        </button>
+        </BaseButton>
```
joshxg, line 28: Using the base button in various stories because I wanted to make the diff really big, apparently.

packages/common/src/intl/locales/ar.json
```diff
@@ -51,9 +53,9 @@
   "label.color": "Colour",
   "label.comment": "Comment",
   "label.confirmed": "Confirmed",
+  "label.cost": "Cost",
```
joshxg, line 56: How to make a diff bigger: step1) update a bunch of stories with unneeded work. step 2) Alphabetically sort some lookup tables. Step3 ) make sure you do both in the same pr

packages/common/src/styles/theme.ts
```diff
@@ -41,10 +44,25 @@ declare module '@mui/material/styles/createMixins' {
...
+}
+
+declare module '@mui/material/styles' {
+  interface Palette {
+    darkGrey: Palette['primary'];
+  }
+  interface PaletteOptions {
+    darkGrey: PaletteOptions['primary'];
+  }
+}
```
joshxg, line 60: Because I started using primary/secondary colours as orange/blue, this was ruined. So this adds the `darkGrey` to the `color` prop of the checkbox 🤷

packages/common/src/styles/theme.ts
```diff
@@ -103,14 +122,16 @@ const themeOptions = {
     },
   },
   palette: {
-    darkGrey: '#555770',
+    darkGrey: { main: '#555770' },
```
joshxg, line 125: Because I did this, I then had to also go and update all the references to this colour

packages/common/src/ui/components/buttons/ButtonShowcase.stories.tsx
```diff
@@ -0,0 +1,153 @@
+import React, { FC } from 'react';
+import { Grid, Paper, Typography } from '@mui/material';
+import { Story } from '@storybook/react';
+import { FlatButton } from './FlatButton';
+import { Book } from '../../icons';
+import { BaseButton, ButtonWithIcon } from '.';
+import { Customers } from '../../icons';
+import { DialogButton, IconButton } from '..';
+
+const getOnClick = (someText: string) => () => {
```
joshxg, line 10: I really liked what you did with the icons and tried to do similarly..

packages/common/src/ui/components/buttons/IconButton.stories.tsx
```diff
@@ -2,21 +2,24 @@ import React from 'react';
...
+import { Book, SvgIconProps } from '../../icons';
 
-const Template: ComponentStory<typeof IconButton> = () => (
+const Template: ComponentStory<React.FC<SvgIconProps>> = args => (
   <Box>
     <IconButton
-      icon={<Book />}
-      titleKey="button.docs"
+      icon={<Book {...args} />}
+      labelKey="button.docs"
```
joshxg, line 11: It seemed like half the buttons were using `labelKey` and half `titleKey` (Although it might have actually been more in favour of the one I didn't choose) so I tried to consolidate on a single one

packages/common/src/ui/components/buttons/IconButton.tsx
```diff
@@ -1,36 +1,33 @@
...
+export const IconButton: React.FC<ButtonProps> = ({
+  disabled,
+  icon,
+  onClick,
+  labelKey,
+}) => {
   const t = useTranslation();
-  const { disabled, icon, onClick, size, titleKey } = props;
+
+  const tooltip = t(labelKey);
```
joshxg, line 20: It seemed like half the buttons were using `labelKey` and half `titleKey` (Although it might have actually been more in favour of the one I didn't choose) so I tried to consolidate on a single one

packages/common/src/ui/components/buttons/IconButton.tsx
```diff
@@ -1,36 +1,33 @@
 import React from 'react';
-import { Button, Tooltip } from '@mui/material';
-import { styled } from '@mui/material/styles';
+import { IconButton as MuiIconButton, Tooltip } from '@mui/material';
```
joshxg, line 2: I made the `IconButton` into `ButtonWithIcon` and `UnstyledIconButton` into `IconButton` as it made more sense to me at the time..

packages/common/src/ui/components/buttons/standard/BaseButton.tsx
```diff
@@ -0,0 +1,53 @@
...
+      '&:hover': {
+        border: 'none',
+        color: hoverColor,
+        backgroundColor: hoverBgColor,
+      },
+    };
+  }
+);
+
+export const BaseButton: React.FC<MuiButtonProps> = ({ ...rest }) => {
```
joshxg, line 44: This has taken all the default styling and put it into a base which also manages the slightly different hover behaviour we are using which is then used to compose the other buttons

packages/common/src/ui/components/buttons/standard/ButtonWithIcon.tsx
```diff
@@ -0,0 +1,54 @@
...
+}) => {
+  const t = useTranslation();
+  const isSmallScreen = useIsSmallScreen();
+
+  // On small screens, if the button shouldShrink, then
+  // only display a centered icon, with no text.
+  const shrink = isSmallScreen && shouldShrink;
+  const startIcon = shrink ? null : Icon;
+  const centeredIcon = shrink ? Icon : null;
+  const text = shrink ? null : t(labelKey);
```
joshxg, line 35: I shifted this here because it didn't make sense to have a shrinkable button that wasn't an icon button?

packages/common/src/ui/components/buttons/standard/DialogButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+      color="secondary"
+      disabled={disabled}
+      onClick={onClick}
+      Icon={icon}
+      variant={buttonVariant}
+      labelKey={labelKey}
+      sx={
+        disabled ? { '& svg': { color: theme => theme.palette.midGrey } } : {}
+      }
+    />
```
joshxg, line 61: this is the same except it's using `ButtonWithIcon` which has the styling and things in `BaseButton`

packages/common/src/ui/components/buttons/standard/ShrinkableBaseButton.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export const ShrinkableBaseButton: FC<ShrinkableBaseButtonProps> =
+  React.forwardRef(({ shrink = false, ...props }, ref) => (
+    <StyledShrinkableBaseButton
+      ref={ref}
+      shrink={shrink}
+      size="small"
+      {...props}
+    />
+  ));
```
joshxg, line 29: Just `BaseButton` which shrinks..

packages/common/src/ui/icons/ArrowRight.tsx
```diff
@@ -2,9 +2,8 @@ import React from 'react';
 import SvgIcon, { SvgIconProps } from '@mui/material/SvgIcon';
 
 export const ArrowRightIcon = (props: SvgIconProps): JSX.Element => {
-  const combinedProps: SvgIconProps = { color: 'primary', ...props };
   return (
-    <SvgIcon {...combinedProps} viewBox="0 0 16 16">
+    <SvgIcon {...props} viewBox="0 0 16 16">
```
joshxg, line 6: The buttons will correctly set the colour now.. but they won't if we set a default colour.. so I thought to just pass through props and let them do it themselves now

packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
```diff
@@ -227,6 +235,38 @@ export const OutboundShipmentDetailViewComponent: FC = () => {
...
+              variant="contained"
+              color="secondary"
+              onClick={() => {
+                success('Saved invoice! 🥳 ')();
+                draft.update?.('status', 'finalised');
+                save(draft);
+              }}
+            />
+          </Box>
+        </Box>
```
joshxg, line 266: aaanndd what I was actually meaning to do. A few buttons in the UI and think it might be best to follow up in a new PR with some actual logic.. They are sending requests, just not nicely.

joshxg (commented):
.. Hopefully everything looks and works the same and this helps in some way with extending buttons. Though, I suspect this has been another lesson in useless time wasting. *sigh*

packages/common/src/hooks/useQueryParams/useQueryParams.test.tsx
mark-prins: When doing this sort of thing previously, I've ended up scrolling in some way. There's a lot you can do with determining height and being tricksy, but there are so many edge cases and the whole thing ends up being so fragile, just doesn't seem worth it.

that said, I think stick with what we have for now - there's other stuff to tidy up first :)

packages/common/src/intl/locales/ar.json
```diff
@@ -31,6 +31,8 @@
   "button.ok": "نعم",
   "button.open-the-menu": "افتح القائمة",
   "button.print": "Print",
+  "button.save-and": "Save & #{status}",
```
mark-prins, line 34: The `#` shouldn't be here

packages/common/src/intl/locales/ar.json
mark-prins: Guilty of doing both (1) and (2).. and (3) as well, now that I think of it.

packages/common/src/styles/theme.ts
mark-prins: yeah, I have hovered on doing the same thing. wasn't sure that secondary should be blue throughout, I guess we'll find out as we go on which will be the more useful. I thought that having the grey as a secondary would fit with the mui components.. 🤷‍♂️

packages/common/src/ui/components/buttons/ButtonShowcase.stories.tsx
mark-prins: that looks awesome! great idea

packages/common/src/ui/components/buttons/ButtonShowcase.stories.tsx
```diff
@@ -0,0 +1,153 @@
...
+        icon={<Book />}
+        labelKey="button.docs"
+        onClick={() => console.info('clicked')}
+      />
+    </Wrapper>
+  </Grid>
+);
+
+export const Primary = Template.bind({});
+export const Secondary = Template.bind({});
```
mark-prins, line 149: don't need the `Secondary` now? same as `Primary` isn't it?

packages/common/src/ui/components/buttons/IconButton.stories.tsx
mark-prins: oh oops.

packages/common/src/ui/icons/ArrowRight.tsx
mark-prins: Removing the default colour has made the Svg icon story go strange, they were defaulting to primary, and now only some are.
Could you add a primary prop to the `Primary` of that story and maybe add another for 'Default' or something, so show what happens without providing a colour?

packages/customers/src/Nav.tsx
```diff
@@ -40,7 +40,7 @@ const Nav: FC = () => {
       <NavLink
         end={false}
         to={AppRoute.Customers}
-        icon={<Customers fontSize="small" />}
+        icon={<Customers color="primary" />}
```
mark-prins, line 43: this is what I was trying to avoid, by defaulting to the primary colour. 
do we have to specify a colour now, every time we use an icon? I would expect the icons to default to the primary colour

mark-prins: on reflection.. probably ok to go with the mui default which doesn't specify colour. 
original thinking was that we almost always want the icons to be the primary colour, and so to prevent errors and make them easier to use we should just default to the primary

PR #278, opened 2021-10-18
Add nodemon to mock-server

### LINKED ISSUE #276: Fix demo server mock server not auto refreshing
I replaced babel-node with using ts-node which stopped the auto refreshing? However I believe if you just run babel-node with ts-node as a dep, it works the same?

joshxg:
Fixes #276


mark-prins (approved):
👍  very nifty!

PR #279, opened 2021-10-19
#275 cell formatter

### LINKED ISSUE #275: table/cell value formatter feature
- We're going to be sent a bunch of data from the server like `'draft'` (or maybe `DRAFT` if we're oh so lucky) and maybe `outbound_shipment` (or wait for it `OUTBOUND_SHIPMENT`)
- We will need a nice way to format these values in tables
- At the moment our columns have a `ColumnFormat` which has `Numeric` and `Date` etc
- I think a `formatter` field with an optional function to take a value and produce a string would be nice. This way when we use the status of an invoice, we can pass a function to format it nicely for all status' and then we can also use that same formatter func for other places?

josh-griffin:
Fixes #275 

I had forgotten that we had `accessor` on the column. Technically, could just use this for this purpose. I had been thinking about how we were going to access more deeply nested values (as I had a peak at the API and it looks like invoices will return a nested `pricing: { total }` object....) so was nice to remember this. It was pretty quick to do this change and I think separating the accessing and formatting separately is nice.

Additionally I could have used a new `ColumnFormat.Status` but I wasn't sure as at the moment the formats are just for data types 🤔 

I added that we had to pass the translator fns when calling `formatter()` - I'm not sure. It makes writing your formatter for a cell nice and easy which is nice and generally it's going to only be in the generic places where `formatter()` is generally called so I think it's ok.

### DIFF
--- a/packages/common/src/ui/components/StatusCrumbs.stories.tsx
+++ b/packages/common/src/ui/components/StatusCrumbs.stories.tsx
@@ -1,0 +1,81 @@
+import React, { useState } from 'react';
+import { ComponentStory, ComponentMeta } from '@storybook/react';
+import { LoadingSpinner } from './LoadingSpinner';
+import { StatusCrumbs } from './StatusCrumbs';
+import { OutboundShipmentStatus } from '../..';
+import { LocaleKey, useTranslation } from '../../intl/intlHelpers';
+import Stack from '@mui/material/Stack';
+import Radio from '@mui/material/Radio';
+import RadioGroup from '@mui/material/RadioGroup';
+import FormControlLabel from '@mui/material/FormControlLabel';
+import FormControl from '@mui/material/FormControl';
+import FormLabel from '@mui/material/FormLabel';
+
+const StatusTranslation: Record<OutboundShipmentStatus, LocaleKey> = {
+  draft: 'label.draft',
+  allocated: 'label.allocated',
+  picked: 'label.picked',
+  shipped: 'label.shipped',
+  delivered: 'label.delivered',
+};
+
+export const getStatusTranslation = (
+  currentStatus: OutboundShipmentStatus
+): LocaleKey => {
+  return StatusTranslation[currentStatus] ?? StatusTranslation.draft;
+};
+
+const outboundStatuses: OutboundShipmentStatus[] = [
+  'draft',
+  'allocated',
+  'picked',
+  'shipped',
+  'delivered',
+];
+
+const Template: ComponentStory<typeof StatusCrumbs> = () => {
+  const [currentStatus, setCurrentStatus] = useState(
+    outboundStatuses[0] as OutboundShipmentStatus
+  );
+
+  const t = useTranslation();
+
+  return (
+    <Stack gap={2}>
+      <FormControl>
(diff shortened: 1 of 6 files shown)


packages/common/src/ui/layout/tables/columns/types.ts
```diff
@@ -15,10 +19,10 @@ export interface HeaderProps<T extends DomainObject> {
...
 
 export enum ColumnFormat {
-  date,
-  integer,
-  real,
-  text,
+  Date,
+  Integer,
+  Real,
+  Text,
```
joshxg, line 25: 🙄  I read a thing that the convention is `PascalCase` and I'm a 🐑

packages/common/src/ui/layout/tables/columns/types.ts
```diff
@@ -27,7 +31,19 @@ export enum ColumnAlign {
   Center = 'center',
 }
 
-export type ColumnDataAccessor<T extends DomainObject> = (rowData: T) => string;
+export type ColumnDataAccessor<T extends DomainObject> = (
+  rowData: T
+) => unknown;
```
joshxg, line 36: woah these types for the accessor and formatter are baaad.

Maybe it's essentially for the data type of the underlying row datas .. data. So a number for `invoice.invoiceNumber` or `invoice.pricing.total` etc.. I was thinking to tighten them up we could add another generic to the column so it would be like, `Column<Invoice, number>` for a number column..

packages/common/src/ui/layout/tables/components/index.tsx
```diff
@@ -1,18 +1,18 @@
...
-const capitalize = (str: string) =>
-  str.slice(0, 1).toUpperCase() + str.slice(1, str.length).toLowerCase();
-
 export const BasicCell = <T extends DomainObject>({
   column,
   rowData,
 }: CellProps<T>): ReactElement => {
-  return <>{capitalize(String(column.accessor(rowData)))}</>;
+  const t = useTranslation();
+  const d = useFormatDate();
```
joshxg, line 13: ... 🤔

packages/invoices/src/OutboundShipment/ListView/ListView.tsx
```diff
@@ -108,7 +110,13 @@ export const OutboundShipmentListViewComponent: FC = () => {
...
     [
       getNameAndColorColumn(onColorUpdate),
-      'status',
+      [
+        'status',
+        {
+          formatter: (status, { t }) =>
+            t(getStatusTranslation(status as OutboundShipmentStatus)),
+        },
+      ],
```
joshxg, line 119: not quite as nice as just passing a string.. 😁  but I think it still ends up being quite nice..

packages/common/src/ui/layout/tables/components/index.tsx
mark-prins: yes.. I was wondering too. but no good ideas

packages/invoices/src/OutboundShipment/ListView/ListView.tsx
mark-prins: maybe not as simple as just a string.. still very neat tho

PR #280, opened 2021-10-19
#272 status bar to detail view

### LINKED ISSUE #272: Add status bar to detail view
- Add a little bar of statuses to the bottom fo the detail view

UI:
image

Workflow:
- Nothing here https://docs.google.com/spreadsheets/d/1H0Rv1DEOgW0cE07wpfc_KivLf8TcNhYDqlwRp92x-To/edit#gid=890672777
- Assuming the status only changes when you click Save & status and there is a success response?
- Unsure if you can 'skip' steps or if the statuses are clickable

Implementation:
- Seems using the mui breadcrumbs would be a quick and easy way
- https://codesandbox.io/s/customseparator-material-demo-forked-2xp8j?file=/demo.js

josh-griffin:
Fixes #272 

- Adds a `StatusCrumbs` component (Uhh.. it's all I could think of)
- Still need to do the pop up part

### DIFF
--- a/packages/common/src/ui/components/StatusCrumbs.stories.tsx
+++ b/packages/common/src/ui/components/StatusCrumbs.stories.tsx
@@ -33,7 +33,7 @@ const outboundStatuses: OutboundShipmentStatus[] = [
   'delivered',
 ];
 
-const Template: ComponentStory<typeof LoadingSpinner> = () => {
+const Template: ComponentStory<typeof StatusCrumbs> = () => {
   const [currentStatus, setCurrentStatus] = useState(
     outboundStatuses[0] as OutboundShipmentStatus
   );


packages/host/src/Footer.tsx
```diff
@@ -20,11 +20,12 @@ export const Footer: React.FC = () => {
 
   return (
     <Box
+      gap={2}
       sx={{
         backgroundColor: 'background.toolbar',
         height: '32px',
-        paddingLeft: '10px',
       }}
+      marginLeft={3}
```
joshxg, line 28: I changed this to line up with the stuff on the left margin of this page

packages/common/src/ui/components/StatusCrumbs.tsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
```
joshxg, line 1: there is a better place to put this but I couldn't find it 😭

mark-prins (approved):
👍  that's looking fantastic - fast work!

It's another one where the 'bold' in zeplin looks better than real 'bold' but I've had that in other zeplin projects too.. they must cheat somehow.

Nice work with the two-part story too!

Issue #260, opened 2021-10-15
Add Item: Implement FEFO logic


mark-prins:
There needs to be some allocation of quantities to stock lines in order to fulfil the requested quantity

PR #281, opened 2021-10-19
#260 fefo logic

### LINKED ISSUE #259: Add item: refactor of modal
Would like to refactor the `ItemDetail` and `ItemBatch` along the lines in [this comment](https://github.com/openmsupply/openmsupply-client/pull/230#discussion_r725396362)

Something like this:

- `ItemDetails` : Has the `useDialog`/`Modal`, `useForm` and any data things, i.e. `useQuery` and is passed an `onSubmit`/`open` prop
    - `Modal` : Handles `Modal` things
        - `InvoiceLine` : basically is the `ItemDetails` component now with 
        - `StockLines`: basically the current `ItemBatches` component

Wondering about using `InvoiceLine` and `StockLines` for the component names, which makes it more consistent with the data structure and less consistent with the site which talks about `Item` (Add/Edit) 🤷‍♂️

mark-prins:
Fixes #259 , #260 

Got a little carried away and did both at the same time.
Mostly working - will need to look at placeholders and pack sizes though.
The placholders work, though I haven't added a date or correct identifier and so won't display correctly in the table yet. Will need to research and discuss how to handle placeholder lines.

The quantity is working only because pack size is 1 - will need to distribute taking the pack size into account, thought to do that in another PR though. enough in this one already.

In the designs there was also some additional messaging / warning which needs to be looked into as well.


packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
```diff
@@ -0,0 +1,239 @@
...
+  const [selectedItem, setSelectedItem] = React.useState<Item | null>(null);
+  const [quantity, setQuantity] = React.useState(0);
+  const [isAllocated, setIsAllocated] = React.useState(false);
+  const [lines, setLines] = React.useState<InvoiceLine[]>(
+    invoiceLine ? [invoiceLine] : []
+  );
+  const { hideDialog, showDialog, Modal } = useDialog({
+    title: 'heading.add-item',
+  });
+  const methods = useForm({ mode: 'onBlur' });
```
mark-prins, line 105: `onBlur` to trigger the form validation. However in the end I've removed the form validation - it wasn't reporting `isDirty` when it was dirty and the `isValid` was always true - and then it wouldn't allow the form to submit 🙄 
Obvs I'm doing something wrong, but didn't see what it was.
Ignoring the form and using explicit event methods and local state has been more reliable

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
```diff
@@ -0,0 +1,239 @@
...
+    setValue('code', value?.code || '');
+    trigger('itemId');
+  };
+
+  const { data, isLoading } = useQuery(['item', 'list'], listQueryFn);
+  const onReset = () => {
+    reset();
+    setBatchRows([]);
+    setQuantity(0);
+    setValue('quantity', 0);
```
mark-prins, line 129: the problem with using local state - the `reset` is handling the form values ok

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
```diff
@@ -0,0 +1,239 @@
...
+  isOpen,
+  onClose,
+  upsertInvoiceLine,
+}) => {
+  const [batchRows, setBatchRows] = React.useState<BatchRow[]>([]);
+  const [selectedItem, setSelectedItem] = React.useState<Item | null>(null);
+  const [quantity, setQuantity] = React.useState(0);
+  const [isAllocated, setIsAllocated] = React.useState(false);
+  const [lines, setLines] = React.useState<InvoiceLine[]>(
+    invoiceLine ? [invoiceLine] : []
```
joshxg, line 100: is `lines` a subset of batches mapped slightly differently? Maybe we can get rid of `lines` in favour of filtering/pruning unused `batches` and mapping them when saving when we revisit the state later

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
```diff
@@ -0,0 +1,239 @@
...
+  return 0;
+};
+
+export const ItemDetailsModal: React.FC<ItemDetailsModalProps> = ({
+  invoiceLine,
+  isOpen,
+  onClose,
+  upsertInvoiceLine,
+}) => {
+  const [batchRows, setBatchRows] = React.useState<BatchRow[]>([]);
```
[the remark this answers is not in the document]
joshxg: I suspect nothing a `useReducer` can't fix 🤣

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
mark-prins: genius!
silly me, have implemented and that's much better - thanks!

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
```diff
@@ -0,0 +1,239 @@
...
+  const [selectedItem, setSelectedItem] = React.useState<Item | null>(null);
+  const [quantity, setQuantity] = React.useState(0);
+  const [isAllocated, setIsAllocated] = React.useState(false);
+  const [lines, setLines] = React.useState<InvoiceLine[]>(
+    invoiceLine ? [invoiceLine] : []
+  );
+  const { hideDialog, showDialog, Modal } = useDialog({
+    title: 'heading.add-item',
```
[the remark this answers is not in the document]
mark-prins: thanks

PR #282, opened 2021-10-19
save icon

### LINKED ISSUE #270: Add save icon
Need to add the save icon for the save button 

image

haven't looked where to get it yet

mark-prins:
Fixes #270 

Adds the save icon. only a small offroad, to fix a merge issue I caused elsewhere.
Also popped in a `Default Values` test case to show that the icons look like without props.. because 🤷‍♂️  
the only real reason is the mSupply guy because that one isn't behaving in a standard way. not sure that it is worth it, though the cost of adding it in is really minimal. storybook is great


joshxg (approved):
Ah thanks..!

PR #286, opened 2021-10-19
#269 hold toggle button

### LINKED ISSUE #269: Add hold toggle button
- Add this toggle button for hold
- Assuming not to add the authorise button

UI:
image

Workflow:
- There's no workflow here
https://docs.google.com/spreadsheets/d/1H0Rv1DEOgW0cE07wpfc_KivLf8TcNhYDqlwRp92x-To/edit#gid=890672777
- Assuming that it is the same as mSupply where pushing the hold toggle prevents the status from changing
- Unsure if permissions etc are needed, but that can be later

Implementation:
- Create a generic `ToggleButton` with styles using: https://mui.com/components/toggle-button/#main-content
- Checked it out quick in a sandbox, pretty easy to do https://codesandbox.io/s/standalonetogglebutton-material-demo-forked-17o3x?file=/demo.js
- don't think we'd need the `ToggleButtonGroup` for now

joshxg:
Fixes #269 

A little off script to the designs if you have ideas on how to make it better.. I changed the footer colour as it was a little weird to have that the same colour as the body but separated by the different colour of the button bar.

### DIFF
--- a/packages/mock-server/src/schema/Invoice.ts
+++ b/packages/mock-server/src/schema/Invoice.ts
@@ -17,6 +17,7 @@ type Invoice {
     total: String
     name: Name
     otherPartyName: String
+    hold: Bool
     lines: [InvoiceLine]
 }
 
--- a/packages/common/src/intl/index.ts
+++ b/packages/common/src/intl/index.ts
@@ -1,19 +1,4 @@
-import { useIntl } from 'react-intl';
-import {
-  useFormatDate,
-  useTranslation,
-  useTranslationWithFallback,
-  useRtl,
-} from './intlHelpers';
-import { IntlProvider } from './IntlProvider';
-import { IntlTestProvider } from './IntlTestProvider';
-
-export {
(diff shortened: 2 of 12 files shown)


packages/mock-server/src/schema/Invoice.ts
```diff
@@ -17,6 +17,7 @@ type Invoice {
     total: String
     name: Name
     otherPartyName: String
+    hold: Bool
```
mark-prins, line 20: does this work for you? 😕  I have to change to `Boolean`

joshxg:
For multiple buttons, the `ToggleButtonGroup` handles it and makes it look just like the pics (well, if you're using my eyes at least!): https://codesandbox.io/s/standalonetogglebutton-material-demo-forked-17o3x?file=/demo.js

I've done (I think) all the styling suggestions except for adding more pixel space to the bottom of the user/store which I'll do when I do a follow up to make this all a bit neater

PR #296, opened 2021-10-20
#292 clean up footer

### LINKED ISSUE #292: Clean up footer
Now that there's more stuff in the footer, would be good to clean it up and get it out of the detail view.

Thinking a `FooterPortal` which has the user/store text by default.

As well as, most likely, an `OutboundDetailFooter` specifically

joshxg:
Fixes #292 

Unfortunately had to use two refs, but turned out OK


mark-prins (approved):
👍  Looking good!

PR #304, opened 2021-10-20
#274 Disable things in an invoice when shipped+

### LINKED ISSUE #274: Add disabling of all of the things when the status of an outbound is..
When the status is shipped (I think?), we should disable editing of all of the things

joshxg:
Fixes #274 

- Renamed `OutboundShipmentDetailViewToolbar` -> `OutboundDetailToolbar`?
- Renamed `isInvoiceSaveable` -> `isInvoiceEditable`?
- Disabled the customer/customer ref inputs and the add item button.


joshxg:
Sorry I totally forgot about my open PRs.. I agree and will make sure to get the button change in a bit

PR #309, opened 2021-10-21
#308 update invoice list view query

### LINKED ISSUE #308: Update queries to use the rust api
- [ ] ListView query
- [ ] Delete invoice
- [ ] Specific invoice : Note only can filter by id currently, not invoice number. Think we should just use the id and have ugly urls until it's supported
- [ ] update invoice [big one]

joshxg:
Fixes #308 

- I wanted some data in my sqlite db and I didn't see too many options, so I thought to just insert directly into an sqlite db.
- I thought this was a good way to tell what was different from what we currently are doing and with the backend api is to insert our mock data into the sqlite db

This has ended up being pretty successful, I think. Though the PR is bigger than what I had hoped.

- I've added the code I used to insert the mock data into the sqlite. Not sure if this is helpful
- I've not updated the mock server properly. We're in a bit of a shambles. I thought it would be best to PR this and not merge until I have a couple other pieces in place

I am not sure I see that the mock-server is needed but I do find having to run the backend in a seperate terminal, mSupply if I need to sync and then run our app in a second terminal is extremely cumbersome. Personally I think keeping some mocks working is beneficial. Maybe a good middle ground is to just have the mock-worker return data from a query, rather than trying to keep the apollo server up to date.

There are only a few small pieces missing and a couple pieces I'm not sure about - everything is looking great.

- Add: `[invoice]hold`.
- Add: `invoiceNumber` filter to `invoice` query.
- Nit: I have found myself mistyping `XDatetime` with `xDateTime` many times. I don't know what's right but it's an annoyance for me 🤣 
- The invoice pricing. I think it's probably fine and nice if there will be multiple fields in there, but being flat seems nice. Though I think there might be restrictions on the backend for that 🤷 
-


packages/common/src/hooks/useListData/useListData.ts
```diff
@@ -16,15 +16,15 @@ export interface ListApi<T extends ObjectWithStringKeys> {
     first: number;
     offset: number;
     sortBy: SortBy<T>;
-  }) => () => Promise<{ data: T[]; totalLength: number }>;
+  }) => () => Promise<{ nodes: T[]; totalCount: number }>;
```
joshxg, line 19: The api does `nodes/totalCount` rather than what we had `data/totalLength` - oops

packages/common/src/types/index.ts
```diff
@@ -73,14 +73,17 @@ export interface Invoice extends DomainObject {
...
+  confirmedDatetime: string;
   invoiceNumber: string;
   total: string;
   name?: Name;
   otherPartyName: string;
   hold: boolean;
   lines: InvoiceLine[];
+  pricing: {
+    totalAfterTax: number;
+  };
```
joshxg, line 86: these were the main differences - `entered -> entryDatetime`, `confirmed` -> `confirmedDatetime` and the pricing instead of `total`

packages/invoices/src/api.ts
```diff
@@ -148,15 +181,23 @@ export const listQueryFn = async <T extends ObjectWithStringKeys>(queryParams: {
...
-  });
+}): Promise<{ nodes: Invoice[]; totalCount: number }> => {
+  const {
+    first = 20,
+    offset = 0,
+    sortBy = { key: 'TYPE', isDesc: false },
+  } = queryParams;
+
+  const { invoices } = await request(
+    'http://localhost:8000/graphql',
```
joshxg, line 192: I haven't updated the mock server so this query is targeting this url which is the default for the rust server, for now. Lots of cross over between things ..

packages/invoices/src/api.ts
```diff
@@ -148,15 +181,23 @@ export const listQueryFn = async <T extends ObjectWithStringKeys>(queryParams: {
...
+    getListQuery(),
+    {
+      first,
+      offset,
+      key: 'TYPE',
```
joshxg, line 197: obsessed with SHOUTING AT EVERYONE!

packages/mock-server/package.json
```diff
@@ -23,6 +25,9 @@
     "apollo-server": "^3.1.1",
     "faker": "^5.5.3",
     "graphql": "^15.5.1",
-    "nodemon": "^2.0.13"
+    "knex": "^0.95.11",
+    "lodash.chunk": "^4.2.0",
+    "nodemon": "^2.0.13",
+    "sqlite3": "^5.0.2"
```
joshxg, line 31: I left all this in in case it was helpful at all in the short term

packages/mock-server/src/data/data.ts
```diff
@@ -25,62 +25,67 @@ export const getStockLinesForItem = (
   return stockLines.filter(getFilter(item.id, 'itemId'));
 };
 
-const createStockLines = (items: Item[]) => {
+export const createStockLines = (
```
joshxg, line 28: Looks like a lot of changes, but the main changes in here are just to get the export into sqlite working:

- added stores
- a few date changes - expiry dates require a different date format
- invoice type needs to YELL
- pricing change

packages/mock-server/src/populate.ts
```diff
@@ -0,0 +1,187 @@
...
+import knex from 'knex';
+import chunk from 'lodash.chunk';
+import {
+  InvoiceLineData,
+  ItemData,
+  NameData,
+  StoreData,
+  StockLineData,
+  InvoiceData,
+} from './data/data';
```
joshxg, line 11: just a bunch of mapping to the sqlite schema..

PR #344, opened 2021-10-30
2 - Add refactor popover out to own folder

joshxg:
This factors out the popover (is that a word?) into a more generic implementation to be used elsewhere.

Also made a hook to use it copying your hook for the dialog which I thought was another nice use


packages/common/src/ui/components/popover/BasePopover/BasePopover.tsx
```diff
@@ -0,0 +1,45 @@
...
+import Popper, { PopperProps } from '@mui/material/Popper';
+
+export interface BasePopoverProps extends Omit<PopperProps, 'open'> {
+  // I like having is as a prefix for boolean props
+  isOpen: boolean;
+  anchorEl: PopperProps['anchorEl'];
+  paperProps?: PaperProps;
+}
+
+export const BasePopover: FC<BasePopoverProps> = ({
```
joshxg, line 13: I don't know if this is a good name - I can see there being more specific variants - `DetailPopover` and could (should?) be for the `ColorMenu`.

packages/common/src/ui/components/popover/BasePopover/usePopover.tsx
```diff
@@ -0,0 +1,90 @@
...
+  React.SetStateAction<boolean>
+> | null>;
+
+interface UsePopoverControl {
+  show: React.MouseEventHandler<HTMLDivElement | HTMLButtonElement>;
+  hide: () => void;
+  Popover: FC;
+}
+
+export const usePopover = (): UsePopoverControl => {
```
joshxg, line 22: I think this ends up being nice to use, though the implementation I can see does become a little confusing

packages/common/src/ui/components/popover/BasePopover/usePopover.tsx
```diff
@@ -0,0 +1,90 @@
...
+    setAnchorElCallback.current = setInternalAnchorEl;
+
+    return (
+      <BasePopover
+        {...props}
+        anchorEl={internalAnchorEl}
+        isOpen={internalIsOpen}
+      />
+    );
+  }, []);
```
joshxg, line 83: Basically this is all to avoid putting in the `isOpen`/`anchorEl` here - because that will mean the component is essentially uncached and will guarantee a re-mount every time, rather than re-rendering, which we get with this

packages/common/src/ui/components/popover/BasePopover/BasePopover.tsx
```diff
@@ -0,0 +1,45 @@
...
+      anchorEl={anchorEl}
+      transition
+      placement="bottom-start"
+    >
+      {({ TransitionProps }) => (
+        <Fade {...TransitionProps} timeout={350}>
+          <Paper
+            {...paperProps}
+            sx={{
+              width: 240,
```
mark-prins, line 33: I like having the dimensions as props, since they are forced here: currently to override you have to set the `sx` and it isn't clear that this is required. having explicit dimension props would help, I think

packages/common/src/ui/components/popover/BasePopover/BasePopover.tsx
```diff
@@ -0,0 +1,45 @@
...
+              ...paperProps?.sx,
+            }}
+          >
+            {children}
+          </Paper>
+        </Fade>
+      )}
+    </Popper>
+  );
+};
```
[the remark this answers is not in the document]
mark-prins: probably valid - I can see the order history and the colour thingy being reused elsewhere

packages/common/src/ui/components/popover/BasePopover/usePopover.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+  // TODO: Maybe this should be a generic element. If you come across this and find
+  // yourself wanting to have the anchor of a popover be some other element,
+  // there was no real reason I made this not generic other than that was all I needed
+  // at the time.
+
+  const showCallback: React.MouseEventHandler<
+    HTMLDivElement | HTMLButtonElement
+  > = e => {
+    // TODO: This virtual rect is use to create a virtual element to anchor the popover on. These
```
mark-prins, line 41: :nit.tiny
```suggestion
    // TODO: This virtual rect is used to create a virtual element to anchor the popover on. These
```

packages/common/src/ui/components/popover/BasePopover/usePopover.tsx
mark-prins: uh huh. usage is very nice though.. so implementation doesn't really affect anyone

packages/common/src/ui/components/popover/BasePopover/usePopover.tsx
```diff
@@ -0,0 +1,90 @@
...
+  show: React.MouseEventHandler<HTMLDivElement | HTMLButtonElement>;
+  hide: () => void;
+  Popover: FC;
+}
+
+export const usePopover = (): UsePopoverControl => {
+  // The Popover component itself carries the state and
+  // assigns callbacks to these refs which can control
+  // the state. This is done so that we can control the
+  // state, but also ensure the Popover component doesn't
```
mark-prins, line 26: :nit.tiny
is this right?
```suggestion
  // state, but also ensures the Popover component doesn't
```

packages/common/src/ui/components/popover/BasePopover/usePopover.tsx
mark-prins: wow - nice!

mark-prins (approved):
👍  That's fantastic! the hook implementation is great too - nice and easy to use!

packages/common/src/ui/components/popover/BasePopover/BasePopover.tsx
joshxg: nice! 
7ecb0b7

PR #345, opened 2021-10-30
3 - #status crumb cleanup

joshxg:
Final one, just doing a bit of tidying while i'm here..


packages/common/src/ui/components/domain/StatusCrumbs/StatusCrumbs.tsx
```diff
@@ -2,97 +2,72 @@ import React from 'react';
 import Breadcrumbs from '@mui/material/Breadcrumbs';
 import Typography from '@mui/material/Typography';
 import Box from '@mui/material/Box';
-
-import { useAppTheme } from '../../styles';
-import { ChevronDownIcon } from '../icons';
+import { ChevronDownIcon } from '../../../icons';
```
joshxg, line 5: Moved everything into a folder in `/components` because it seems neat. Then I had trouble naming things, of course.

packages/common/src/ui/components/domain/StatusCrumbs/StatusCrumbs.tsx
```diff
@@ -2,97 +2,72 @@ import React from 'react';
...
 import {
   LocaleKey,
   useTranslation,
   useFormatDate,
-} from '../../intl/intlHelpers';
-import { VerticalStepper } from './steppers/VerticalStepper';
-import { usePopover } from './popover';
+} from '../../../../intl/intlHelpers';
```
joshxg, line 10: just gone through and removed all the things that ended up being unneeded and added touchups like extracting the loop out to a hook

mark-prins (approved):
👍  very nice!

PR #346, opened 2021-10-30
4 - #316 more button

### LINKED ISSUE #316: Detail Panel: More button
There are complaints about the more button - people feel it should be a hamburger menu icon with a pop up menu and that maybe one of the items in the list is 'show detail panel'. In which case the button wouldn't disappear when the panel is opened.

Alternatively just change the icon; the suggestion is 
- (closed)[https://thenounproject.com/search/?q=right+panel&i=3894774]
- (opened)[https://thenounproject.com/search/?q=right+panel&i=3894747]

though with currently functionality there is no need for an opened version, so I'm thinking not to use that for now.
These icons are noun project and I think it would be better to retain consistency and use the feather version:

Screen Shot 2021-10-22 at 3 56 43 PM

joshxg:
Fixes #316

### DIFF
--- a/packages/common/src/hooks/useDetailPanel.tsx
+++ b/packages/common/src/hooks/useDetailPanel.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import create from 'zustand';
-import { MenuDotsIcon, ButtonWithIcon } from '../ui';
+import { SidebarIcon, ButtonWithIcon } from '../ui';
 
 type DetailPanelController = {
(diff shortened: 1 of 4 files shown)


joshxg:
image

🤷

mark-prins (approved):
👍  Looks good!

I wondered if the button should also be renamed, something more specific like 'Details' maybe.. but that didn't come up, so I think we see what the feedback is on this :)

PR #347, opened 2021-10-30
#291 add item extras

### LINKED ISSUE #291: OS Detail: stock line allocation
Some additional parts to complete:

- Some messaging:
  - the text `## allocated` beside the quantity input
  - make the expiry date red if..?
  - make the placeholder row text red
  - add available quantity to the far right of the autocomplete row? is that possible?
  - show `## allocated` in red if different from the quantity value
- Take account of pack size when allocating; will need to round up to nearest pack size
- Do something with the info icon 🤷‍♂️

mark-prins:
Fixes #291 

I got a little carried away with this one, sorry. Tried to fix everything and should have stopped earlier.

Did try to get the autocompleter to auto open, and would still really like to. The issue I had is that the control is then controlled which prevents the keyboard nav with it, and the drop down arrow. Solvable I think; it was a little weird and needed a little more time, so I stopped looking at it. I've done enough damage for now.. I think.

Changes:
- [ ] the text ## allocated beside the quantity input
- [ ] make the expiry date red if less than 3 months remaining
- [ ] the available quantity is already in the autocomplete row, have added a new field to display it once an item is selected
- [ ] show ## allocated in red if different from the quantity value
- [ ] The hold checkbox is read only
- [ ] Stock lines cannot be used if on hold
- [ ] Sort the stock lines so that available are on top
- [ ] Move the Pack Size column to be second
- [ ] Rename issue to “# packs”
- [ ] Add total quantity
- [ ] Show location column
- [ ] Hide cost price column
- [ ] The in-store and total-available are the wrong way around
- [ ] The info icon popup could show what is allocated on other invoices if the API provided this detail
- [ ] Quantity input: add units to the right.. ( eg ’bottles’ ) after the input
- [ ] Drop down to the right of the quantity input showing distinct values of pack sizes available. We didn’t fully discuss what this would do - my understanding is that it would restrict which batch rows are used for allocation. Add a value of 1 to the drop down so that you can request a specific number of units rather than a number of packs


packages/common/src/ui/components/Popper.tsx
```diff
@@ -0,0 +1,72 @@
+import React, { useState } from 'react';
```
mark-prins, line 1: A quick addition of the Popper you've done for the crumbs.. just popped it in here for now 🤷‍♂️

packages/common/src/ui/components/inputs/TextInput/ReadOnlyInput.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+
+export interface ReadOnlyInputProps {
+  number?: boolean;
+  style?: React.CSSProperties;
+  width?: string;
+}
+
+export const ReadOnlyInput: React.FC<ReadOnlyInputProps> = React.forwardRef<
```
mark-prins, line 9: possibly a daft idea. I wanted to add text which is controlled by the form, and which didn't look like an input.

packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
```diff
@@ -61,7 +61,7 @@ export const DetailView: FC = () => {
   return draft ? (
     <TableProvider createStore={createTableStore}>
       <AppBarButtons
-        isDisabled={isInvoiceEditable(draft)}
+        isDisabled={!isInvoiceEditable(draft)}
```
mark-prins, line 64: it was proving to be a pain when testing the add items!

packages/invoices/src/OutboundShipment/DetailView/modals/BatchesTable.tsx
```diff
@@ -16,62 +19,117 @@ import {
...
+        <Checkbox disabled checked={batch.onHold} />
+      </BasicCell>
       <BasicCell>
-        <InfoIcon
-          fontSize="small"
-          sx={{ color: theme => theme.palette.lightGrey }}
-        />
+        <Popper
+          content={
+            <Grid spacing={2} container sx={{ padding: '20px' }}>
```
mark-prins, line 95: this is all a bit too far probably - was just wanting to demo something to start the discussion about what to show here. if anything, maybe we don't want the info icon after all?

Thought that we could have the API extended slightly to return invoice names and quantities of 'other' invoices, in order to detail where the other stock is allocated 🤷

packages/invoices/src/OutboundShipment/DetailView/modals/BatchesTable.tsx
```diff
@@ -154,10 +213,7 @@ export const BatchesTable: React.FC<BatchesTableProps> = ({
                 {t('label.placeholder')}
               </BasicCell>
               <BasicCell sx={{ paddingTop: '3px' }}>
-                <NumericTextInput
-                  {...placeholderInputProps}
-                  onChange={onChangeValue}
-                />
+                <ModalNumericInput inputProps={placeholderInputProps} />
```
mark-prins, line 216: these can be `NumericInput` again, now that I'm not showing the error message..

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
```diff
@@ -83,7 +83,16 @@ const listQueryFn = async (): Promise<Item[]> => {
   return items.nodes;
 };
 
-const sortByExpiryDate = (a: BatchRow, b: BatchRow) => {
+const sortByDisabledThenExpiryDate = (a: BatchRow, b: BatchRow) => {
```
mark-prins, line 86: it was another suggestion - put the disabled rows last

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
```diff
@@ -169,44 +181,64 @@ export const ItemDetailsModal: React.FC<ItemDetailsModalProps> = ({
...
+    let batchAllocation = 0;
 
     batchRows.forEach(batch => {
-      const allocatedQuantity = Math.min(
-        toAllocate,
-        batch.availableNumberOfPacks * batch.packSize
-      );
-      toAllocate -= allocatedQuantity;
-      setValue(batch.id, allocatedQuantity);
+      batchAllocation = 0;
```
mark-prins, line 209: the addition of pack size selection has made this bigger than I wanted. I _think_ it is all worth it?
we do have a lot of trouble with quantity / pack size confusion

packages/common/src/utils/dateFunctions.ts
```diff
@@ -0,0 +1,6 @@
+import { differenceInMonths } from 'date-fns';
+
+export const MINIMUM_EXPIRY_MONTHS = 3;
+
+export const isAlmostExpired = (expiryDate: Date): boolean =>
+  differenceInMonths(expiryDate, Date.now()) <= MINIMUM_EXPIRY_MONTHS;
```
joshxg, line 6: this is cool!

packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
joshxg: oops!

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
joshxg: Ah right, I was a little confused by the order before seeing this.

I think something like a different colour for hold and 0 quantity (so, 3 different colours) or indicator.. maybe the on hold rows have very little opacity and 0quant rows have the quantity column with a background or something.. and then an order like: 0>Quantity, <=0 Quantity, On Hold

hehe

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
joshxg: Hopefully 😁  it is becoming quite a beast!

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
```diff
@@ -169,44 +181,64 @@ export const ItemDetailsModal: React.FC<ItemDetailsModalProps> = ({
...
+      // skip bigger pack sizes
```
joshxg, line 210: I would have thought that if we were selecting pack sizes we would only assign to that specific pack size, rather than anything smaller? Maybe I don't understand the use case 🤷

packages/invoices/src/OutboundShipment/DetailView/modals/BatchesTable.tsx
joshxg: All good, nice to have something concrete in place to get feedback

joshxg (approved):
sooo badass! loving it all coming together

packages/common/src/ui/components/buttons/standard/ButtonWithIcon.tsx
```diff
@@ -42,19 +42,21 @@ export const ButtonWithIcon: React.FC<ButtonWithIconProps> = ({
...
-        variant={variant}
-        color={color}
-        size="small"
-        startIcon={startIcon}
-        {...buttonProps}
-      >
-        {centeredIcon}
-        {text}
-      </ShrinkableBaseButton>
+      <span>
```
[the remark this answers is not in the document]
mark-prins: the tooltip throws a warning if it is around a disabled button because events aren't raised on disabled elements. wrapping the button in a span, which isn't disabled, removes a warning 🤷

packages/common/src/ui/components/inputs/Select/Select.stories.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { Box } from '@mui/material';
+
+export default {
+  title: 'Inputs/Select',
+  component: Select,
+} as ComponentMeta<typeof Select>;
+
+const Template: ComponentStory<typeof Select> = args => (
+  <Box>
+    <Select {...args} />
```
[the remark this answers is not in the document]
mark-prins: yep. it *really* annoyed me but I didn't see what it was and in the end gave up

packages/common/src/ui/components/inputs/TextInput/ReadOnlyInput.tsx
mark-prins: > Right now looks just like a disabled input

Only used in the batch table currently - to show the unit quantity here:

Screen Shot 2021-11-01 at 12 33 52 PM

> I wondered about a more informational approach

yeah - that's what I was thinking. it troubled me having the disabled inputs. I've kept them so far because they show the distinction between label and value pretty well and also make the form a little less boring.

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
mark-prins: good idea - and good feedback that the strange ordering is not clear!

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
mark-prins: I think we probably should stick to the exact pack size and avoid the weirdness that I've introduced here. maybe also remove the pack size of 1 that is added to the list then, and only show available pack sizes? and then change placeholder input to be # packs not # units.. and then record the pack size when creating placeholder lines..

Issue #311, opened 2021-10-22
List View: group by item


mark-prins:
Currently all invoice lines are shown. Add the ability to have items grouped by item:

- add a button ( somewhere? perhaps the plus icon? ) to expand all
- and the reverse.. a collapse all when expanded ( the minus icons? )
- If only one batch then batch details appear in the row
- If more than 1 batch, then “[# batches]” appear and you can expand the row - some indication within the row
- Add a keyboard shortcut. Include in KBar

There are no designs, so for now please just wing it and then add to the list of [stuff we made up](https://docs.google.com/spreadsheets/d/1H0Rv1DEOgW0cE07wpfc_KivLf8TcNhYDqlwRp92x-To/edit?pli=1#gid=1543225647)

joshxg:
Unsure what to do here so will wait for more details or a mock up

mark-prins:
Argh! sorry - I must have missed your earlier comment!
Yes - I thought to do a mock up today of what I understand is the request and see if you think it might work. Agree with your concern about column alignment, so I'm not sure how this is going to work.

And yes.. oops. Group by item. I keep saying 'batch' 🙄

mark-prins:
Ok - here is something:

### Normal view
when the 'group by item' toggle is off

<img width="1366" alt="Screen Shot 2021-11-02 at 4 55 48 PM" src="https://user-images.githubusercontent.com/9192912/139791466-d1c939c3-6aba-4ae9-a8ac-37648c4f449c.png">

### Grouped view
<img width="1367" alt="Screen Shot 2021-11-02 at 6 24 32 PM" src="https://user-images.githubusercontent.com/9192912/139791550-9dff5f8b-6cf4-418c-a515-38e283cebf71.png">

and then expanded when you click on the chevron:
<img width="1367" alt="Screen Shot 2021-11-02 at 6 24 46 PM" src="https://user-images.githubusercontent.com/9192912/139791580-5561d8db-b0fc-4d5f-8860-b5854c887284.png">

which looks something like this..
https://user-images.githubusercontent.com/9192912/139791774-00390842-870e-4c9e-93d4-bcbd6bb88c7a.mp4

DhanyaHerath:
+1 to the Group by item button, though will that just do the same thing as the column chevron?

 The expansion chevron itself is nice - though not sure it's clear that it's a separate section,  on my first look I thought the columns were just misaligned, but maybe on the browser the color change is more clear!

[...]

PR #348, opened 2021-10-30
5 - #311 group by batch

joshxg:
I messed up commits - whoops.

First one is https://github.com/openmsupply/openmsupply-client/pull/349 - Accidentally committed to that branch *and* pushed, so took an earlier commit and made that PR.

----------

This one introduces row expanding

image

- It's in the `TableContext` which made this quite easy
- I used chevron because the plus looked strange, then I googled and it said use chevron for accordions - the double chevrons.. uh.. maybe not..
- The content is just placeholder
- I am fairly convinced that using an accordion for rows AND a modal is too much - I don't see why we need both (and now have a preference for the accordion...)

### DIFF
--- a/packages/common/src/ui/layout/tables/columns/RowExpand/RowExpand.tsx
+++ b/packages/common/src/ui/layout/tables/columns/RowExpand/RowExpand.tsx
@@ -1,0 +1,102 @@
+import { Box } from '@mui/material';
+import React, { useCallback } from 'react';
+import { ColumnAlign, ColumnDefinition } from '..';
+import { TableStore, useTableStore } from '../..';
+import { IconButton } from '../../../..';
+
+import {
+  ChevronDownIcon,
+  ChevronsDownIcon,
+  DomainObject,
+} from '../../../../..';
+
+const useExpanded = (rowId: string) => {
+  const selector = useCallback(
+    (state: TableStore) => {
+      return {
+        rowId,
+        isExpanded: state.rowState[rowId]?.isExpanded,
+        toggleExpanded: () => state.toggleExpanded(rowId),
+        expanded: state.numberExpanded > 0,
+      };
+    },
+    [rowId]
+  );
+
+  const equalityFn = (
+    oldState: ReturnType<typeof selector>,
+    newState: ReturnType<typeof selector>
+  ) =>
+    oldState?.isExpanded === newState?.isExpanded &&
+    oldState.rowId === newState.rowId;
+
+  const { isExpanded, toggleExpanded, expanded } = useTableStore(
+    selector,
+    equalityFn
+  );
+
+  return { isExpanded, toggleExpanded, expanded };
+};
+
+export const getRowExpandColumn = <
+  T extends DomainObject
+>(): ColumnDefinition<T> => ({
+  key: 'expand',
+  sortable: false,
+  align: ColumnAlign.Center,
+  width: 60,
+  Header: () => {
+    const { numberExpanded, toggleAllExpanded } = useTableStore();
+    return (
+      <IconButton
+        labelKey="app.admin"
+        onClick={toggleAllExpanded}
+        icon={
+          <Box
+            sx={{
+              transition: theme =>
+                theme.transitions.create('transform', {
+                  duration: theme.transitions.duration.leavingScreen,
+                }),
+              transform: !!numberExpanded ? 'rotate(180deg)' : 'rotate(360deg)',
+            }}
+          >
+            <ChevronsDownIcon />
+          </Box>
+        }
+      />
+    );
+  },
+  Cell: ({ rowData }) => {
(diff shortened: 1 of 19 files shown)


packages/common/src/ui/layout/tables/columns/RowExpand/RowExpand.tsx
```diff
@@ -0,0 +1,102 @@
...
+            // }}
+            sx={{
+              transition: theme =>
+                theme.transitions.create('transform', {
+                  easing: theme.transitions.easing.sharp,
+                  duration: theme.transitions.duration.leavingScreen,
+                }),
+              transform: expanded ? 'rotate(180deg)' : 'rotate(360deg)',
+            }}
+          >
```
joshxg, line 95: Could make a `Spinnerer` component to wrap things in this styling if we need more spinning

packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
```diff
@@ -1,68 +1,102 @@
...
+    oldState: ReturnType<typeof selector>,
+    newState: ReturnType<typeof selector>
+  ) =>
+    oldState?.isExpanded === newState?.isExpanded &&
+    oldState.rowId === newState.rowId;
+
+  const { isExpanded, toggleExpanded } = useTableStore(selector, equalityFn);
+
+  return { isExpanded, toggleExpanded };
+};
```
joshxg, line 39: this is duplicated and doesn't need to be - think I will put it in the `TableContext` folder

packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
```diff
@@ -1,68 +1,102 @@
...
+                rowKey={rowKey}
+              />
+            </TableCell>
+          );
+        })}
+      </TableRow>
+      <Collapse in={isExpanded}>
+        {ExpandContent ? <ExpandContent /> : null}
+      </Collapse>
+    </>
```
joshxg, line 100: I don't really like how this component is here and you need to use the `Expand` column to trigger the state for it. I would have preferred that the expand column itself had the `Expand` UI, but I wasn't sure on the markup for that heh

packages/common/src/ui/layout/tables/context/TableContext.test.ts
```diff
@@ -139,4 +139,109 @@ describe('TableContext', () => {
...
+    const { result } = renderHook<unknown, TableStore>(useStore);
+
+    const { setActiveRows } = result.current;
+
+    act(() => {
+      setActiveRows(['a', 'b', 'c']);
+    });
+
+    expect(result.current.numberExpanded).toBe(0);
+  });
```
joshxg, line 246: quite a few .. just general issues I had during and kept them in, can remove some if you think

packages/common/src/ui/layout/tables/context/TableContext.ts
```diff
@@ -83,9 +102,60 @@ export const createTableStore = (): UseStore<TableStore> =>
...
+
+        return {
+          ...state,
+          numberExpanded: newNumberExpanded,
+          rowState: {
+            ...rowState,
+            [id]: {
+              ...rowState[id],
+              isSelected: rowState[id]?.isSelected ?? false,
+              isExpanded: newExpanded,
```
joshxg, line 130: Ugh... immer? 🤔

packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
mark-prins: yeah. and yeah - not sure either.

packages/common/src/ui/layout/tables/context/TableContext.test.ts
mark-prins: I'd prefer to keep them in.. all appear useful to me.
Am getting some errors when running the tests.. just console errors rather than failing tests, so will look at them later

mark-prins (commented):
🎉  badass! it works!!

few minor comments..
- just made a note of the empty test when reading through. will cause the test suite to barf.. but something to do later
- would be nice if the chevrons had a `cursor: 'pointer'` to show that they are clickable. Now that I think about it.. the headers could do with the same.
- I like the chevrons.. neat idea! Was a little confused by the single ones initially though; would have been more obvious what they were for if they were 'down' ones, like the accordion
- is it possible to trigger from another column: the talk was to show batch in the batch column if there was one batch for that item.. and if more than one to show a `multiple batches` ( or something ) message which you could click to show them all

mark-prins (approved):

joshxg:
Fixed the invalid dom and empty test case and made the chevrons look down.

The `IconButton` and headers all have `cursor: pointer`, for me at least?

Can trigger the expand from anywhere

PR #349, opened 2021-10-31
6 - #311 group by batch 1

joshxg:
Oh, smaller than I thought..! Nice suprise

- I am in the process of adding expanding to the rows, which means we need scroll.
- I had a bit of rough time with this of course, trying to have a flex table which when overflowing, horizontally and vertically scrolls.
- I've gotten there with the scrolling (You can just duplicate columns to see the column scroll - I think some of our columns have minWidths which are too small, though) however I've made the headers non-sticky, now. I've played around with mui `stickHeader` and also apparently you can put `position:'sticky'` on the `td` for a header, however it doesn't seem to work with `display:'flex'` on the header cells. Sigh!
- I spent way too much time messing with this so I've stopped here with the messed up headers which I will come back to
- I've also set the pagination to be `20` by default. I thought to get feedback first. With this scrolling we can have a 'number of rows selector' thing, which people had mentioned in our demo

### DIFF
--- a/packages/invoices/src/OutboundShipment/DetailView/tabs/GeneralTab.tsx
+++ b/packages/invoices/src/OutboundShipment/DetailView/tabs/GeneralTab.tsx
@@ -1,11 +1,11 @@
-import React, { FC, useEffect } from 'react';
+import React, { FC } from 'react';
 import {
   DataTable,
   ObjectWithStringKeys,
   usePagination,
-  useRowRenderCount,
   Column,
   DomainObject,
+  useContentAreaHeight,
 } from '@openmsupply-client/common';
 import { ItemRow } from '../types';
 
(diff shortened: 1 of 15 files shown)


packages/invoices/src/OutboundShipment/DetailView/tabs/GeneralTab.tsx
```diff
@@ -18,12 +17,7 @@ export const GeneralTabComponent: FC<GeneralTabProps<ItemRow>> = ({
   data,
   columns,
 }) => {
-  const numberOfRows = useRowRenderCount();
-  const { pagination } = usePagination(numberOfRows);
-
-  useEffect(() => {
-    pagination.onChangeFirst(numberOfRows);
-  }, [numberOfRows, pagination.first]);
+  const { pagination } = usePagination(20);
```
mark-prins, line 20: I wonder if adding a default to the `usePagination` would be better and avoid having the `20` in a number of different places

mark-prins (approved):
👍  that works well!

PR #350, opened 2021-10-31
7 - Add example batch grouping

joshxg:
Just intended as an example ui 

I'm struggling to understand what is wanted, maybe.

The modal shows which batches are allocated to a line and which quantities and, I assume, is 1 click on a line. We want an expansion which is also 1 click, which shows which batches are allocated to a line and what quantities..

It makes sense to me that we remove the modal and just use an accordion to assign batches. When adding a new line, just add an empty line and the `name` column is the auto complete - when choosing a name, expand the row 

image

The only other idea I have that is wanted for this feature is a read only view of the batches when you expand, but surely not? That would mean that you do a single click to see a single read only view or a single editable view - why bother with the read only view 🤷

### DIFF
--- a/packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
+++ b/packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
@@ -1,7 +1,8 @@
-import React, { FC } from 'react';
+import React from 'react';
 import TableRow from '@mui/material/TableRow';
 import TableCell from '@mui/material/TableCell';
 import { Column } from '../../columns/types';
+import { ExpandContentProps } from '../../types';
 import { DomainObject } from '../../../../../types';
 import { useExpanded } from '../..';
 import { Collapse } from '@mui/material';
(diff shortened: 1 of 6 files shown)


mark-prins:
> why bother with the read only view

for a single line, yep, that's not helpful. the idea was that you could should a `grouped` view which has one line per item and you could expand all rows to see all the batches for all the rows. this is available in desktop in a different tab, though we're unsure whether people use it or not.

In the grouped view, you would see values in the batch & expiry columns if there was one allocated stock line for that item - or a message about multiple batches if not.

mark-prins (commented):
As I understand it - we only want the expanding rows to show a view for 'group by item' rather than the full batch allocation, which will still be in a modal sorry.

Tricky to handle stock lines with different pack sizes when all in one row too

joshxg:
So.. the same rows just disabled and hiding the rows which have issue < 0? (Could even add a 'hide unallocated' button.. or, maybe a "show unallocated?"?) I'm not really sure what to do with this view

Can put the packsize drop down above the rows 🤷 

Seems easier to me - especially like how context is kept while editing

PR #351, opened 2021-10-31
Add debounced column sorting

### LINKED ISSUE #326: Debounce column sorting
Column sorting should be debounced

joshxg:
Fixes #326 

Went this route because I don't think a column should ever be not sortable - even though there is a slight lag when pressed


mark-prins (approved):
👍  
the lag is almost noticeable - I think its fine tho

PR #352, opened 2021-10-31
#331 column format default

### LINKED ISSUE #331: Column alignment should default based on the column format assigned
For example, if I set `format: ColumnFormat.Integer` then the `format` should default to `ColumnAlign.Right`

joshxg:
Fixes #331 

- Goal was to ensure the default column alignment took into account the format
- Then I offroaded into extracting the logic of `useColumns` into a function and just having the hook basially memoise and call this function. This is because I have been meaning to do this for a bit, as it makes some tests easier to write, in some cases - so I updated the reducer test cases
- Then I added some tests for `useColumns`

### DIFF
--- a/packages/common/src/ui/layout/tables/hooks/useColumns/useColumns.test.tsx
+++ b/packages/common/src/ui/layout/tables/hooks/useColumns/useColumns.test.tsx
@@ -1,0 +1,138 @@
+import { renderHook } from '@testing-library/react-hooks';
+import { ColumnAlign, ColumnFormat } from '../..';
+import { useColumns } from '../../..';
+import { DomainObject } from '../../../../..';
+
+interface Test extends DomainObject {
+  id: string;
+}
+
+describe('useColumns', () => {
+  it('assigns sensible defaults for an unspecified column', () => {
+    const { result } = renderHook(() => useColumns<Test>([{ key: 'default' }]));
+
+    const defaults = {
+      format: ColumnFormat.Text,
+      sortable: true,
+      sortInverted: false,
+      sortDescFirst: false,
+      align: ColumnAlign.Left,
+      width: 100,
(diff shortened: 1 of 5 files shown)


packages/common/src/ui/layout/tables/hooks/useColumns/useColumns.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+      useColumns<Test>([{ key: 'default', format: ColumnFormat.Integer }])
+    );
+
+    const defaults = {
+      format: ColumnFormat.Integer,
+      sortable: true,
+      sortInverted: false,
+      sortDescFirst: false,
+      align: ColumnAlign.Right,
+      width: 100,
```
mark-prins, line 42: though possibly integers don't need to be 100px default?

packages/common/src/ui/layout/tables/hooks/useColumns/useColumns.test.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+  it('uses the correct width and min width if specified', () => {
+    const { result } = renderHook(() =>
+      useColumns<Test>([{ key: 'default', width: 200, minWidth: 100 }])
+    );
+    const defaults = { width: 200, minWidth: 100 };
+    expect(result.current[0]).toEqual(expect.objectContaining(defaults));
+  });
+
+  it('has a stable reference when re-rendering', () => {
```
mark-prins, line 131: oh, very nice!

mark-prins (commented):
🎉  sweeet! that's very nice indeed

mark-prins (approved):
oops, mis-clicked the `Approve`

joshxg:
Made the default width for an int 60 🤷  
a884f16

Issue #158, opened 2021-09-13
Outbound: Comment


mark-prins:
Another nice little table challenge: adding the comment column. We don't have designs which include both the colour and the comment. The colour was done earlier, and the comment later - the combo appears to have been missed.

If a row has a comment - a mouse over (press on mobile?) to show the comment and a different treatment on the icon to show there is a value

<img width="705" alt="Screen Shot 2021-09-14 at 9 40 22 AM" src="https://user-images.githubusercontent.com/9192912/133160265-196e8998-b764-43b3-bbcb-3fe8e21c3546.png">

and the idea of clicking an empty speech box icon to add a new comment. not sure why some rows have no comment icon in this case

![Screen Shot 2021-09-14 at 9 40 56 AM](https://user-images.githubusercontent.com/9192912/133161036-ccaacc11-1fe3-42ae-82ed-64980268f42e.png)

joshxg:
Oh cool. Wonder if the color column would be better just as it's own, small column like this with no header. Probably.

Maybe: 
- Desktop: Hover and click to open the edit dialog
- Mobile: Click to open the edit dialog (Which is read only when finalised)

joshxg:
Also wondering if just have two small columns one for the color and one for the comment would be OK - probably need to check it out to know!

mark-prins:
Yes, I wondered if two small headerless columns would work. I think so.
Wondering about trying it to see how it would work: the item was missed from the design chat this morning, so likely won't get any design direction for a couple of weeks.

richardmoizeau:
email sent to Manu to (try to) finalise the shipment UI and I did request a preview where color picker and comment icon but no response from Manu (family issues). Would you be happy to give it a go ?

mark-prins:
Possibly we need to bring the comment column back after Craig's comments this morning?

richardmoizeau:
Yes, Shipment-level comment should be in a column as per Craig's instructions. I believe we have a consistent way to deal with comments so I would also recommend to have the Shipment Line level comment in a column too.

joshxg:
I think the comment on a line at the detail level will ensure there is horizontal scroll on the detail views with more columns like a customer requisition, which I would think will make tabbing to the next cell and general usage annoying.

Maybe worth checking if users are really using line level comments to know if we should clutter the ui with this?

PR #353, opened 2021-10-31
Add MessageSquare icon

joshxg:
Starts #158 

So I just wanted to PR this so we can see the problem:

image

In the zeplin:

image

- So, each row doesn't have a comment square and the popover isn't editable. Where do we edit these? At the moment the only place we could have this is in the `BatchesTable` - but I am not sure there is enough room?
- Then, since each line is an aggregate of batches, what comment shows?
- Could show it in the row expansion, whatever that is
- In mSupply, the note shows as a column in a CI (Not SI) but I can't find a way to edit it.. 🤷


mark-prins:
I _think_ the idea is this

Screen Shot 2021-11-01 at 2 50 16 PM

[zeplin](https://app.zeplin.io/project/60e44fff6fa942168092f1ff/screen/610a7f901c36e2b128f4bb2c)

in which the comment is read only on hover and editable on click - at least for web. maybe just drop the hover for mobile? or use a long-press for edit?

packages/invoices/src/OutboundShipment/DetailView/reducer.ts
```diff
@@ -152,6 +153,19 @@ export const reducer = (
           break;
         }
 
+        case ActionType.UpdateComment: {
+          const { payload } = action;
+          const { rowKey, comment } = payload;
```
mark-prins, line 158: :nit I'd prefer `rowId` as this is matching against `id`

mark-prins (approved):
👍  I think it looks great!

joshxg:
Sure

Still unsure what comment is to be shown (Maybe i'm still unclear on what the rows are) and also when the comment message icon shows. In the zepplin it appears to only be shown when there is a comment - I suppose we could be going for the mSupply style: a comment column and no way to edit it?

mark-prins:
I think it should always show - maybe with the icon in orange if there is already a comment and in grey if there isn't one? Either way, clicking on it would bring up the edit modal.

Though...

The API doesn't have a comment of course - and in desktop there isn't one on the `trans_line` there's only the `item_line_note` - which would make sense perhaps, if we brought that through as a read only 'comment' for the stock line 🤷  In which case, the icon only shows if there is a note. this all fits in better with existing data

joshxg:
Ah, is `[trans_line]note` the note from the `[item_line]`?

joshxg:
Finally figured out how to show a note in mSupply! 🎉 

So the CI does show the `[item_line]note` 

since it seems each line represents multiple batches, I suppose we need to change the design to have subheadings for each batch?

mark-prins:
aha. yes - I hadn't connected the dots.. the note is for an `[item_line]` not an `[item]`!
That would be a good option - have a heading of the batch (in bold) and then the note

Issue #314, opened 2021-10-22
Detail View: bulk actions - allocate placeholders


mark-prins:
Add the bulk actions to the detail view. Should have the following options in it:
- Delete
- Auto Allocate Stock

though, sorry, I don't have detail for what the auto allocation will actually do. This is the new home for the 'grinder' which allocates stock to placeholder lines.. so that. But for selected lines only rather than all, as currently is the case.

joshxg:
Thinking about this more..

At the moment we're querying for all items and all their batches and have that query sitting around to utilise. However I am not sure in practise this will be possible - I imagine stores might have a ton of item lines. 

If this is right, then we either need to have some sort of query for all the items lines for each placeholder item/line, which we can then FEFO assign.

I wonder if this actually might be better as a mutation .. 'assign placeholder lines'

mark-prins:
+1 for a mutation. This feels like a task which requires business logic and therefore should live in the API

joshxg:
Cool.

We will need to save before this, so the backend has all the data. So we'll need at least a couple of mutations, I suppose.

- Blocking spinner
- Send request to save
- Send request for mutation `assignPlaceholders`

?

joshxg:
hmm maybe the blocking isn't needed 🤔

joshxg:
b/e tracking: https://github.com/openmsupply/openmsupply-client/issues/385

PR #354, opened 2021-11-01
#314 bulk actions 1 selection

joshxg:
Starts #314

- Adds the drop down bulk select and a checkbox column to the `DetailView`
- Adds a pseudo delete, for now - hopefully can do actual mutations soon
- Not sure about the state shape - think it could be nicer


packages/invoices/src/OutboundShipment/DetailView/Toolbar.tsx
```diff
@@ -48,6 +81,11 @@ export const Toolbar: FC<ToolbarProps> = ({ draft }) => {
             />
           </Box>
         </Grid>
+        <DropdownMenu label="Select">
```
mark-prins, line 84: can you add `disabled` if the OS is uneditable please

packages/invoices/src/OutboundShipment/DetailView/Toolbar.tsx
```diff
@@ -5,19 +5,52 @@ import {
...
+  }));
+
+  const deleteAction = () => {
+    if (selectedRows && selectedRows?.length > 0) {
+      selectedRows.forEach(line => draft.deleteLine?.(line));
+      const successSnack = success(`Deleted ${selectedRows?.length} lines`);
+      successSnack();
+    } else {
+      const infoSnack = info('Select rows to delete them');
+      infoSnack();
```
mark-prins, line 41: great names - liking the `infoSnack` and `successSnack`! 😄 
:nit just noticed after commenting.. translation for the message?

mark-prins (approved):
👍  
yeah, a thing of beauty. very nice to use 😃  - and looks like it will slot in nicely when the mutations are possible

PR #355, opened 2021-11-01
#339 cache busting

### LINKED ISSUE #339: Trouble with browser caching
I've had a couple of complaints about a 'white screen' on the demo site, which turned out to be an error of some kind due to browser caching. A hard refresh of the browser fixed the issue - trouble is, the users didn't know about hard refresh or, if they did, didn't think to try.

It is probably time to append a build string to the resource URLs to prevent this; ideally each bundle will only have the URL change if it is updated, but failing that - a change on every build will be fine.

mark-prins:
Fixes #339 

Adding a `[contenthash]` to remove issues some reviewers are having on the demo site.

### DIFF
--- a/packages/host/webpack.config.js
+++ b/packages/host/webpack.config.js
@@ -37,7 +37,8 @@ module.exports = env => {
     },
     output: {
       publicPath: isProduction ? '/' : 'http://localhost:3003/',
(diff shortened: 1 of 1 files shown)


joshxg:
sweet!

joshxg (approved):

PR #357, opened 2021-11-02
add status values

mark-prins:
Scratches #an-itch

Just adding this in as I've been getting errors raised about invalid status values


mark-prins:
no longer required

Issue #356, opened 2021-11-02
Organise types 


joshxg:
Our types, specifically the 'domain' ones, are somewhat unstructured and messy

Some are in common and some are in `invoices` etc.

1. Start using graphql code gen to generate types
2. Put them in common. I think they would likely be better in specific places - i.e. the outbound shipment types with outbound shipment, however it seems easier to just codegen to the same place, for now.
3. Don't add codegen to the development workflow. I think it's better, but when the API is more stable and where we don't need to add changes ourselves, as much. i.e. `hold`

I also personally get annoyed with the shapes of the returned objects, in that I have to access things like `invoice.lines.nodes`. Personally, I wonder if it would be better to map each response into a new shape without the `nodes` 🤷 .. though, this can be another issue..

mark-prins:
just to chip in - would love to map the returned object to a better shape!!

PR #358, opened 2021-11-02
#356 organise types

joshxg:
#356 Start - will do one more PR after this one.

- This just adds the packages for code generating. I had a couple of issues with some hoisting which was annoying, which meant I removed the lock file which then meant notistack & webpack-cli went to a weird version and I needed to upgrade those.
- Also had an offroad with a button styling
- I added `graphql-import-node` to be able to import the graphql file directly into the mock server
- I code generated from the graphql schema we had which is slightly different to the current server version: https://www.diffchecker.com/32OFduUL
- I'm not using any of the generated types, yet. I'll do that in the next PR.


mark-prins (approved):
👍  wooo! that's very neat & looks really useful

PR #363, opened 2021-11-02
Add use of typed invoice detail query

joshxg:
Well this is turning out to be bigger than I anticipated :)

I have only converted over the detail query here and thought to make a new PR for others as it was a lot more changes than I expected.

### DIFF
--- a/packages/common/src/hooks/useRowRenderCount/useRowRenderCount.test.tsx
+++ b/packages/common/src/hooks/useRowRenderCount/useRowRenderCount.test.tsx
@@ -1,4 +1,4 @@
-import React, { FC, ReactNode } from 'react';
+import React, { FC } from 'react';
 import { act } from 'react-dom/test-utils';
 import { waitFor } from '@testing-library/dom';
 import { renderHook } from '@testing-library/react-hooks';
@@ -7,6 +7,7 @@ import { useTheme } from '../../styles';
 import { TestingProvider, TestingRouter } from '../../utils';
 import { useRowRenderCount } from './useRowRenderCount';
 import { useAppBarRectStore } from '../useAppBarRect';
+import { Route } from 'react-router';
 
 type ThemeChangererProps = {
   paginationRowHeight: number;
@@ -42,20 +43,19 @@ describe('useRowRenderCount', () => {
       paginationRowHeight = 0,
       saveButtonRowHeight = 0,
       footerHeight = 0
-    ) =>
-    ({ children }: { children: ReactNode[] }) => {
+    ): FC =>
+    ({ children }) => {
       return (
         <TestingProvider>
-          <TestingRouter initialEntries={['']}>
-            <ThemeChangerer
-              paginationRowHeight={paginationRowHeight}
-              dataRowHeight={dataRowHeight}
-              headerRowHeight={headerRowHeight}
-              footerHeight={footerHeight}
-              saveButtonRowHeight={saveButtonRowHeight}
-            >
-              {children}
-            </ThemeChangerer>
+          <ThemeChangerer
+            paginationRowHeight={paginationRowHeight}
+            dataRowHeight={dataRowHeight}
+            headerRowHeight={headerRowHeight}
+            footerHeight={footerHeight}
+            saveButtonRowHeight={saveButtonRowHeight}
+          />
+          <TestingRouter initialEntries={['/']}>
+            <Route path="/" element={<>{children}</>} />
           </TestingRouter>
         </TestingProvider>
       );
--- a/packages/common/src/operations.graphql
+++ b/packages/common/src/operations.graphql
@@ -8,6 +8,15 @@ query invoice($id: String!) {
       finalisedDatetime
       invoiceNumber
 
+      draftDatetime
+      allocatedDatetime
+      pickedDatetime
+      shippedDatetime
+      deliveredDatetime
+
+      hold
+      color
+
       lines {
         ... on InvoiceLineConnector {
           nodes {
@@ -18,6 +27,7 @@ query invoice($id: String!) {
             itemCode
             itemId
             itemName
+            itemUnit
             numberOfPacks
             packSize
             sellPricePerPack
--- a/packages/common/src/operations.graphql
+++ b/packages/common/src/operations.graphql
@@ -71,57 +71,6 @@ query names {
   }
 }
 
-# query invoices(
-#   $first: Int
-#   $offset: Int
-#   $key: InvoiceSortFieldInput!
-#   $desc: Boolean
-# ) {
-#   invoices(
-#     page: { first: $first, offset: $offset }
-#     sort: { key: $key, desc: $desc }
-#   ) {
-#     ... on ConnectorError {
-#       __typename
-#       error {
-#         description
-#         ... on DatabaseError {
-#           __typename
-#           description
-#           fullError
-#         }
-#       }
-#     }
-#     ... on InvoiceConnector {
-#       nodes {
-#         id
-#         invoiceNumber
-#         finalisedDatetime
-#         entryDatetime
-#         confirmedDatetime
-#         comment
-#         otherPartyName
-#         status
-#         theirReference
-#         type
-#         pricing {
-#           ... on NodeError {
-#             __typename
-#             error {
-#               description
-#             }
-#           }
-#           ... on InvoicePricingNode {
-#             __typename
-#             totalAfterTax
-#           }
-#         }
-#       }
-#       totalCount
-#     }
-#   }
-# }
-
 query invoices(
   $first: Int
   $offset: Int
--- a/packages/common/src/operations.graphql
+++ b/packages/common/src/operations.graphql
@@ -71,6 +71,57 @@ query names {
   }
 }
 
+# query invoices(
+#   $first: Int
+#   $offset: Int
+#   $key: InvoiceSortFieldInput!
+#   $desc: Boolean
+# ) {
+#   invoices(
+#     page: { first: $first, offset: $offset }
+#     sort: { key: $key, desc: $desc }
+#   ) {
+#     ... on ConnectorError {
+#       __typename
+#       error {
+#         description
+#         ... on DatabaseError {
+#           __typename
+#           description
+#           fullError
+#         }
+#       }
+#     }
+#     ... on InvoiceConnector {
+#       nodes {
+#         id
+#         invoiceNumber
+#         finalisedDatetime
+#         entryDatetime
+#         confirmedDatetime
+#         comment
+#         otherPartyName
+#         status
+#         theirReference
+#         type
+#         pricing {
+#           ... on NodeError {
+#             __typename
+#             error {
+#               description
+#             }
+#           }
+#           ... on InvoicePricingNode {
+#             __typename
+#             totalAfterTax
+#           }
+#         }
+#       }
+#       totalCount
+#     }
+#   }
+# }
+
 query invoices(
   $first: Int
   $offset: Int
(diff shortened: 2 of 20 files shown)


packages/common/src/hooks/useRowRenderCount/useRowRenderCount.test.tsx
```diff
@@ -42,20 +43,19 @@ describe('useRowRenderCount', () => {
...
-            </ThemeChangerer>
+          <ThemeChangerer
+            paginationRowHeight={paginationRowHeight}
+            dataRowHeight={dataRowHeight}
+            headerRowHeight={headerRowHeight}
+            footerHeight={footerHeight}
+            saveButtonRowHeight={saveButtonRowHeight}
+          />
+          <TestingRouter initialEntries={['/']}>
+            <Route path="/" element={<>{children}</>} />
```
joshxg, line 58: Part of why this was larger than expected was because there were a few tests that were failing, I think, from my previous PRs.

I wanted to make sure everything was passing from this somewhat sweeping change, so I fixed them up in here.

This test was failing because of some change where `initialEntries={['']}` was causing nothing to render. I suspect from my change of the lock file 🙄

packages/common/src/operations.graphql
```diff
@@ -18,6 +27,7 @@ query invoice($id: String!) {
             itemCode
             itemId
             itemName
+            itemUnit
```
joshxg, line 30: this adds some data which we weren't actually querying for into the detail query

packages/common/src/schema.graphql
```diff
@@ -323,9 +323,13 @@ type InvoiceNode {
   entryDatetime: DateTime!
   confirmedDatetime: DateTime
   finalisedDatetime: DateTime
+
   lines: InvoiceLinesResponse!
   pricing: InvoicePriceResponse!
 
+  color: String!
+  hold: Boolean!
+
```
joshxg, line 332: adding these to the gql schema

packages/common/src/types/index.ts
```diff
@@ -73,26 +57,47 @@ export type Store = {
...
+
+  expiryDate?: string | null;
+
+  batch?: string | null;
+
+  stockLine?: StockLine | null;
+
+  location?: string;
+  comment?: string;
+}
```
joshxg, line 80: this now aligns with the `InvoiceLineNode`, where it would arguably be better to use that instead. But, for now, I've just converted this over for now.

Personally, apart from the API fns/layer, I would rather not know gql exists throughout the rest of the app, so I would like to use the generated types for typing the api fns, and then map these into our internal types. 

I think it keeps it cleaner and also means when we need to create 'pseudo' entities, we don't have to create an invoice with `lines: { nodes: [] }` and things like that.. 🤷 

There would also be an additional type, though, which has the `update` fn's etc, attached. So, maybe it's getting all a bit complicated

packages/common/src/types/index.ts
```diff
@@ -73,26 +57,47 @@ export type Store = {
...
+
 export interface Invoice extends DomainObject {
   id: string;
   color: string;
-  comment: string;
-  theirReference: string;
+  comment?: string | null;
+  theirReference?: string | null;
```
joshxg, line 86: undefined | null | string is a bit ugh

packages/common/src/types/schema.ts
```diff
@@ -401,34 +385,36 @@ export type InvoiceLinesResponse = ConnectorError | InvoiceLineConnector;
...
-   */
+  Allocated = 'ALLOCATED',
   Confirmed = 'CONFIRMED',
-  /**
-   * For customer invoices: In DRAFT mode only the available_number_of_packs in a stock line gets
-   * updated when items are added to the invoice.
-   */
+  Delivered = 'DELIVERED',
   Draft = 'DRAFT',
-  /** A FINALISED invoice can't be edited nor deleted. */
```
joshxg, line 430: I guess it initially generated with the comments and then decided it didn't want to anymore..

packages/common/src/types/schema.ts
```diff
@@ -401,34 +385,36 @@ export type InvoiceLinesResponse = ConnectorError | InvoiceLineConnector;
...
   Finalised = 'FINALISED',
+  Picked = 'PICKED',
+  Shipped = 'SHIPPED',
```
joshxg, line 417: For now, just piling on the statuses. Will need to make a new issue to align *again* with the new new ones..

packages/common/src/types/schema.ts
```diff
@@ -915,6 +881,13 @@ export type InvoiceQuery = {
         entryDatetime: any;
         finalisedDatetime?: any | null | undefined;
         invoiceNumber: number;
+        draftDatetime?: any | null | undefined;
+        allocatedDatetime?: any | null | undefined;
+        pickedDatetime?: any | null | undefined;
+        shippedDatetime?: any | null | undefined;
+        deliveredDatetime?: any | null | undefined;
```
joshxg, line 888: this is whats generated which doesn't seem ideal 🤔

joshxg: Ah, IC, the custom scalars `DateTime` and `NaiveDate` are of type `any`. I'm sure there must be a way to fix that.

packages/common/src/ui/components/domain/StatusCrumbs/StatusCrumbs.tsx
```diff
@@ -13,7 +13,7 @@ import { usePopover } from '../../popover';
 
 interface StatusCrumbsProps<StatusType extends string> {
   statuses: StatusType[];
-  statusLog: Record<StatusType, string | null>;
+  statusLog: Record<StatusType, string | null | undefined>;
```
joshxg, line 16: date times are now null | undefined ..

packages/common/src/ui/layout/tables/columns/PaginationRow/PaginationRow.test.tsx
```diff
@@ -149,36 +149,6 @@ describe('PaginationRow', () => {
...
-
-    const node = getByLabelText(/ellipsis/i);
-
-    act(() => {
-      userEvent.click(node);
-    });
-
-    expect(onChange).toBeCalledTimes(0);
-  });
-
```
joshxg, line 181: Uhh. This test was to prevent the ellipsis of the pagination row being clickable. Somehow it's started to fail and now I can't 'find' the ellipsis button with the dom elements and didn't want to spend more time making it work.....

packages/common/src/ui/layout/tables/components/Header/Header.test.tsx
```diff
@@ -92,10 +91,12 @@ describe('HeaderRow', () => {
...
-    const idHeader = getByRole('button', { name: /name/i });
+    const nameHeader = getByRole('button', { name: /name/i });
 
-    act(() => userEvent.click(idHeader));
+    userEvent.click(nameHeader);
 
-    expect(onSortBy).toBeCalledWith(expect.objectContaining({ key: 'name' }));
+    waitFor(() => {
+      expect(onSortBy).toBeCalledWith(expect.objectContaining({ key: 'name' }));
+    });
```
joshxg, line 100: this was failing because I added debounce to the headers and didn't update the test.

packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
```diff
@@ -16,7 +16,7 @@ import { getOutboundShipmentDetailViewApi } from '../../api';
 import { GeneralTab } from './tabs/GeneralTab';
 import { ItemDetailsModal } from './modals/ItemDetailsModal';
 
-import { ItemRow } from './types';
+import { InvoiceLineRow } from './types';
```
joshxg, line 19: as you'll see below, I've renamed `ItemRow` to `InvoiceLineRow` because `ItemRow` was an old name and wasn't very representative of what it was. Though, `InvoiceLineRow` isn't better.

Maybe head towards something like:
`InvoiceNode` -> `Invoice` -> `OutboundShipment`
`InvoiceLineNode` -> `InvoiceLine` -> `OutboundShipmentLine`

Where the `XNode` are GQL types, `Invoice` are simple 'mappings' into the more simple forms, and `OutboundShipmentX` can be the more specific type with methods attached and any extra details?

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
```diff
@@ -159,7 +165,9 @@ export const ItemDetailsModal: React.FC<ItemDetailsModalProps> = ({
     const invoiceLines = batchRows.map(batch =>
       getInvoiceLine('', selectedItem, batch, Number(values[batch.id] || 0))
     );
-    invoiceLines.filter(line => line.quantity > 0).forEach(upsertInvoiceLine);
+    invoiceLines
+      .filter(line => line.numberOfPacks > 0)
+      .forEach(upsertInvoiceLine);
```
joshxg, line 170: hopefully these changes in the `ItemDetailsModal` are OK

packages/invoices/src/OutboundShipment/DetailView/reducer.test.ts
```diff
@@ -5,85 +5,105 @@ import {
 } from '@openmsupply-client/common';
 import { placeholderInvoice } from './index';
 import { reducer, OutboundShipmentStateShape, OutboundAction } from './reducer';
-import { ItemRow } from './types';
+import { InvoiceLineRow } from './types';
 
 describe('DetailView reducer', () => {
-  const lines: ItemRow[] = [
+  const lines: InvoiceLineRow[] = [
```
joshxg, line 11: now that I look at it. Why didn't I just make a simpler type to test with?

packages/invoices/src/api.ts
```diff
@@ -245,14 +250,27 @@ export const listQueryFn = async <T extends ObjectWithStringKeys>(queryParams: {
...
+          : [],
+      pricing: {
+        totalAfterTax:
+          invoice.pricing.__typename === 'InvoicePricingNode'
+            ? invoice.pricing.totalAfterTax
+            : 0,
+      },
+    };
+  } else {
+    throw new Error('uhoh');
```
joshxg, line 272: So uhh..

We're typed, but this is.. not great.

I am just not sure about this conversion of GQL types into 'normal types'  to use without the app. I am quite sure it is a good idea, but I am hesistent as I don't want to make things confusing.

I believe here we can simple throw an error if there is an error when getting the invoice price or the lines or if there is an error with the invoice itself. In all of those cases it is surely something wrong with the invoice, in which case we just have to show a 'Nope'!.

But, we can likely do something with the description. Maybe..

joshxg:
ah crap.. I accidentally committed the list view query change also.......

packages/common/src/types/index.ts
mark-prins: it does get a little complicated, though I think worth it. adding in the `update` and other funcs isn't too much of a drama and once we establish a pattern and naming convention then I think it will be pretty obvious
 
I like the separation of the API types from internal types - and would prefer to have a different shape to the types internally, so would like to keep it. 
just worried that there is some duplication, and whether it might bring in some confusion: where someone in the front end codebase is talking about x and that is actually a y in the API / gql which actually maps to database field z. I think all that is pretty minor.

packages/common/src/types/index.ts
mark-prins: yeah. needs both, I take it? can't just have `comment: string | null`?

packages/common/src/ui/components/domain/StatusCrumbs/StatusCrumbs.tsx
mark-prins: 😢

packages/common/src/ui/layout/tables/columns/PaginationRow/PaginationRow.test.tsx
mark-prins: odd. it used to be a button, now just a `div` 🤷

packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
mark-prins: that sounds good!

packages/invoices/src/api.ts
```diff
@@ -222,40 +118,79 @@ export const listQueryFn = async <T extends ObjectWithStringKeys>(queryParams: {
...
+                ? invoice.pricing.totalAfterTax
+                : 0,
+          },
+        };
+      }),
+    };
+  }
 
-  return invoices;
+  throw new Error('uh oh');
```
mark-prins, line 152: 🤣

packages/invoices/src/api.ts
mark-prins: needs to be done somewhere - better to have it in one place and have the rest of the app work with normal data

mark-prins (approved):
👍  wow, thanks for doing all that. a big job - worth it though! it's setting up some good patterns and making life much easier for us

PR #364, opened 2021-11-03
Add item query to use code gen

joshxg:
This is the last of the query shifting.

This has ended up being so much bigger than I thought, sorry.

From here I would like to:
1. Create a more common way of creating the sdk/api object. I am thinking to make a `context` which the host can create the `sdk/api` object on load, and then each package can use a `useApi` hook. This also has the added bonus of allow the host to inject the credentials into the `api` object so each package doesn't need to worry about it (also the URL)
2. Sort the sorting out. I think if I get to it soon, I'm just going to add a `sortKey` to the `column` - I don't want to wait..
3. Organise the different enums and use them a bit better in our code - i.e. the statuses
4. Sort the mapping out. I'm sure there's a better pattern to be using here to map things, taking away the `nodes`..
5. Organise the errors- uh oh - betterer. '
6. Add as much as I can to the graphql schema of things we want
7. I know there's a better way to be generating with multiple packages in the yaml config, but it's going to take me some time to figure it out. I think it might be best to just skip this step.
8. Sort out the mutations. I can probably dig into remote server and find the schema and add it to ours
9. Finally, I think I might PR the graphql schema into the remote server repo. Unsure on this step though..

### DIFF
--- a/packages/common/src/operations.graphql
+++ b/packages/common/src/operations.graphql
@@ -178,3 +178,88 @@ query names(
     }
   }
 }
+
+query items(
+  $first: Int
+  $offset: Int
+  $key: ItemSortFieldInput!
+  $desc: Boolean
+) {
+  items(
+    page: { first: $first, offset: $offset }
+    sort: { key: $key, desc: $desc }
+  ) {
+    ... on ConnectorError {
+      __typename
+      error {
+        description
+        ... on DatabaseError {
+          __typename
+          description
+          fullError
+        }
+        ... on PaginationError {
+          __typename
+          description
+          rangeError {
+            description
+            field
+            max
+            min
+          }
+        }
+      }
+    }
+    ... on ItemConnector {
+      __typename
+      nodes {
+        __typename
+        availableBatches {
+          ... on ConnectorError {
+            __typename
+            error {
+              description
+              ... on DatabaseError {
+                __typename
+                description
+                fullError
+              }
+              ... on PaginationError {
+                __typename
+                description
+                rangeError {
+                  description
+                  field
+                  max
+                  min
+                }
+              }
+            }
+          }
+          ... on StockLineConnector {
+            nodes {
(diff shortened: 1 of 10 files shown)


packages/common/src/operations.graphql
```diff
@@ -178,3 +178,88 @@ query names(
...
+              batch
+              costPricePerPack
+              expiryDate
+              id
+              itemId
+              packSize
+              sellPricePerPack
+              storeId
+              totalNumberOfPacks
+              onHold
```
mark-prins, line 252: I had to add `name` here, and to the `StockLineNode` in the schema: the Item query below as insisting that there should be a `name` on each available batch node

packages/system/src/Item/ListView/ListView.tsx
```diff
@@ -27,64 +31,33 @@ const listQueryFn = async ({
...
+    key,
+    desc: sortBy.isDesc,
+  });
+
+  if (items.__typename === 'ItemConnector') {
+    const itemRows: Item[] = items.nodes.map(item => ({
+      ...item,
+      availableQuantity: 0,
+      unit: '',
+      availableBatches:
```
mark-prins, line 49: Nice! 

There's another use of the item list unfortunately - the add item modal has another version of the query and assigning an item throws an error. My solution is to export this query and use it in that modal.. but also requires updating of the `useListData` hook to remove the page size, otherwise only 20 items show!

mark-prins (approved):
The add item modal is suffering with the change - I can update that if you'd rather just merge this and move on 😃

joshxg:
Ugh! Sorry! Too much of a rush before going to pick up ada 🤣 

I fixed up the query by just copy and pasting.. 🤔 Will look more closely when moving all the item things to system/item..

As for the `name`.. it doesnt seem to be used anywhere, so I just removed it 🤷

PR #370, opened 2021-11-04
#368 map gql returns

### LINKED ISSUE #368: Mapping query return values
Currently the mapping of returned values from gql queries is quite messy and I am sure there must be a nicer way!

joshxg:
Fixes #368 

*Sigh* cool diff..

Wondering what you think of this...

I wanted to make some sort of generic transformer which gets rid of all the 'nodes' but it ended up being way to messy. So went with the simpler duplicate a bunch of stuff approach..

currently if you have an error in one the guarded fields, an error is thrown - i.e.

```
    context.data({
      invoices: {
        ...result,
        error: { description: 'Error!' },
        __typename: 'NotInvoiceConnector',
      },
    })
```

will show "no records found" with a toast of the description..

I think it needs more work, handling all the different errors, but if there are errors in any of these guarded fields, I think   there's basically a big fuck up..

### DIFF
--- a/packages/common/src/hooks/useListData/useListData.test.tsx
+++ b/packages/common/src/hooks/useListData/useListData.test.tsx
@@ -30,7 +30,7 @@ describe('useListData', () => {
   );
 
   const ServerErrorApi: ListApi<Test> = {
-    onQuery: () => async () => {
+    onRead: () => async () => {
       return await request('http://localhost:4000', getServerErrorQuery());
     },
     onDelete: async () => {},
@@ -47,7 +47,7 @@ describe('useListData', () => {
   `;
 
   const PermissionErrorApi: ListApi<Test> = {
-    onQuery: () => async () => {
+    onRead: () => async () => {
       return await request(
         'http://localhost:4000',
         getPermissionErrorQuery(),
--- a/packages/common/src/types/index.ts
+++ b/packages/common/src/types/index.ts
@@ -69,9 +69,7 @@ export interface InvoiceLine extends DomainObject {
 
   batch?: string | null;
 
-  stockLine?: StockLine | null;
-
-  location?: string;
+  location?: string | null;
   comment?: string;
 }
 
--- a/packages/common/src/types/index.ts
+++ b/packages/common/src/types/index.ts
@@ -31,8 +31,6 @@ export interface StockLine extends DomainObject {
   costPricePerPack: number;
   expiryDate?: string | null;
   batch?: string | null;
-  // item: Item;
-  name: string;
   packSize: number;
   sellPricePerPack: number;
   totalNumberOfPacks: number;
--- a/packages/invoices/src/OutboundShipment/ListView/api.ts
+++ b/packages/invoices/src/OutboundShipment/ListView/api.ts
@@ -1,0 +1,158 @@
+import {
+  ObjectWithStringKeys,
+  InvoicesQuery,
+  request,
+  gql,
+  batchRequests,
+  SortBy,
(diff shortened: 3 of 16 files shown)


packages/common/src/hooks/useListData/useListData.test.tsx
```diff
@@ -30,7 +30,7 @@ describe('useListData', () => {
   );
 
   const ServerErrorApi: ListApi<Test> = {
-    onQuery: () => async () => {
+    onRead: () => async () => {
```
joshxg, line 33: ahhh.. 🤷

packages/common/src/types/index.ts
```diff
@@ -71,9 +69,7 @@ export interface InvoiceLine extends DomainObject {
 
   batch?: string | null;
 
-  stockLine?: StockLine | null;
```
joshxg, line 74: Turns out this isn't used..?

packages/invoices/src/OutboundShipment/ListView/api.ts
```diff
@@ -0,0 +1,158 @@
+import {
```
joshxg, line 1: Just shifted this out

packages/invoices/src/OutboundShipment/ListView/api.ts
```diff
@@ -0,0 +1,158 @@
...
+
+const pricingGuard = (pricing: InvoicePriceResponse) => {
+  if (pricing.__typename === 'InvoicePricingNode') {
+    return pricing;
+  } else if (pricing.__typename === 'NodeError') {
+    throw new Error(pricing.error.description);
+  } else {
+    throw new Error('Unknown');
+  }
+};
```
joshxg, line 29: so I put these guards (and some types have optional `__typename` and I can't figure out why, but thats the unknown error..)

Maybe it's wrong but I figure if there's been an error fetching the invoices 'totalAfterTax' then there's some error in the server .. 🤷  - but maybe I should instead be returning some different value

packages/common/src/types/index.ts
mark-prins: ah, no. I did wonder about that

packages/invoices/src/OutboundShipment/ListView/api.ts
mark-prins: I'd prefer not to have too much logic in the front end, trying to interpret errors and behave differently depending on exactly what has gone wrong. Useful in some specific cases, but as a general rule - I think better to let the user know that things are just plain broke.

So yes - I agree. Just throw the error

packages/invoices/src/api.ts
```diff
@@ -1,60 +1,51 @@
...
   Name,
   SortBy,
-  ListApi,
   Invoice,
   GraphQLClient,
   getSdk,
-  InvoiceSortFieldInput,
-  InvoiceRow,
   NameSortFieldInput,
+  InvoiceQuery,
```
mark-prins, line 10: is there a reason that `ListView` has a separate `api.ts` and the `DetailView` is instead using this `api` in the root of invoices? I'm thinking now to have an `api.ts` in each, what do you think?

mark-prins (approved):
👍  Great!
That's all super useful, makes worthing with the gql api much neater

packages/invoices/src/api.ts
joshxg: Ah yes that was my intention. Going to jump on shifting the items/customer stuff shortly so will do it there!

PR #371, opened 2021-11-04
#367 sort out sorting

### LINKED ISSUE #367: Sort out sorting
Currently the queries we're using are not quite aligned with the backends use of enums for sort keys. I think it's best to just go ahead and use the enums and add a `sortKey` to the column. Who knows, maybe that'll be a good thing

joshxg:
Fixes #367 

- This aligns with the backend sorting
- I've just made it so we map the sort key to the enum of what the backend provides in the api layer. I was going to add a `sortKey` to the column, but then I would need to add another generic to a bunch of stuff which defines the set of sort keys, and then also we would have the client side sorting needing to use different types.. I figured to just do this, which isn't as 'safe' in terms of ensuring that a columns sort key is guaranteed to be sortable but is simpler and easier, I think

### DIFF
--- a/packages/common/src/ui/layout/tables/utils/ColumnDefinitionSetBuilder.ts
+++ b/packages/common/src/ui/layout/tables/utils/ColumnDefinitionSetBuilder.ts
@@ -65,11 +65,13 @@ const getColumnLookup = <T extends DomainObject>(): Record<
     key: 'name',
     label: 'label.name',
     width: 75,
+    sortable: false,
   },
   invoiceNumber: {
     key: 'invoiceNumber',
     label: 'label.invoice-number',
     width: 50,
+    sortable: false,
   },
   type: {
     label: 'label.type',
@@ -100,11 +102,13 @@ const getColumnLookup = <T extends DomainObject>(): Record<
     width: 100,
     format: ColumnFormat.Currency,
     align: ColumnAlign.Right,
+    sortable: false,
   },
   comment: {
     label: 'label.comment',
     key: 'comment',
     width: 250,
+    sortable: false,
   },
   selection: getCheckboxSelectionColumn(),
   code: {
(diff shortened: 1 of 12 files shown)


packages/common/src/ui/layout/tables/utils/ColumnDefinitionSetBuilder.ts
```diff
@@ -65,11 +65,13 @@ const getColumnLookup = <T extends DomainObject>(): Record<
     key: 'name',
     label: 'label.name',
     width: 75,
+    sortable: false,
```
mark-prins, line 68: The item list would be better with the name sortable as well as the code, I think

packages/common/src/ui/layout/tables/utils/ColumnDefinitionSetBuilder.ts
```diff
@@ -65,11 +65,13 @@ const getColumnLookup = <T extends DomainObject>(): Record<
...
   },
   invoiceNumber: {
     key: 'invoiceNumber',
     label: 'label.invoice-number',
     width: 50,
+    sortable: false,
```
mark-prins, line 74: also surprised to see the invoice number and total non-sortable. Are these API limitations that I didn't notice?

packages/invoices/src/OutboundShipment/ListView/api.ts
```diff
@@ -10,6 +10,7 @@ import {
   GraphQLClient,
   getSdk,
   InvoiceSortFieldInput,
+  InvoicesQueryVariables,
```
mark-prins, line 13: oh, nice!

mark-prins (approved):
🎉  Thanks!

packages/common/src/ui/layout/tables/utils/ColumnDefinitionSetBuilder.ts
joshxg: Ah of course! I removed it thinking it was for the invoices list. I've since added the columns as sortable to the api in  #374

packages/common/src/ui/layout/tables/utils/ColumnDefinitionSetBuilder.ts
joshxg: Yea - I've added them to the spreadsheet, and now to our mock server

PR #372, opened 2021-11-04
#366 api context

### LINKED ISSUE #366: Create an API context
Using the new code gen sdk, would be good to create an api object in the host and pass this down with context.

this is cleaning up some of the work that's been done as well as setting us up for when auth comes in

joshxg:
Fixes #366 

- Adds an `OmSupplyApiContext` which passes down a single client for api requests so there is a single place we need to inject the url and sets us up for adding auth details, soon
- Went offroad and factored out some of the code in the outbound `ListView` similarly to the `DetailView`


packages/common/src/api/OmSupplyApiContext.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+interface OmSupplyApiControl {
+  api: OmSupplyApi;
+  client: GraphQLClient;
+  setHeader: (header: string, value: string) => void;
+  setUrl: (url: string) => void;
+}
+
+const OmSupplyApiContext = createContext<OmSupplyApiControl>({
+  ...createOmSupplyApi(''),
```
joshxg, line 24: Hopefully setting this with an empty url initially is fine. I didn't want optionals in the types.. 😁 
 We pass the correct url from `Environment` straight away however, I can see that it might be better to just use that URL here rather than passing it as a prop. 

Though I suspect this to not be a problem, I'm not sure how we're going to 'get' that URL in production and whether using the hardcoded string would work 🤔

packages/common/src/api/OmSupplyApiContext.tsx
```diff
@@ -0,0 +1,68 @@
...
+    api: OmSupplyApi;
+  }>(() => createOmSupplyApi(url));
+
+  const setUrl = useCallback((newUrl: string) => {
+    client.setEndpoint(newUrl);
+  }, []);
+
+  const setHeader = useCallback((key: string, value: string) => {
+    client.setHeader(key, value);
+  }, []);
```
joshxg: Not using these but just set them up as it was fresh in my mind..

packages/host/src/Host.tsx
```diff
@@ -182,106 +183,113 @@ const Host: FC = () => {
...
-                              />
-                              <Route
-                                path={RouteBuilder.create(AppRoute.Messages)
-                                  .addWildCard()
-                                  .build()}
-                                element={
-                                  <Heading locale={locale}>messages</Heading>
-                                }
-                              />
+          <OmSupplyApiProvider url={Environment.API_URL}>
```
joshxg, line 186: this is the only change here 🙄 

I can see passing any other auth details here also if needed

packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
```diff
@@ -25,11 +26,12 @@ import { SidePanel } from './SidePanel';
 
 const useDraftOutbound = () => {
   const { id } = useParams();
+  const { api } = useOmSupplyApi();
 
   const { draft, save, dispatch, state } = useDocument(
     ['invoice', id],
     reducer,
-    OutboundShipmentDetailViewApi
+    getOutboundShipmentDetailViewApi(api)
```
joshxg, line 34: These are changed to pass in the api so that if we can change headers on the single instance etc

packages/invoices/src/OutboundShipment/ListView/AppBarButtons.tsx
```diff
@@ -0,0 +1,49 @@
+import React, { FC } from 'react';
```
joshxg, line 1: extracted out of ListView..

packages/invoices/src/OutboundShipment/ListView/ListView.tsx
```diff
@@ -1,78 +1,27 @@
-import React, { FC, useEffect, useRef, useState } from 'react';
+import React, { FC, useState } from 'react';
```
joshxg, line 1: Took a bunch of stuff out...

Then just passed the omsupply api object into the listview api object..

packages/invoices/src/OutboundShipment/ListView/Toolbar.tsx
```diff
@@ -0,0 +1,53 @@
+import React, { FC, useEffect, useRef } from 'react';
```
joshxg, line 1: extracted out of ListView..

joshxg (commented):
I think it would also be good to put the generated things into the `api` folder

packages/common/src/api/OmSupplyApiContext.tsx
mark-prins: yes, getting the URL in production will be interesting and I'm not sure yet how that might look.. good not to assume that we can source it immediately from `Environment` or load it from a text file

mark-prins (approved):
👍  Wicked! that's much gooder 😄

PR #377, opened 2021-11-05
#328 shift item code

### LINKED ISSUE #328: Shift item related code to `system/item`
- The item related queries and components could be shifted to the `system` package to make them more reusable by other packages

joshxg:
Fixes #328 

*Sigh* - Coin jar for each mess up with the generated file diff?

- I made `ItemSearchInput` in the `system/items` and all the queries are now in there
- I separated the `ItemListView` and general item queries, as we don't need to query for the stock lines for the list view

### DIFF
--- a/packages/system/src/Item/Components/ItemSearchInput/ItemSearchInput.stories.tsx
+++ b/packages/system/src/Item/Components/ItemSearchInput/ItemSearchInput.stories.tsx
@@ -1,0 +1,24 @@
+import React from 'react';
+import { Story } from '@storybook/react';
+import { ItemSearchInput } from './ItemSearchInput';
+import { Item } from '@openmsupply-client/common';
+
+export default {
+  title: 'Item/ItemSearchInput',
+  component: ItemSearchInput,
+};
+
+const Template: Story = () => {
+  const [selectedItem, setSelectedItem] = React.useState<Item | null>(null);
+
+  return (
+    <ItemSearchInput
+      value={selectedItem}
+      onChange={item => {
+        setSelectedItem(item);
+      }}
+    />
+  );
+};
+
+export const Primary = Template.bind({});
--- a/packages/common/src/operations.graphql
+++ b/packages/common/src/operations.graphql
@@ -250,7 +250,7 @@ query names(
   }
 }
 
-query items(
+query itemsWithStockLines(
   $first: Int
   $offset: Int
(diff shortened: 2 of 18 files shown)


packages/system/src/Item/Components/ItemSearchInput/ItemSearchInput.stories.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { Story } from '@storybook/react';
+import { ItemSearchInput } from './ItemSearchInput';
+import { Item } from '@openmsupply-client/common';
+
+export default {
+  title: 'Item/ItemSearchInput',
```
mark-prins, line 7: might be better under `Inputs` - we will end up with a lot of stories I expect, and grouping them nicely is going to be a problem.

mark-prins (approved):
👍  That's much better - like that a lot!

mark-prins:
Couple of afterthoughts
- the sorting on the item table wasn't working properly, in that it doesn't indicate what is currently sorted
- I expect we'll need pagination ( & filtering? ) since we'll be showing a big list of items under 'catalogue'

packages/system/src/Item/Components/ItemSearchInput/ItemSearchInput.stories.tsx
joshxg: Heh I had the same thought and my version of grouping nicely was to put it under the type 🤣 

I do think the story book will likely get massive. I figured if someone was to look for a component they could use to search for items or names, looking under a names or items header. Though at the same time if someone was looking for a component they could copy to create an autocomplete for looking up periods, then they might look at inputs.

I'm not sure, but I can also see `inputs` become massive in itself and needing to be separated out. I'd like to keep it like this for now? It's just changing the string from `Item` to `Inputs` so should be an easy change when it's more obvious you're right? 😁

joshxg:
> I expect we'll need pagination ( & filtering? ) since we'll be showing a big list of items under 'catalogue'

The catalogue list view already has pagination and sorting. Nothing has filtering yet, but yes I think we need to add a search bar to all of the list views.

I also am curious what's going to happen with the customers/item searchinputs/modals fetching all of the records i.e. the item input in the `ItemDetailsModal`. I suspect we'll need to use server side filtering there

Issue #362, opened 2021-11-02
Initial dashboard


mark-prins:
I would like to have something in place which is more indicative than the current skeleton for the dashboard, without going overboard or creating a full implementation.

![Screen Shot 2021-11-03 at 11 00 28 AM](https://user-images.githubusercontent.com/9192912/139957868-e4ef0f6d-0330-4fc2-b896-dd07fe74cc5b.png)

Am basing this off [Dashboard V1](https://zpl.io/VOEAlkk) as I _think_ this was the preferred option ( not certain! )

**Inbound Shipments**
- Show number created today
- Show number created this week ( both could potentially be determined using the list query, but that would be terribly inefficient )

**Outbound Shipments**
- Show number to be picked ( i.e. status === `ALLOCATED` )
- other ideas? The mockup has 'to be picked: This week' but what does that mean?!

**Stock**
- Quantity expired ( quantity? packs? value? )
- Expiring soon ( what is soon? same questions about the unit to use.. quantity, packs, value )

Note: these can be taken inefficiently from the existing API and we'll probably have to do that before the API provides useful information

Unsure what to include under `More Actions` - at this stage perhaps just a link to the list view
I don't believe that we can include any of the 'wizard' buttons right now.. these will require some analysis and then API work

joshxg:
I wonder if it would be better to just make the queries we want for this type of thing in the mock server, then we can just say 'implement this'

mark-prins:
Sounds good to me - I've done the UI, will make some queries

PR #378, opened 2021-11-05
#362 basic dashboard 1

mark-prins:
First part of the quick n easy version of the dashboard
The change was getting too large, so I've had to branch to get a reasonable PR in. This bit sets up the layout and adds some hard-coded values. I've left off the wizard buttons ( we don't have any suggestions for what they should do ) and the 'more actions' button ( likewise, not clear what should be in it ). 

Ends up looking like this:

Screen Shot 2021-11-06 at 9 07 09 AM


package.json
```diff
@@ -66,6 +65,7 @@
   "dependencies": {
     "@types/react-dom": "^17.0.9",
     "graphql": "16.0.0",
+    "lerna": "^4.0.0",
```
mark-prins, line 68: oops. I upgraded to Big Sur, which turned out to be a Big Pain and among other things broke storybook. In the process of fixing I found that lerna wasn't working, so reinstalled. causing this. thought to leave it in, might as well be up to date.

packages/common/src/ui/icons/BarChart.tsx
```diff
@@ -0,0 +1,113 @@
+import React from 'react';
+import SvgIcon, { SvgIconProps } from '@mui/material/SvgIcon';
+
+export const BarChartIcon = ({ ...props }: SvgIconProps): JSX.Element => {
```
mark-prins, line 4: have hard coded colours for this, and I feel bad about that. But Manu's version has three different colours 🤷 
could use secondary / primary + light, main and dark variants - but we've used different versions of the secondary variants already. I think I'll update in the next PR anyway :)

packages/dashboard/src/StatsPanel.tsx
```diff
@@ -0,0 +1,95 @@
...
+    <Grid container alignItems="center" style={{ height: 30 }}>
+      <Grid item>
+        <Typography style={{ fontSize: 24, fontWeight: 'bold' }}>
+          {value}
+        </Typography>
+      </Grid>
+      <Grid
+        item
+        sx={{
+          color: '#8f90a6',
```
mark-prins, line 34: oh, will need to update to use the theme.

mark-prins:
sorry! just committed a bit of an offroad. It bugged me that we had secondary colours which had the dark lighter than the main 🙄  so I've updated.. have also reordered the drop shadows, because I didn't want the new one, which is very slight, to be number 3 when 1 & 2 were darker. I just think it will cause confusion later..

Also trying (and failing) to keep colour names out of theme keys - better, I think, to have roles instead of colours. So have renamed `darkGrey` to `outline` 🤷

packages/common/src/styles/theme.ts
```diff
@@ -122,15 +113,14 @@ const themeOptions = {
     },
   },
   palette: {
-    darkGrey: { main: '#555770' },
+    outline: { main: '#555770' },
     divider: '#eaeaea',
     error: { main: '#e63535' },
-    lightGrey: '#c7c9d9',
-    midGrey: '#8f90a6',
+    gray: { main: '#8f90a6', light: '#c7c9d9', dark: '#555770' },
```
mark-prins, line 119: I just think it is neater to have an object instead of `lightGrey` `midGrey` etc. Would prefer not to have `gray` and `paleGrey` but 🤷

packages/dashboard/src/widgets/InboundShipmentsWidget.tsx
```diff
@@ -0,0 +1,40 @@
...
+          variant="contained"
+          color="secondary"
+          Icon={<PlusCircleIcon />}
+          labelKey="button.new-inbound-shipment"
+          onClick={() => alert('create')}
+        />
+      </Grid>
+    </Grid>
+  </Widget>
+);
```
joshxg, line 40: I wonder if the dashboard widgets would be more at home in their respective package - i.e. `Invoices/Outboundshipment` etc.. though, I can see them loving it here also

packages/dashboard/src/widgets/Widget.tsx
```diff
@@ -0,0 +1,49 @@
...
+        flexDirection: 'column',
+        margin: 11,
+      }}
+    >
+      <Typography sx={{ fontSize: '14px', fontWeight: 'bold' }}>
+        {t(titleKey)}
+      </Typography>
+
+      <React.Suspense fallback={<Loading />}>{children}</React.Suspense>
+    </Paper>
```
joshxg, line 45: This is cool. I can easily see the dashboard remote loading plugin components to display which would be pretty wicked

packages/common/src/ui/icons/BarChart.tsx
joshxg: Also I don't think it's too bad to hard code some colors in specific places

Possibly having the `getPaletteColour` just return a tuple of three colours depending on what colour was passed `primary` or `secondary` (And then throw an error if not primary or secondary saying unsupported 🤔  .. or just support primary..) 🤷

joshxg (approved):
Ooof that's a lot of style updates, thanks it's looking sharp! 🙇  The dashboard is also going to be wicked!

Some things to mention just to make sure for later:

on my laptop screen/tablet the third panel drops off and there's no scroll:
image


And on my desktop monitor they seem spaced oddly. It might be me, but I also can't really see the outline for the panels
image

Was also curious about the AppBar bottom border thing being different to that in the invoice service

packages/common/src/ui/icons/BarChart.tsx
mark-prins: nice, thanks - have done.

mark-prins:
thanks - have added the pale grey, made the toolbar better and fixed for smaller screens.

packages/dashboard/src/widgets/Widget.tsx
mark-prins: would be nice to move the widgets into their functional areas ..and load them up suspensfully 😄

PR #379, opened 2021-11-05
#158 stock line notes

### LINKED ISSUE #158: Outbound: Comment

joshxg:
Fixes #158 

image

- Sorted out the duplicate popper/popover situation - went with a combination of what we both had
- Sorted the localisation files
- Removed the `updateComment` action
- Off-road fix for the hover colors for buttons
- added a hover to the comment column thing


packages/common/src/ui/components/popover/PaperPopover/PaperPopover.tsx
```diff
@@ -0,0 +1,62 @@
...
+}) => {
+  return (
+    <HoverPopover
+      placement={placement}
+      Content={
+        <Paper
+          sx={{
+            width,
+            height,
+            boxShadow: theme => theme.shadows[7],
```
mark-prins, line 32: would you mind adding 
```
            borderRadius: '16px',
```
as well? Manu's designs have the exaggerated round corners, so I think worth putting in at the `Paper` level

mark-prins (approved):
👍  Fantastic! Loving the component inheritance that you're doing.
The popover is looking great, and the shift over to the `Note` column for the detail rows looks very smart!

For me the `BasePopover` story isn't working - but wanted to look at the other PRs.. so haven't investigated. not important, I think

joshxg:
👍 
43050b7

PR #381, opened 2021-11-06
#380 Colour picker in detail panel

### LINKED ISSUE #380: Add colour picker to detail panel
Need to add the colour picker to change the colour of an invoice into the detail panel

joshxg:
Fixes #380 

- Not sure this is indicative enough that you can click it, but this is what I have. I tried playing around with making it look more like a button but couldn't make it look any better so just stuck with this


packages/common/src/ui/layout/tables/columns/NameAndColorColumn.tsx
```diff
@@ -17,55 +17,21 @@ export const getNameAndColorColumn = <T extends DomainObjectWithRequiredFields>(
...
-            e.stopPropagation();
-            handleClick(e);
-          }}
-        />
-        <Box ml={1} />
-        {column.accessor(rowData)}
-      </Box>
-    );
-  },
+  Cell: ({ rowData, column }) => (
```
mark-prins, line 20: I changed the color using this one and got 
```
Error
Unknown type "InvoicePatch". Did you mean "InvoiceNode"?:
```
which it shouldn't be either. I've been meaning to look at mutations.. maybe it is time

mark-prins (approved):
👍  woo! spot on! I think it's clear enough that the colour is clickable.. there's a mouse pointer change and a tooltip. what more do users need?

packages/common/src/ui/layout/tables/columns/NameAndColorColumn.tsx
joshxg: Hmm true. I will tackle the mutations now that I can get the schema from the backend development branch

PR #388, opened 2021-11-07
Add related documents section

### LINKED ISSUE #384: Add related document fields to detail panel
I suppose we just show all 4 of these regardless if they're present, with an `N/A` or something otherwise.

Show the link as an `<a/>` - toast for now?

joshxg:
Fixes #384 

- Just adds the not-fully-implemented related documents section

image


packages/common/src/ui/forms/Panel/PanelField.tsx
```diff
@@ -8,6 +8,7 @@ export const PanelField: React.FC = props => (
     {...props}
     sx={{
       color: theme => theme.palette.form.field,
+      textAlign: 'right',
```
joshxg, line 11: I adjust styles slightly to have the rows space-between and the value text align to the right with the field taking up 2/3rds of the space. Mainly this was to suit my need of having longish field names and because I thought it looked decent. Maybe I shouldn't have touched it though

mark-prins: looks much better!

mark-prins: I think we should update the default `<a>` styles sometime though

mark-prins (approved):
🎉  Sweeet! looks really good 😃

packages/common/src/ui/forms/Panel/PanelField.tsx
joshxg: Sure, what did you have in mind? I just played around with moving the `Link` to have the `PanelField` as the child and that makes it not look like a link, I think

mark-prins: didn't have anything in mind, just don't like the default. next time it comes up I'll do something.

PR #390, opened 2021-11-07
#389 make buttons responsive

### LINKED ISSUE #389: Detail view bottom button text is too long
Having `Save & Confirm Allocation` etc on the button is very long when the screen size is small and the overflow doesn't look nice.

Think having the buttons shrink earlier will be good to solve this more easily

joshxg:
Fixes #389 

- Made the buttons shrink slightly earlier
image
- Went off road and made the status crumb component just a single crumb when the screens too small:
image

which is trying to handle this situation

image

(And as I look at this, I notice we will need to also sort out the normal breadcrumbs for when we add the invoice number)

Not ideal, but more ideal than taking more vertical space by adding even more to the footer

### DIFF
--- a/packages/common/src/hooks/useIsScreen/useIsExtraLargeScreen.ts
+++ b/packages/common/src/hooks/useIsScreen/useIsExtraLargeScreen.ts
@@ -1,7 +1,6 @@
-import { useMediaQuery, useTheme } from '@mui/material';
+import { useTheme, useMediaQuery } from '@mui/material';
 
 export const useIsExtraLargeScreen = (): boolean => {
   const theme = useTheme();
-
   return useMediaQuery(theme.breakpoints.up('xl'));
 };
--- a/packages/common/src/styles/theme.ts
+++ b/packages/common/src/styles/theme.ts
@@ -82,7 +82,7 @@ const themeOptions = {
       sm: 599,
       md: 1025,
       lg: 1441,
-      xl: 1535,
+      xl: 1537,
     },
   },
   direction: 'rtl' as Direction,
--- a/packages/common/src/styles/theme.ts
+++ b/packages/common/src/styles/theme.ts
@@ -79,10 +79,10 @@ const themeOptions = {
   breakpoints: {
     values: {
       xs: 0,
-      sm: 600,
-      md: 1024,
-      lg: 1440,
-      xl: 1536,
+      sm: 599,
+      md: 1025,
+      lg: 1441,
+      xl: 1535,
     },
   },
   direction: 'rtl' as Direction,
--- a/packages/common/src/styles/theme.ts
+++ b/packages/common/src/styles/theme.ts
@@ -79,7 +79,7 @@ const themeOptions = {
   breakpoints: {
     values: {
       xs: 0,
-      sm: 599,
+      sm: 601,
       md: 1025,
       lg: 1441,
       xl: 1537,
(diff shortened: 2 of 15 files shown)


packages/common/src/styles/theme.ts
```diff
@@ -79,10 +79,10 @@ const themeOptions = {
...
     values: {
       xs: 0,
-      sm: 600,
-      md: 1024,
-      lg: 1440,
-      xl: 1536,
+      sm: 599,
+      md: 1025,
+      lg: 1441,
+      xl: 1537,
```
joshxg, line 85: when I use `.down` then the value is exclusive - so `down(1024)` is false and so I've been using `smallScreen` as `1024` and below which was using `down('md')`.. I can see how this  could be confusing so if you have any better ideas 🤷

packages/common/src/ui/components/domain/StatusCrumbs/StatusCrumbs.tsx
```diff
@@ -42,6 +50,30 @@ export const StatusCrumbs = <StatusType extends string>(
...
+    Crumbs = statuses.map(status => {
+      const date = statusLog[status];
+
+      return (
+        <StyledText color={date ? 'secondary' : 'gray.main'} key={status}>
+          {t(statusFormatter(status))}
+        </StyledText>
+      );
+    });
+  }
```
joshxg, line 75: can see that this looks a bit ugly. Though, I think it's OK as a solution 🤷

packages/common/src/hooks/useIsScreen/useIsExtraLargeScreen.ts
```diff
@@ -1,7 +1,6 @@
-import { useMediaQuery, useTheme } from '@mui/material';
+import { useTheme, useMediaQuery } from '@mui/material';
 
 export const useIsExtraLargeScreen = (): boolean => {
   const theme = useTheme();
-
   return useMediaQuery(theme.breakpoints.up('xl'));
```
mark-prins, line 5: this one should use `useIsScreen` as well?

packages/common/src/styles/theme.ts
mark-prins: I'm already confused! why is `sm` `599` when `md` is `1025`?
I think having `down('md')` is fine - it is to me anyway. I justified it with thinking that smaller than medium is small ( and xtra small ).. similarly medium is 1025 and up. All nicely hidden behind hooks anyway, so people won't even need to think about it!

packages/common/src/ui/components/domain/StatusCrumbs/StatusCrumbs.tsx
mark-prins: looks fine to me 🤷

mark-prins (approved):
👍  Good catch! looks way better on a small screen!!

packages/common/src/hooks/useIsScreen/useIsExtraLargeScreen.ts
joshxg: the confusion is at a maximum!

`useIsScreen` is using `down`, so `useIsLargeScreen` is `down, xl` - where as extra large is `up` - anything above xl..

packages/common/src/styles/theme.ts
joshxg: ah yes i've changed it to 601 ..

😕

PR #392, opened 2021-11-08
#327 add filtering

### LINKED ISSUE #327: Add filtering to `useListData`
The backend API is supporting a `filter` object which we haven't been doing at all.

- Implement some sort of filtering mechanism to `useListData`
- Add filtering to the outbound shipments list view & customers

joshxg:
Fixes #327 - or starts

- Added a `useFilterBy` hook to help with filtering. I was really unsure of the api for this hook and the types are still struggling
- Added a small text box in the outbound shipment area for filtering to test this out - will build on that to make a better ui, which i have no idea what it would look like - but will add debouncing etc etc

### DIFF
--- a/packages/common/src/hooks/useFilterBy/useFilterBy.tsx
+++ b/packages/common/src/hooks/useFilterBy/useFilterBy.tsx
@@ -13,7 +13,7 @@ type FilterRule = {
 
 export type FilterBy<T> = Partial<Record<Partial<keyof T>, FilterRule>>;
 
-interface UseFilterByControl<T> {
+export interface FilterState<T> {
   filterBy: FilterBy<T> | null;
 
   onChangeDateFilterRule: (
@@ -33,7 +33,7 @@ interface UseFilterByControl<T> {
 
 export const useFilterBy = <T extends unknown>(
   initialFilterBy?: FilterBy<T> | null
-): UseFilterByControl<T> => {
+): FilterState<T> => {
   const [filterBy, setFilterBy] = useState<FilterBy<T> | null>(
     initialFilterBy ?? null
   );
--- a/packages/common/src/hooks/useFilterBy/useFilterBy.tsx
+++ b/packages/common/src/hooks/useFilterBy/useFilterBy.tsx
@@ -11,55 +11,52 @@ type FilterRule = {
     | FilterByConditionByType['date']]?: unknown;
 };
 
-type FilterBy<PossibleFilterKey extends string> = Record<
-  PossibleFilterKey,
-  FilterRule
->;
+export type FilterBy<T> = Partial<Record<Partial<keyof T>, FilterRule>>;
 
-interface UseFilterByControl<PossibleFilterKey extends string> {
-  filterBy: FilterBy<PossibleFilterKey>;
+interface UseFilterByControl<T> {
+  filterBy: FilterBy<T> | null;
 
   onChangeDateFilterRule: (
-    key: PossibleFilterKey,
+    key: keyof T,
     condition: FilterByConditionByType['date'],
     value: Date
   ) => void;
 
   onChangeStringFilterRule: (
-    key: PossibleFilterKey,
+    key: keyof T,
     condition: FilterByConditionByType['string'],
     value: string
   ) => void;
 
-  onClearFilterRule: (key: PossibleFilterKey) => void;
+  onClearFilterRule: (key: keyof T) => void;
 }
 
-export const useFilterBy = <PossibleFilterKey extends string>(
-  initialFilterBy: FilterBy<PossibleFilterKey>
-): UseFilterByControl<PossibleFilterKey> => {
-  const [filterBy, setFilterBy] =
-    useState<FilterBy<PossibleFilterKey>>(initialFilterBy);
+export const useFilterBy = <T extends unknown>(
+  initialFilterBy?: FilterBy<T> | null
+): UseFilterByControl<T> => {
+  const [filterBy, setFilterBy] = useState<FilterBy<T> | null>(
+    initialFilterBy ?? null
+  );
 
   const onChangeStringFilterRule = (
-    key: PossibleFilterKey,
+    key: keyof T,
     condition: FilterByConditionByType['string'],
     value: string
   ) => {
-    const newFilterRule = { [key]: { [condition]: value } };
-
-    setFilterBy({ ...filterBy, ...newFilterRule });
+    const newFilter = { [key]: { [condition]: value } };
+    setFilterBy({ ...filterBy, ...newFilter });
   };
 
   const onChangeDateFilterRule = (
-    key: PossibleFilterKey,
+    key: keyof T,
     condition: FilterByConditionByType['date'],
     value: Date
   ) => {
-    const newFilterRule = { [key]: { [condition]: value } };
-    setFilterBy({ ...filterBy, ...newFilterRule });
+    const newFilter = { [key]: { [condition]: value } };
+    setFilterBy({ ...filterBy, ...newFilter });
   };
 
-  const onClearFilterRule = (key: PossibleFilterKey) => {
+  const onClearFilterRule = (key: keyof T) => {
     setFilterBy({ ...filterBy, [key]: null });
   };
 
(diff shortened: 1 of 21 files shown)


packages/common/src/hooks/useFilterBy/useFilterBy.tsx
```diff
@@ -0,0 +1,75 @@
+import { useState } from 'react';
+
+interface FilterByConditionByType {
+  string: 'equalTo' | 'like';
+  date: 'beforeOrEqualTo' | 'afterOrEqualTo' | 'equalTo';
```
joshxg, line 5: In general the types here are really hard ..

We need to send an object like:

```
{
comment: { equalTo: value },
otherPartyName: {like: value}
}
```
where there is an implicit AND for the fields

packages/common/src/hooks/useFilterBy/useFilterBy.tsx
```diff
@@ -0,0 +1,75 @@
...
+  };
+
+  const onChangeDateFilterRule = (
+    key: keyof T,
+    condition: FilterByConditionByType['date'],
+    value: Date
+  ) => {
+    const newFilter = { [key]: { [condition]: value } };
+    setFilterBy({ ...filterBy, ...newFilter });
+  };
```
joshxg, line 61: So, I couldn't get a way to us the `typeof` the `key` to index `FilterByConditionByType` so I went with two explicit callbacks for the different types

packages/common/src/hooks/useFilterBy/useFilterBy.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  const onClearFilterRule = (key: keyof T) => {
+    setFilterBy({ ...filterBy, [key]: null });
```
joshxg, line 64: if sending `null` it seems to do nothing in thebackend api. Unsure if intentional

packages/common/src/hooks/useListData/useListData.test.tsx
```diff
@@ -102,7 +102,7 @@ describe('useListData', () => {
   it('calls error boundary on server error', async () => {
     const ErrorTest = () => {
       const { data } = useListData(
-        { key: 'message' },
+        { initialSortBy: { key: 'message' } },
```
joshxg, line 105: just changing the calling signature to be an object `{initialSortBy, initialFilterBy}` rather than just an `initialSortBy`

packages/common/src/hooks/useListData/useListData.ts
```diff
@@ -39,21 +41,31 @@ interface ListDataState<T extends ObjectWithStringKeys> extends QueryParams<T> {
...
 
   const { data, isLoading: isQueryLoading } = useQuery(
     fullQueryKey,
-    api.onRead({ first, offset, sortBy }),
+    api.onRead({
+      first,
+      offset,
+      sortBy,
+      filterBy,
+    }),
```
joshxg, line 68: Adding `filterBy` to `useListData`

packages/common/src/hooks/useQueryParams/useQueryParams.test.tsx
```diff
@@ -68,7 +68,7 @@ describe('useQueryParams', () => {
     window.resizeTo(1000, 1000);
 
     const { result } = renderHook(
-      () => useQueryParams<TestSortBy>({ key: 'id' }),
+      () => useQueryParams<TestSortBy>({ initialSortBy: { key: 'id' } }),
```
joshxg, line 71: just changing the calling signature to be an object `{initialSortBy, initialFilterBy}` rather than just an `initialSortBy`

packages/common/src/hooks/useQueryParams/useQueryParams.ts
```diff
@@ -1,31 +1,50 @@
...
+    sort,
+    pagination,
+  };
 
   return {
     ...pagination,
+    ...sort,
+    ...filter,
+    sort,
+    filter,
```
joshxg, line 46: adding `useFilterBy` to `useQueryParams` (and also returning objects `sort` and `filter` with the respective states - eaiser to pass around 🤷

packages/invoices/src/OutboundShipment/ListView/Toolbar.tsx
```diff
@@ -41,8 +44,28 @@ export const Toolbar: FC<{
...
+      <BasicTextInput
+        value={filterString}
+        onChange={e =>
+          filter.onChangeStringFilterRule(
+            'otherPartyName',
+            'like',
+            e.target.value
+          )
+        }
+      />
```
joshxg, line 68: not sure what to do about the ui here

packages/mock-server/src/api/resolvers.ts
```diff
@@ -141,7 +142,26 @@ export const ResolverService = {
...
+          }
+
+          return true;
+        });
+      }
+
+      return createListResponse(
+        invoices.length,
+        filteredData,
+        'InvoiceConnector'
```
joshxg, line 163: just doing some really basic filtering

packages/invoices/src/OutboundShipment/ListView/Toolbar.tsx
mark-prins: looks ok, fairly subtle and out of the way. too much to ask to have a label on it?

mark-prins:
the pagination ends up a little funky - not sure if this just the MSW implementation or not. I tried switching to the remote server, but that fails ( `Error: Failed to parse "NameSortFieldInput": Enumeration type does not contain value "NAME". (occurred while parsing "NameSortInput")` and `react_devtools_backend.js:2540 Error: Invalid value for argument "sort.key", enumeration type "InvoiceSortFieldInput" does not contain the value "OTHER_PARTY_NAME"` )

When filtering, the page number, number of pages shown and number of pages aren't updated - as below.
The caching is good though - filtering for "CMS" then "CM" and then back to "CMS" is very slick, with the last clearly coming from the cache.

Screen Shot 2021-11-09 at 10 29 19 AM

Screen Shot 2021-11-09 at 10 29 36 AM

mark-prins:
ah, I see. just needs the other PRs merged for the various schema changes. this is a bit tedious.. glad that there's the mock server to work with 😄

joshxg:
Argh I completely forgot to mention, sorry!

I left the pagination because I really have no idea what to do with it!!

When filtering down to 20/100, should it say `showing 1-20 of 20` or still `showing 1-20 of 100`? Or should we be displaying `1-20 of 20 (100 total)` or something.. I dunno!  (Side note: I now hate pagination!!! 🤣 ) 

I think the pages should be brought down to the number to display

packages/common/src/hooks/useFilterBy/useFilterBy.tsx
mark-prins: oh, that's a shame was hoping that we might be able to use `typeof` the `key`. Like what you've done though, and the implementation in toolbar is pretty slick.

mark-prins (approved):
👍  Awesome to see the filtering in action! makes the list view even more usable

mark-prins:
> I think the pages should be brought down to the number to display

It would make sense to me to show values relating to the current ( i.e. filtered ) result set. I found the `Showing 1-20 of 174` weird, when there were only 6 rows in the table. And there were `174` results before filtering and after filtering 🤷 

What would be sweet, is for the API to return the number of results which match the filter, _and_ the total number, so we can say `Showing 1-20 of 53 (174 total)` or maybe enough to say `Showing 1-20 of 53 (filtered)`

packages/common/src/hooks/useFilterBy/useFilterBy.tsx
joshxg: I think the types can be improved, just need to find the time 😁

packages/invoices/src/OutboundShipment/ListView/Toolbar.tsx
joshxg: Not too much - i'll do the pagination fixes and the label in a new PR so I can trim this branches up!

PR #393, opened 2021-11-08
Add updated scehma

joshxg:
Was starting to work on mutations and realised there was a lot of renaming in the new gql schema as well as other updates, so thought best to update first.

What changed:
- CustomerInvoice/SupplierInvoice -> XShipment
- A few extra fields
- batch mutations

There's still a significant number of changes (mainly fields, but a few types at the bottom of the schema file) still missing in the latest backend develop branch which we are using that isn't related to placeholders, service charges, locations or statuses which will still prevent us from integrating. I've marked all changes with a comment `## Additions`.

Unsure whether I should be raising these, putting them in the spreadsheet, asking in telegram do a pr myself or waiting


mark-prins (approved):
👍  Ooof! thanks for sorting that out!

Issue #273, opened 2021-10-18
Add proper saving of an outbound


joshxg:
Currently we're still just sending the whole invoice document to the backend to be saved. This has been changed to be separate create/update/delete mutations for both the invoice and invoice lines.

So, we need to send separate C/U/D mutations for the invoice and each of it's lines rather than just the full document.

We then need to either send these mutations as we perform actions or flag the records and map the records to it's correct mutation.

The flag and map strategy is what we're going for as otherwise we might need rollbacks or some other type of mechanism

For the invoice:
- Whenever the invoice itself is edited flag the invoice as dirty. 
- We won't need to track deleted or created (I think)

For the lines:
- Whenever a line is created, flag the line as created
- Whenever a line is edited, flag the line as dirty
- Whenever a line is deleted, flag the line as deleted

On save:
- Send a mutation for the invoice if it is dirty.
- Iterate through the lines, and map each line which has been marked as deleted into a delete and send them first (so stock is released and can be assigned to any other subsequent lines)
- Iterate through the lines and map each created line to a created mutation 
- Iterate through the lines and map each updated line to an updated mutation

On rendering:
- Filter out any lines which have been marked with delete.

Implementation

Mapping:

I think all of this mapping can go into the `onUpdate` method, so we still just pass the async method the whole document, and the 'api' layer handles this partitioning. This means that 🤞 if the backend is able to do this all for us, nothing will really change except for removing a bunch of code

Flagging:

I'm not really sure where to put all this. I suspect something like this might be OK, having some shapes where the values/data are separate from the ui state/update fns

```
type OutboundShipment {
    dirty: boolean
    values: InvoiceValues
}

type InvoiceItem {
    dirty: boolean;
    created: boolean;
    deleted: boolean;
    updateQuantity: (q: number) => void
    batches: InvoiceLineRow[]
}

type InvoiceLineRow {
    dirty: boolean;
    created: boolean;
    deleted: boolean;
    updateQuantity: (q: number) => void
    values: InvoiceLine
}

```

PR #395, opened 2021-11-08
Add updated invoice creation

joshxg:
Starts #273 

- This adjusts the insert mutation to work with the most recent backend changes. 
- Querying still doesn't work as we have a bunch of sorting fields etc which the current backend doesn't, but you should still be able to make an invoice.
- Not sure why, but the navigation doesn't happen until after the retries of fetching the list of invoices ... 🤔 
- I was just using sqlite and checking the db to make sure it was actually created
image
- Should add some sort of uuid package, I Suppose. Us adding the IDs is to make adding lines in the invoice easier to track
- Probably shouldn't be able to enter invoices with an empty string


mark-prins (approved):
👍  looks ok to me!
having trouble testing.. but I trust you!

PR #396, opened 2021-11-08
Add update to update OS mutation

joshxg:
#273 

- This one updates the update mutation!
- Needed to add `color` to the `updateInput` and still can't query for invoices (doens't have color and missing sorting etc) so still can't test nicely against the real api.


joshxg (commented):
I like how no react component was touched

mark-prins (approved):
👍  impressive!

PR #398, opened 2021-11-08
Add update to invoice deletion mutation

joshxg:
#273 

- This one updates the delete to use the new api


packages/common/src/schema.graphql
```diff
@@ -925,3 +925,12 @@ type LocationNode {
...
 union NameResponse = NodeError | NameNode
+
+input BatchOutboundShipmentInput {
+  insertOutboundShipments: [InsertOutboundShipmentInput!]
+  insertOutboundShipmentLines: [InsertOutboundShipmentLineInput!]
+  updateOutboundShipmentLines: [UpdateOutboundShipmentLineInput!]
+  deleteOutboundShipmentLines: [DeleteOutboundShipmentLineInput!]
+  updateOutboundShipments: [UpdateOutboundShipmentInput!]
+  deleteOutboundShipments: [String!]
+}
```
joshxg, line 936: I did this so I could auto generate a type to use ..

mark-prins (approved):
👍  cool - thanks again! 
am interested how we can interpret any problems and show a meaningful error.. say if one delete in the batch fails or something 🤷 

The mutation worked fine.. all looks ok to me!

joshxg:
Right now I'm thinking that if we throw an error, we can catch the error in the react query `onError` which can dispatch an `updateError` action with what went wrong and we can handle that in the reducer 🤔 

But.. it's really hard to meaningfully sort the errors out without having the api fully integrated

PR #399, opened 2021-11-09
Add detail view to use new invoice updating api

joshxg:
#273 

- The one makes the detail view save buttons work 🎉 
- BUT.. it doesn't save the lines yet :) to do that, we need to tag each line as `updated`, `deleted` or `created` and then in the `onUpdate`, make a `batchOutboundShipmentMutation` and create the correct mutation for each one. This will require a bit more changes, so have just done this and will work on the lines next.


packages/common/src/hooks/useDocument/types.ts
```diff
@@ -10,11 +10,11 @@ export type ReducerCreator<ServerData, State, ActionSet> = (
   dispatch: Dispatch<DocumentActionSet<ActionSet>> | null
 ) => (state: State | undefined, action: ActionSet) => State;
 
-export interface DocumentState<Document, State, ServerData, ActionSet> {
+export interface DocumentState<Document, State, ActionSet> {
   draft: Document;
   state: State;
   dispatch: Dispatch<DocumentActionSet<ActionSet>>;
-  save: (draft: Document) => Promise<ServerData>;
+  save: () => Promise<void>;
```
joshxg, line 17: wasn't using the return and was making the types complain so removed it for now

packages/common/src/hooks/useDocument/useDocument.ts
```diff
@@ -89,6 +89,12 @@ export const useDocument = <
   }, [data]);
 
   const { draft } = state;
+  const draftRef = useRef(draft);
+  draftRef.current = draft;
 
-  return { state, draft, save: mutateAsync, dispatch };
+  const save = async () => {
+    mutateAsync(draftRef.current);
+  };
```
joshxg, line 97: instead of passing the draft in which I thought would be best, instead making a ref so that we can just call `save()` and the most up to date draft is saved (dispatches are async and so was causing me troubles as react waits till the end of an event handler to fire the dispatches, so it can batch them into a single re-render)

packages/common/src/operations.graphql
```diff
@@ -92,6 +92,7 @@ query invoice($id: String!) {
             itemName
             numberOfPacks
             packSize
+            note
```
joshxg, line 95: off-road fix for the mock server note not working correctly

packages/invoices/src/OutboundShipment/DetailView/Footer.tsx
```diff
@@ -110,10 +111,14 @@ export const Footer: FC<OutboundDetailFooterProps> = ({ draft, save }) => {
...
                   color="secondary"
-                  onClick={() => {
+                  onClick={async () => {
                     success('Saved invoice! 🥳 ')();
-                    draft.update?.('status', 'finalised');
-                    save(draft);
+                    await draft.update?.(
+                      'status',
+                      getNextOutboundStatus(draft?.status ?? '') ?? ''
+                    );
```
joshxg, line 119: as above, the `draft.update` isn't triggered until the end of the callback, so awaiting to fire the save afterward.

A little funny behaviour when saving the status, as we're optimistically updating the status, but the crumb is determined by the status dates (i.e. checks if there's a date to know the status which was easier to implement at the time..) so the button updates nicely however the crumb is slightly delayed as it waits for the response from the server with the new status date.

Could fix this by also optimistically updating the date for the status. However I wondered if we'd be putting a spinner up, anyway, in which case it wouldn't matter.

Could also be good to just have the button have a spinner

... Also debouncing probably

packages/common/src/hooks/useDocument/useDocument.ts
mark-prins: wow! nifty

packages/common/src/operations.graphql
mark-prins: thanks - only just noticed this one

packages/invoices/src/OutboundShipment/DetailView/Footer.tsx
mark-prins: actually quite like the status crumb updating just afterwards. seems neat, like something is happening. +1 for adding a spinner to the save & status buttons, and doing that would mean the status update makes more sense too

packages/invoices/src/OutboundShipment/DetailView/Footer.tsx
```diff
@@ -94,7 +95,7 @@ export const Footer: FC<OutboundDetailFooterProps> = ({ draft, save }) => {
                   sx={{ fontSize: '12px' }}
                   onClick={() => {
                     success('Saved invoice! 🥳 ')();
-                    save(draft);
+                    save();
```
mark-prins, line 98: nice to `await save()` and toast afterwards, even though it is too quick to notice when using the mock server 😀

mark-prins (approved):
👍  badass!!

packages/invoices/src/OutboundShipment/DetailView/Footer.tsx
joshxg: ah yes, will do!

PR #400, opened 2021-11-09
#273 mutations 5

joshxg:
Bit big!

- Extremely tedious and frustrating without the real api but this adds inserting of invoice lines using the extracted schema......
- It's 'working' for both the mock server and mock worker (persisting, at least)
- There are a bunch of updates still needed, but thought to PR this for now as it's getting too big
- There are values which when added show in the table as `null`/`undefined` - i.e. for `batch` which 1) likely need values added and 2) need to handle this better in cells, probably
- You can only add a single line at a time, at the moment as the `id` given in `ItemDetailsModal` is always `""`, and so when updating/inserting lines we insert with id `""` and then update id `""`. Will add a `uuid` package to generate these in the next one
- Not handling the response right now - it's too difficult to try and handle this meaningfully without the proper api setup

### DIFF
--- a/packages/common/src/operations.graphql
+++ b/packages/common/src/operations.graphql
@@ -529,3 +529,20 @@ query stockCounts {
     }
   }
 }
+
+mutation upsertOutboundShipment(
+  $insertOutboundShipmentLines: [InsertOutboundShipmentLineInput!]
+  $updateOutboundShipments: [UpdateOutboundShipmentInput!]
+) {
+  batchOutboundShipment(
+    insertOutboundShipmentLines: $insertOutboundShipmentLines
+    updateOutboundShipments: $updateOutboundShipments
+  ) {
+    insertOutboundShipmentLines {
+      id
+    }
+    updateOutboundShipments {
+      id
+    }
+  }
+}
--- a/packages/common/src/operations.graphql
+++ b/packages/common/src/operations.graphql
@@ -538,10 +538,13 @@ mutation upsertOutboundShipment(
     insertOutboundShipmentLines: $insertOutboundShipmentLines
     updateOutboundShipments: $updateOutboundShipments
   ) {
+    __typename
     insertOutboundShipmentLines {
+      __typename
       id
     }
     updateOutboundShipments {
+      __typename
       id
     }
   }
--- a/packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
+++ b/packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
@@ -17,7 +17,7 @@ import { getOutboundShipmentDetailViewApi } from '../../api';
 import { GeneralTab } from './tabs/GeneralTab';
 import { ItemDetailsModal } from './modals/ItemDetailsModal';
 
-import { InvoiceLineRow } from './types';
+import { OutboundShipmentRow } from './types';
 import { Toolbar } from './Toolbar';
 import { isInvoiceEditable } from '../utils';
 import { Footer } from './Footer';
@@ -34,7 +34,7 @@ const useDraftOutbound = () => {
     getOutboundShipmentDetailViewApi(api)
   );
 
-  const onChangeSortBy = (column: Column<InvoiceLineRow>) => {
+  const onChangeSortBy = (column: Column<OutboundShipmentRow>) => {
     dispatch(OutboundAction.onSortBy(column));
   };
 
@@ -48,7 +48,7 @@ export const DetailView: FC = () => {
 
   const columns = useColumns(
     [
-      getNotePopoverColumn<InvoiceLineRow>(),
+      getNotePopoverColumn<OutboundShipmentRow>(),
       'itemCode',
       'itemName',
       'batch',
@@ -59,7 +59,7 @@ export const DetailView: FC = () => {
       'itemUnit',
       'unitQuantity',
       'numberOfPacks',
-      getRowExpandColumn<InvoiceLineRow>(),
+      getRowExpandColumn<OutboundShipmentRow>(),
       GenericColumnKey.Selection,
     ],
     { onChangeSortBy, sortBy },
--- a/packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
+++ b/packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
@@ -3,7 +3,6 @@ import React from 'react';
 import {
   DialogButton,
   Grid,
-  InvoiceLine,
   Item,
   useForm,
   useDialog,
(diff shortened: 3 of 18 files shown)


packages/common/src/operations.graphql
```diff
@@ -529,3 +529,23 @@ query stockCounts {
...
+    __typename
+    insertOutboundShipmentLines {
+      __typename
+      id
+    }
+    updateOutboundShipments {
+      __typename
+      id
+    }
+  }
```
joshxg, line 550: not really working with the response right now, so have just left it like this

packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
```diff
@@ -17,7 +17,7 @@ import { getOutboundShipmentDetailViewApi } from '../../api';
 import { GeneralTab } from './tabs/GeneralTab';
 import { ItemDetailsModal } from './modals/ItemDetailsModal';
 
-import { InvoiceLineRow } from './types';
+import { OutboundShipmentRow } from './types';
```
joshxg, line 20: I changed what we were using as `InvoiceLineRow` to be `OutboundShipmentRow`

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
```diff
@@ -12,28 +11,28 @@ import {
 
 import { BatchesTable } from './BatchesTable';
 import { ItemDetailsForm } from './ItemDetailsForm';
-import { BatchRow } from '../types';
+import { BatchRow, OutboundShipmentRow } from '../types';
 
 interface ItemDetailsModalProps {
-  invoiceLine?: InvoiceLine;
+  invoiceLine?: OutboundShipmentRow;
```
joshxg, line 17: item details modal using `OutboundShipmentRow` over `InvoiceLine` now

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
```diff
@@ -12,28 +11,28 @@ import {
...
 export const getInvoiceLine = (
   id: string,
   item: Item,
   line: { id: string; expiryDate?: string | null },
   quantity: number
-): InvoiceLine => ({
+): OutboundShipmentRow => ({
   id,
-  itemId: '',
+  itemId: item.id,
```
joshxg, line 30: need this for the mutation

packages/invoices/src/OutboundShipment/DetailView/reducer.test.ts
```diff
@@ -2,13 +2,14 @@ import {
   DocumentAction,
   Invoice,
   createColumns,
+  DocumentActionSet,
 } from '@openmsupply-client/common';
 import { placeholderInvoice } from './index';
 import { reducer, OutboundShipmentStateShape, OutboundAction } from './reducer';
-import { InvoiceLineRow } from './types';
+import { OutboundShipmentRow, OutboundShipmentAction } from './types';
```
joshxg, line 9: quite a few test updates to try make them easier to write as I can see quite a few being written!

packages/invoices/src/OutboundShipment/DetailView/reducer.test.ts
```diff
@@ -20,6 +21,11 @@ describe('DetailView reducer', () => {
       costPricePerPack: 0,
       sellPricePerPack: 0,
       itemName: 'a',
+      stockLineId: '',
+      invoiceId: '',
+      isCreated: false,
+      isUpdated: false,
+      isDeleted: false,
```
joshxg, line 28: I've added these `isX` fields to track what mutation to perform for each row. I've used them from the perspective of the client.. as in "isCreated" is `true` when it is created on the client but not persisted on the server, `isDeleted` is true when the row is deleted on the client, but not on the server. I was unsure whether it should be the reverse - i.e. `isCreated` is `true` when it is a persisted row, and when we create one on the client we set that to `false`  😕

packages/invoices/src/OutboundShipment/DetailView/reducer.ts
```diff
@@ -104,20 +104,31 @@ export const reducer = (
...
+            const draftLine = draft.lines.find(
+              line => line.id === serverLine.id
+            );
+
+            if (draftLine) {
+              return mergeLines(serverLine, draftLine);
+            }
+
+            return createLine(serverLine, draft, dispatch);
+          });
```
joshxg, line 121: Now, after fetching server data, iterate over and if there exists a version of that line in the draft, do a 'merge' - this was primarily done now to make the `isCreated` set to `false` after it had been created.

I don't know if we'd ever be able to set `isUpdated` to false without blocking while saving (i.e. a row `isUpdated=true` -> save -> edit same row again -> save success, don't know what to mark as updated) or having another state field 🤔  - was thinking to just send all lines as updates for now... however, without knowing this we can't determine if the invoice has been fully persisted and show some sort of warning when leaving the page if not.

There's also a chance here:
1. Add a bunch of lines
2. Save
3. Add a new line
4. when merging, only takes the server lines, so newly added line is deleted

Think you can just append any lines in the draft which aren't in the server data, though

packages/invoices/src/api.ts
```diff
@@ -114,7 +118,10 @@ export const onRead =
 
     return {
       ...invoice,
-      lines: linesGuard(invoice.lines),
+      lines: linesGuard(invoice.lines).map(line => ({
+        ...line,
+        stockLineId: '',
```
joshxg, line 123: I added `stockLineId` to `InvoiceLine` and it's not on `InvoiceLineNode`.. 🙄  .. but it is on `OutboundShipmentRow` -  sigh! Needs tidying up, just doing what I have to right now 🤣

packages/invoices/src/api.ts
```diff
@@ -134,20 +141,43 @@ const invoiceToInput = (
...
+      if (
+        updateOutboundShipments?.[0]?.__typename ===
+        'UpdateOutboundShipmentResponseWithId'
+      ) {
+        return patch;
+      }
     }
 
-    throw new Error(updateOutboundShipment.error.description);
+    throw new Error(':shrug');
```
joshxg, line 180: Want to wait for the real api to handle everything better...

packages/mock-server/src/worker/handlers.ts
```diff
@@ -25,7 +27,7 @@ const insertInvoice = graphql.mutation(
     const { variables } = request;
     const { id, otherPartyId } = variables;
 
-    const result = Api.MutationService.insert.invoice({
+    const result = MutationService.insert.invoice({
```
joshxg, line 30: had import problems 🙄

mark-prins (approved):
🎉  wooo!! another huge effort! thanks 🙇‍♂️ 

I've read through it a few times now and it looks fine to me! Runs against mocks - haven't quite finished getting the rust API working with mutations.

would be good to get it merged!

PR #401, opened 2021-11-09
#362 dashboard 3

### LINKED ISSUE #362: Initial dashboard

mark-prins:
Fixes #362 this is the last part of the basic dashboard I was wanting to do.
This tidies up the layout, which wasn't flexing enough, moves the Shipment widgets to be with their buddies, and adds a couple of stories.

Will make an issue to create a `Stock` list and move the widget alongside it.


packages/common/src/api/OmSupplyApiContext.tsx
```diff
@@ -60,9 +60,7 @@ export const OmSupplyApiProvider: FC<ApiProviderProps> = ({
   }, [url]);
 
   return (
-    <Provider value={{ api, client, setUrl, setHeader }}>
-      {children}setApi
-    </Provider>
+    <Provider value={{ api, client, setUrl, setHeader }}>{children}</Provider>
```
mark-prins, line 63: looked like an accident, so I removed it 🤷

mark-prins, line 63: looked like an accident, so I removed it 🤷

packages/host/src/Host.tsx
```diff
@@ -197,7 +197,7 @@ const Host: FC = () => {
                           overflow="hidden"
                         >
                           <AppBar />
-                          <Box display="flex" flex={1} overflow="hidden">
+                          <Box display="flex" flex={1} overflow="scroll">
```
mark-prins, line 200: didn't seem to damage the other pages - and now if the widgets flow to another row, the scrollbar shows

packages/dashboard/src/DashboardService.tsx
```diff
@@ -1,10 +1,10 @@
...
 import { Grid } from '@openmsupply-client/common';
+import { StockWidget } from './widgets';
 import {
-  InboundShipmentsWidget,
-  OutboundShipmentsWidget,
-  StockWidget,
-} from './widgets';
+  InboundShipmentWidget,
+  OutboundShipmentWidget,
+} from '@openmsupply-client/invoices';
```
joshxg, line 7: Maybe `lerna add @openmsupply-client/invoices --scope @openmsupply-client/dashbaord` to be eplicit?

packages/common/src/api/OmSupplyApiContext.tsx
joshxg: 🤣

packages/invoices/src/OutboundShipment/Dashboard/Widget.tsx
```diff
@@ -9,13 +9,13 @@ import {
...
+  Widget,
 } from '@openmsupply-client/common';
-import Widget from './Widget';
-import { StatsPanel } from '../StatsPanel';
 import { useNavigate } from 'react-router';
-import { getOutboundShipmentCountQueryFn } from '../api';
+import { getOutboundShipmentCountQueryFn } from './api';
 
-export const OutboundShipmentsWidget: React.FC = () => {
+export const OutboundShipmentWidget: React.FC = () => {
```
joshxg, line 18: Cool! Might actually not be too much work to do the 'mini' dashboards for the top level navigation areas like `/distribution` etc with widgets like these?

joshxg (commented):
It's great!!

I just realised I am not totally sure what the 3 columns are grouping - in stock, out stock, stock in general?

packages/dashboard/src/DashboardService.tsx
mark-prins: great! thanks

packages/invoices/src/OutboundShipment/Dashboard/Widget.tsx
mark-prins: yeah, should be pretty straightforward.. so maybe that idea is back on the table 😁

mark-prins:
> I just realised I am not totally sure what the 3 columns are grouping - in stock, out stock, stock in general?

Up to us, I suppose. They were in the designs, so I ran with them. So far the in stock is just inbound shipments - could be requisitions & purchase orders as well I guess. We don't have the wizard actions yet, so the panels aren't very specific - i.e. tied to just inbound shipments.

PR #402, opened 2021-11-10
#391 breadcrumbs link wrong

### LINKED ISSUE #391: Breadcrumb link incorrect
When on an outbound shipment detail view, clicking on the breadcrumb for the list view takes me to `http://distribution/outbound-shipment` which doesn't work. This is both local and on the demo server

joshxg:
Fixes #391 

- Suspect this was from a bump in the beta version.. can't find the exact notes on why, but gave up trying pretty quick.
- Offroad upgraded out of the beta for react-router-dom.. everything seems to be working fine.
- thought about sticking all the versions in package.json to a specific version rather than `^`'s 🤔


mark-prins:
specific versions would be safer.. might be a good idea to do that.
I also made the mistake of checking `yarn outdated` and will have a look at bumping some packages

mark-prins (approved):
👍  magic! works like a charm!

PR #403, opened 2021-11-10
#397 search label and pagination state

### LINKED ISSUE #397: Add a label to list view search input and correct pagination state
The pagination state isn't currently taking into account the current filtering.

Also we should add a label to the input in the outbound shipment list view. Thinking `Search:`

joshxg:
Fixes #397 

- Adds a search label to the toolbar and updates the server to send the correct total count on filtered data sets.

### DIFF
--- a/packages/common/src/intl/locales/ar.json
+++ b/packages/common/src/intl/locales/ar.json
@@ -105,6 +105,7 @@
   "label.placeholder": "Placeholder",
   "label.purchase-order": "Purchase order",
   "label.requisition": "Requisition",
+  "label.search": "Search",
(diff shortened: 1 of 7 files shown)


mark-prins (approved):
👍  Lovely - thank! that tidies it up nicely

I have a vague concern about how responsive the search is, with debouncing the input and the API call - but have no ideas what to do about that and think that we need to wait until the rusty api is connected in order to evaluate anyway.

So.. all good!

joshxg:
Do you mean that the search is slow at the moment? I agree and think debouncing will clean that up nicely, when it's done here: https://github.com/openmsupply/openmsupply-client/issues/408

PR #407, opened 2021-11-10
#line insertion issues

### LINKED ISSUE #405: Issues with adding new lines
- Using a blank id, so can only add one line at a time.
- Invoice line doesn't have a code
- lines show undefined for batch, locations as not set
- sell price is not carried over
- unit quantity is not calculated (is this meant to be pack size * pack quantity?)
- expiry is not carried over and/or shows `invalid date` when there isn't one

joshxg:
Fixes #405 

- Pretty straight forward one, maybe?

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -29,12 +29,14 @@
     "react-table": "^7.7.0",
     "stylis": "^4.0.10",
     "stylis-plugin-rtl": "^2.1.0",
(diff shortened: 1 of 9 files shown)


mark-prins (approved):
🎉  wicked! 
thanks 😁  that fixes a few things that I was wondering about. ignore any comments on other PRs

PR #411, opened 2021-11-11
#graphql schema update

joshxg:
Another schema update..

### DIFF
--- a/packages/common/src/schema.graphql
+++ b/packages/common/src/schema.graphql
@@ -6,6 +6,22 @@ schema {
 # Directs the executor to query only when the field exists.
 directive @ifdef on FIELD
 
+type AuthToken {
+  # Bearer token
+  token: String!
+}
+
+# Generic Error Wrapper
+type AuthTokenError {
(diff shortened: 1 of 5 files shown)


packages/common/src/schema.graphql
```diff
@@ -318,31 +341,44 @@ type InsertOutboundShipmentResponseWithId {
...
-  type: SimpleStringFilterInput
-  status: SimpleStringFilterInput
+  invoiceNumber: EqualFilterNumberInput
+  nameId: EqualFilterStringInput
+  storeId: EqualFilterStringInput
+  type: EqualFilterInvoiceTypeInput
+  status: EqualFilterInvoiceStatusInput
   comment: SimpleStringFilterInput
-  theirReference: SimpleStringFilterInput
+  theirReference: EqualFilterStringInput
```
joshxg, line 375: I reverted this change - don't really need to change it, but I do think it's weird

mark-prins (approved):
🎉  awesome! looks great, and happily makes my PR redundant. I think we should leave the dateTime field changes out for now

PR #413, opened 2021-11-11
update schema

### LINKED ISSUE #412: Schema updates
When trying to run against the rust API a couple of errors come up.
Would be good to align them again.

mark-prins:
Fixes #412 
And maybe this is all a terrible idea. Originally I just wanted to update the sort key that was throwing an error when using the rust API. and then I went and updated the `Datetime`s too. probably shouldn't have done that.

Anyway.. this works with the API_URL of `https://demo-open.msupply.org:8000/graphql` which is running the [front-end-fun](https://github.com/openmsupply/remote-server/tree/front-end-fun) branch of the API. That has some of the additional fields we are using, and so the site doesn't crash too much.


joshxg (approved):
Cool! I have a couple PRs with changes in similar files so our merging will be fun 😁 

Actually, It might end up being easier to make the changes to the date times in https://github.com/openmsupply/openmsupply-client/pull/411/files

mark-prins:
not required! 😁

PR #415, opened 2021-11-12
Add basic implementation of summary item idea

joshxg:
just a draft to get feedback


packages/invoices/src/OutboundShipment/DetailView/reducer.ts
```diff
@@ -64,17 +70,65 @@ export const OutboundAction = {
...
+      : [line];
+    acc[itemId] = lines;
+    return acc;
+  }, {} as InvoiceLinesByItemId);
+
+  return Object.keys(byItem).map(itemId =>
+    createItem(itemId, byItem[itemId] as OutboundShipmentRow[])
+  );
+};
+
```
joshxg, line 131: essentially this is the nuts and bolts...

rather than having the table render `OutboundShipmentRow` from `draft.lines`, have this array mapped to an array of `OutboundShipmentItem`s, which is essentially `{itemId: string, batches: OutboundShipmentRow[]}`

Then, the table can render each item. Each row is based an item, and the expando can take `rowData.batches` and then when we click to edit, we can set the `editting line` to be this item, which passes all the invoice lines for that item.

changes a bit of stuff, so wanted to try and get feedback before I went on with it..

joshxg: If we need to toggle between per batch and summary lines, then we can switch between rendering an array of `OutboundShipmentItem` or `OutboundShipmentRow`

joshxg: Though, I would want to avoid having duplicated copies in state, so maybe using ids to reference objects, or something

packages/common/src/hooks/useSortBy/useSortBy.ts
```diff
@@ -1,8 +1,9 @@
 import { ObjectWithStringKeys } from './../../types/utility';
 import { useCallback, useState } from 'react';
+import { GenericColumnKey } from '../..';
 
 export interface SortRule<T extends ObjectWithStringKeys> {
-  key: keyof T;
+  key: keyof T | GenericColumnKey;
```
mark-prins, line 6: sorting by the generic columns? I expect I'll understand why soon enough

packages/common/src/utils/arrays/random.ts
```diff
@@ -0,0 +1,13 @@
+export const ifTheSameElseDefault = <
```
mark-prins, line 1: random? is this a way of creating a `utils` file without calling it that?
love the `ifTheSameElseDefault`! nice use of typescript too 😁

packages/common/src/utils/arrays/reducers.ts
```diff
@@ -0,0 +1,15 @@
+export const getUnitQuantity = (
+  sum: number,
+  someEntity: {
+    numberOfPacks: number;
+    packSize: number;
+  }
+): number => {
+  return sum + someEntity.numberOfPacks * someEntity.packSize;
```
mark-prins, line 8: I get the feeling that this is a calculation we have repeated all over the place in the current codebase 😉

packages/invoices/src/OutboundShipment/DetailView/reducer.ts
```diff
@@ -64,17 +70,65 @@ export const OutboundAction = {
...
+type InvoiceLinesByItemId = Record<string, OutboundShipmentRow[]>;
+
+const createItem = (
+  itemId: string,
+  batches: OutboundShipmentRow[]
+): OutboundShipmentSummaryItem => {
+  const item: OutboundShipmentSummaryItem = {
+    id: itemId,
+    itemId: itemId,
+    itemName: ifTheSameElseDefault(batches, 'itemName', ''),
```
mark-prins, line 95: 😁  very pleasing!

packages/invoices/src/OutboundShipment/DetailView/tabs/GeneralTab.tsx
```diff
@@ -28,22 +28,18 @@ const Expand: FC = () => {
...
-export const GeneralTabComponent: FC<GeneralTabProps<OutboundShipmentRow>> = ({
-  data,
-  columns,
-}) => {
+export const GeneralTabComponent: FC<
+  GeneralTabProps<OutboundShipmentSummaryItem>
+> = ({ data, columns }) => {
   const { pagination } = usePagination();
-  const activeRows = data.filter(({ isDeleted }) => !isDeleted);
+  // const activeRows = data.filter(({ isDeleted }) => !isDeleted);
```
mark-prins, line 35: this is looking like a placeholder, perhaps deleted lines are removed from `items`. Will wait and see!

mark-prins (approved):
😮  wow! that's genius! like where you're heading with it, the approach looks really good

packages/common/src/hooks/useSortBy/useSortBy.ts
joshxg: uhh yeah, sorry. I wanted to push the branch up and this was making TS happy.. will fix!

packages/common/src/utils/arrays/random.ts
joshxg: 🤣 

Honestly, I don't know where to put it 😢

packages/invoices/src/OutboundShipment/DetailView/tabs/GeneralTab.tsx
joshxg: Was just rushing. 

We need to hide the summary rows when all batches are deleted, then the expando needs to hide batches when they're deleted, also.

Removing the item seems to make sense, though that would mean we put the lines elsewhere. Can go back to the `deletedLines` array, I suppose, but then the flag is redundant.

Could add an `isDeleted` to the summary item and do the same thing as we were doing. Seems to make the most sense to me 🤔

PR #417, opened 2021-11-12
#334 nav title length

### LINKED ISSUE #334: Nav menu title length
When a nav menu item length is too long we should do something better then the default html wrapping which is looking funky on my screen

mark-prins:
Fixes #334 

truncated the nav item text and added a title. thought that would look a little more elegant than adding a Mui Tooltip as well.

Screen Shot 2021-11-12 at 5 23 54 PM (2)

and then.. well. I happened to be reworking the rust api schema and should've created a new issue. but thought it was sillier to go full


mark-prins:
now working with the [front-end-fun](https://github.com/openmsupply/remote-server/tree/front-end-fun) branch of the server. 

That branch is running [on the demo server](https://demo-open.msupply.org:8000/graphql) so we don't need to run it locally

joshxg (approved):
Nice! 

Off roads are all good on my opinion 🤣

joshxg:
Going to merge this one to avoid any future conflicts!

Issue #293, opened 2021-10-20
Add a row click to the detail view


joshxg:
Need to add a row click to the detail view to open the edit item modal

joshxg:
@mark-prins have you started this one? I'm just working on the mutations for invoice lines and am doing a bunch of work in the area, so if you haven't, I'll steal it?

mark-prins:
steal away.. haven't started yet

PR #418, opened 2021-11-13
#293 editable rows

joshxg:
Working on #293 ..

- So, we need to edit a row which means we need to set that row as the 'selected row' at some higher level than the modal. This PR lifts the state of the 'selectedItem' in the modal to be at the `DetailView` level. Which means that the `DetailView` passes a callback to change the selected item, also.
- In addition, instead of passing an `InvoiceLine`, we need to have the modal work on all `InvoiceLine`s of an item within the invoice, so have adjusted the modal to work with a `SummaryItem`.

With these changes, a couple of challenges:
- When passing the `SummaryItem`, we have access to the invoice lines, but not the `StockLine`s (In addition, we have the item name, code and ID, but not the actual item - however, it's not needed). I thought of a few solutions: Put the stock lines into the summary item! Bit of an easy one for this, but I feared the stock lines will become stale, so wanted to avoid caching them. Instead, since the item details modal currently has a list of all items and stock lines, we could just use the itemID to search for the stock lines! I was going to go for this, however it has been a concern of mine that we are querying for all items and all stock lines, so I decided to take the long way round..
- Request for the stock lines! Using the `itemCode` of the summary item (As there is no filter right now for an itemID, nor a query for stock lines) which is passed, query for all the stock lines. This means we can remove querying for the stock lines in the `items` query (not done yet) and also add filtering/pagination to the list and not have to rely on all items and stock lines being queried for.
- What this means is that when we open the modal or select an item, the stock lines will need to be fetched, so the `BatchesTable` won't always auto-load. If this is problem in practise, there's some ways we can fix this problem. For example, we can fire the query for the stock lines when hovering a row to edit, or pre-fetch stock lines on invoice load. Similarly for selecting an item - we can prefetch as we filter if we really want to fix the latency. It will still end up as an overall decrease in data sent (and avoid the situation where there's 5000 items and a few stock lines per?).

### DIFF
--- a/packages/common/src/operations.graphql
+++ b/packages/common/src/operations.graphql
@@ -261,10 +261,12 @@ query itemsWithStockLines(
   $offset: Int
   $key: ItemSortFieldInput!
   $desc: Boolean
+  $filter: ItemFilterInput
 ) {
   items(
     page: { first: $first, offset: $offset }
     sort: { key: $key, desc: $desc }
+    filter: $filter
   ) {
     ... on ConnectorError {
       __typename
@@ -351,10 +353,12 @@ query itemsListView(
   $offset: Int
   $key: ItemSortFieldInput!
   $desc: Boolean
+  $filter: ItemFilterInput
 ) {
   items(
     page: { first: $first, offset: $offset }
     sort: { key: $key, desc: $desc }
+    filter: $filter
   ) {
     ... on ConnectorError {
       __typename
--- a/packages/common/src/types/index.ts
+++ b/packages/common/src/types/index.ts
@@ -16,6 +16,15 @@ export interface Name extends DomainObject {
   isSupplier: boolean;
 }
 
+export interface ItemRow extends DomainObject {
+  id: string;
+  isVisible: boolean;
+  code: string;
+  name: string;
+  availableQuantity: number;
+  unitName?: string;
+}
+
 export interface Item extends DomainObject {
   id: string;
   isVisible: boolean;
--- a/packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
+++ b/packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
@@ -11,8 +11,9 @@ import {
   getNotePopoverColumn,
   getRowExpandColumn,
   useOmSupplyApi,
+  Item,
 } from '@openmsupply-client/common';
-import { reducer, OutboundAction } from './reducer';
+import { reducer, OutboundAction, itemToSummaryItem } from './reducer';
 import { getOutboundShipmentDetailViewApi } from '../../api';
 import { GeneralTab } from './tabs/GeneralTab';
 import { ItemDetailsModal } from './modals/ItemDetailsModal';
@@ -44,7 +45,14 @@ const useDraftOutbound = () => {
 export const DetailView: FC = () => {
   const { draft, onChangeSortBy, save, sortBy } = useDraftOutbound();
 
+  const [selectedItem, setSelectedItem] =
+    React.useState<OutboundShipmentSummaryItem | null>(null);
   const itemModalControl = useToggle();
+
+  const onRowClick = (item: OutboundShipmentSummaryItem) => {
+    setSelectedItem(item);
+    itemModalControl.toggle();
+  };
 
   const columns = useColumns(
     [
@@ -66,6 +74,23 @@ export const DetailView: FC = () => {
     [sortBy]
   );
 
+  const onChangeSelectedItem = (newItem: Item | null) => {
+    if (!newItem) return setSelectedItem(newItem);
+
+    // Try and find the outbound summary row that matches the new item
+    const item = draft.items.find(
+      summaryItem => summaryItem.itemId === newItem.id
+    );
+
+    // If we found it, set the selected item.
+    if (item) {
+      setSelectedItem(item);
+    } else {
+      // otherwise, set the selected item to a newly created summary row.
+      setSelectedItem(itemToSummaryItem(newItem));
+    }
+  };
+
   return draft ? (
     <TableProvider createStore={createTableStore}>
       <AppBarButtons
@@ -74,14 +99,20 @@ export const DetailView: FC = () => {
       />
 
       <ItemDetailsModal
+        summaryItem={selectedItem}
         isOpen={itemModalControl.isOn}
         onClose={itemModalControl.toggleOff}
+        onChangeItem={onChangeSelectedItem}
         upsertInvoiceLine={line => draft.upsertLine?.(line)}
       />
 
       <Toolbar draft={draft} />
 
-      <GeneralTab columns={columns} data={draft.items} />
+      <GeneralTab
+        columns={columns}
+        data={draft.items}
+        onRowClick={onRowClick}
+      />
 
       <Footer draft={draft} save={save} />
       <SidePanel draft={draft} />
--- a/packages/invoices/src/OutboundShipment/DetailView/modals/BatchesTable.tsx
+++ b/packages/invoices/src/OutboundShipment/DetailView/modals/BatchesTable.tsx
@@ -7,7 +7,6 @@ import {
   InfoIcon,
   isAlmostExpired,
   ModalNumericInput,
-  Item,
   Table,
   TableBody,
   TableCell,
@@ -26,7 +25,6 @@ import {
 import { BatchRow } from '../types';
 
 export interface BatchesTableProps {
-  item: Item | null;
   onChange: (key: string, value: number, packSize: number) => void;
   register: UseFormRegister<FieldValues>;
   rows: BatchRow[];
@@ -155,13 +153,10 @@ const BasicCell: React.FC<TableCellProps> = ({ sx, ...props }) => (
 );
 
 export const BatchesTable: React.FC<BatchesTableProps> = ({
-  item,
   onChange,
   register,
   rows,
 }) => {
-  if (!item) return null;
-
   const t = useTranslation();
   const onChangeValue: React.ChangeEventHandler<HTMLInputElement> = event =>
     onChange('placeholder', Number(event.target.value), 1);
--- a/packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsForm.tsx
+++ b/packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsForm.tsx
@@ -127,7 +127,11 @@ export const ItemDetailsForm: React.FC<ItemDetailsFormProps> = ({
         {quantityDescription}
         <Grid style={{ display: 'flex' }} justifyContent="flex-end" flex={1}>
           <ModalLabel labelKey="label.unit" justifyContent="flex-end" />
-          <ModalInput disabled width={150} />
+          <ModalInput
+            disabled
+            width={150}
+            defaultValue={summaryItem?.itemUnit ?? ''}
+          />
         </Grid>
       </ModalRow>
     </Grid>
--- a/packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsForm.tsx
+++ b/packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsForm.tsx
@@ -17,13 +17,14 @@ import {
   ModalNumericInput,
 } from '@openmsupply-client/common';
 import { ItemSearchInput } from '@openmsupply-client/system/src/Item';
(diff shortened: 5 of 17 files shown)


packages/common/src/operations.graphql
```diff
@@ -351,10 +353,12 @@ query itemsListView(
   $offset: Int
   $key: ItemSortFieldInput!
   $desc: Boolean
+  $filter: ItemFilterInput
 ) {
   items(
     page: { first: $first, offset: $offset }
     sort: { key: $key, desc: $desc }
+    filter: $filter
```
joshxg, line 361: adding filtering to the items queries

packages/common/src/types/index.ts
```diff
@@ -16,6 +16,15 @@ export interface Name extends DomainObject {
...
 }
 
+export interface ItemRow extends DomainObject {
+  id: string;
+  isVisible: boolean;
+  code: string;
+  name: string;
+  availableQuantity: number;
+  unitName?: string;
+}
```
joshxg, line 26: I wanted to use this in this PR, but the changes needed were a bit much. The idea being similar to others - a specific type for the 'row' variant which contains less things.. (this one, stock lines. I think it's better to have a type where there is no stock lines field at all, rather than a misleading, always empty array?)

packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
```diff
@@ -66,6 +74,23 @@ export const DetailView: FC = () => {
...
+    );
+
+    // If we found it, set the selected item.
+    if (item) {
+      setSelectedItem(item);
+    } else {
+      // otherwise, set the selected item to a newly created summary row.
+      setSelectedItem(itemToSummaryItem(newItem));
+    }
+  };
```
joshxg, line 92: this is the callback for the modal. It will be simplified when we get to filtering out existing items in the invoice out of the items list

packages/invoices/src/OutboundShipment/DetailView/modals/BatchesTable.tsx
```diff
@@ -155,13 +153,10 @@ const BasicCell: React.FC<TableCellProps> = ({ sx, ...props }) => (
 );
 
 export const BatchesTable: React.FC<BatchesTableProps> = ({
-  item,
   onChange,
   register,
   rows,
 }) => {
-  if (!item) return null;
-
```
joshxg, line 164: this was all it was used for, and now there's an `isLoading` state for fetching the batches which i'm using to conditionally render in the parent

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsForm.tsx
```diff
@@ -65,12 +65,6 @@ export const ItemDetailsForm: React.FC<ItemDetailsFormProps> = ({
     setPackSize(newPackSize);
   };
 
-  selectedItem?.availableBatches.forEach(batch => {
-    if (packSizes.every(pack => pack !== batch.packSize)) {
-      packSizes.push(batch.packSize);
-    }
-  });
```
joshxg, line 72: A regression, for now - have tried to minimise them and also make small enough changes to PR heh. We have the stock lines in the parent we can use but the summary item doesn't have the stock lines. So thinking to pass an array of pack sizes

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsForm.tsx
```diff
@@ -85,22 +79,20 @@ export const ItemDetailsForm: React.FC<ItemDetailsFormProps> = ({
       <ModalRow>
         <ModalLabel labelKey="label.item" />
         <Grid item flex={1}>
-          <ItemSearchInput value={selectedItem} onChange={onChangeItem} />
+          <ItemSearchInput
+            currentItemId={summaryItem?.itemId}
+            onChange={onChangeItem}
+          />
```
joshxg, line 85: have added an optional `itemId` to the item search input instead of passing an item if wanted

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsForm.tsx
```diff
@@ -85,22 +79,20 @@ export const ItemDetailsForm: React.FC<ItemDetailsFormProps> = ({
...
       </ModalRow>
       <ModalRow>
         <ModalLabel labelKey="label.available" />
-        <NumericTextInput
-          defaultValue={invoiceLine?.itemCode}
-          inputProps={register('availableQuantity')}
-          disabled
-          sx={{ width: 85 }}
-        />
+        <NumericTextInput value={0} disabled sx={{ width: 85 }} />
```
joshxg, line 90: there was a couple of these I was a little confused about. Seems this is just a disabled text input so I was unsure why it needed to be registered. Seems the `setValue` isn't really needed as we pass a different item the component rerenders.

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsForm.tsx
```diff
@@ -114,7 +106,7 @@ export const ItemDetailsForm: React.FC<ItemDetailsFormProps> = ({
             pattern: { value: /^[0-9]+$/, message: t('error.invalid-value') },
             onChange: event => onChangeQuantity(Number(event.target.value)),
           })}
-          defaultValue={invoiceLine?.numberOfPacks}
+          defaultValue={summaryItem?.numberOfPacks || 0}
```
joshxg, line 109: Less of a regression and more of a TODO!

This is the quantity which needs to be changes for editing an invoice item, so not the invoiceLine.numberOPacks, anyway.

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
```diff
@@ -68,56 +75,88 @@ const sortByDisabledThenExpiryDate = (a: BatchRow, b: BatchRow) => {
...
+    const batchRow = batchRows[batchRowIdx];
+    if (!batchRow) return;
+
+    batchRows[batchRowIdx] = { ...batchRow, quantity };
+    setBatchRows([...batchRows]);
+  };
+
+  return { batchRows, onIssue, isLoading };
+};
+
```
joshxg, line 116: maybe a bad name..

Just querying for the stock lines and then doing the transform into batch rows here.

Need to make a change to take in to account the invoice lines which already exist in the invoice when setting the quantity.

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
```diff
@@ -68,56 +75,88 @@ const sortByDisabledThenExpiryDate = (a: BatchRow, b: BatchRow) => {
...
-  const [allocated, setAllocated] = React.useState(0);
-  const [packSize, setPackSize] = React.useState(1);
+  const [batchRows, setBatchRows] = useState<BatchRow[]>([]);
+
+  const [quantity, setQuantity] = useState(0);
+  const [allocated, setAllocated] = useState(0);
+  const [packSize, setPackSize] = useState(1);
+
+  const methods = useForm({ mode: 'onBlur' });
+  const { reset, register, setValue, getValues } = methods;
```
joshxg, line 132: So, `selectedItem` I lifted up the the `DetailView` and made an `OutboundSummaryItem` and `batchRows` lifted out to `useStockLines`.

packages/invoices/src/OutboundShipment/DetailView/reducer.ts
```diff
@@ -85,9 +86,22 @@ export interface OutboundShipmentStateShape {
...
+    id: item.id,
+    itemId: item.id,
+    itemName: item.name,
+    itemCode: item.code,
+    itemUnit: item.unitName,
+    batches: [],
+    unitQuantity: 0,
+    numberOfPacks: 0,
+  };
+};
```
joshxg, line 100: this is for adding a new item to the invoice - an empty summary item. I was trying to think of a nice way to do this, maybe making pseudo-`OutboundShipmentRow`s from the stock lines instead of using `batchRow` in state, or something like that 🤷

packages/mock-server/src/api/resolvers.ts
```diff
@@ -158,28 +158,53 @@ export const ResolverService = {
...
-      const data = paged.map(item => {
-        return ResolverService.byId.item(item.id);
-      });
+      if (key) {
+        const sortData = getDataSorter(getItemsSortKey(key), !!desc);
+        paged.sort(sortData);
+      }
 
-      return createListResponse(items.length, data, 'ItemConnector');
+      return createListResponse(filtered.length, paged, 'ItemConnector');
```
joshxg, line 207: just adding mock server filtering for items..

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsForm.tsx
joshxg: Ah, also it's being set to just `0` which is likely not ideal!

This will change to be the sum of the pack size selected?

packages/invoices/src/OutboundShipment/DetailView/reducer.ts
mark-prins: I do like the idea of lifting the `batchRow` out, if I understand that correctly

mark-prins (approved):
👍  wow! there's a lot going on in there 😁  looks great to me.. like where you're heading with this

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsForm.tsx
mark-prins: it was `item.availableQuantity` and I think the value should just show that, without doing anything fancy with pack size - the API can do the calculating.
I suppose the availableQuantity must be unit quantity, right?

joshxg: Heh. Just looking at this now.

- Available quantity is something we added and isn't in the 'real' api.
- We also might need to calculate this on the client, anyway. 
    - If the server sends an available quantity of '100', then we go and delete the line for the item, then go and add that item *again*, without having saved the delete, should we add the deleted line packs into the available quantity?
- I thought we had said to change the available quantity based on the pack size selected? i.e. if the total unit quantity is 1000, spread over 5 packs of 100 and 10 packs of 50, then when 50 is selected the available is 10 (and is the maximum which can be entered into the pack quantity input)

What I was heading toward:

```
So, say there are stock lines for a pack size of 2 (A), a pack size of 10 (B) and two for a pack size of 5 (C,D). 

Then, if the user selects a pack size of 5, the modal is essentially pretending the other stock lines don’t exist? So, the available quantity is equal to [C]numberOPacks * 5 + [D]numberOPacks * 5 (and maybe grey out the other stock lines?)

Then if the user changes the pack size, to 10, then everything resets, and only [B] matters, with the available quantity being [B]numberOfPacks*10, and only [B] having stock allocatable
```

mark-prins: oh, yeah. I see.
Side thought - we should label that 'available number of packs' rather than 'available quantity'

you're right about deleting existing allocation, and how this affects available quantity. Makes sense to sum the available on the client for that alone - also helps in implementing pack size selection by the user.

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
```diff
@@ -243,16 +271,17 @@ export const ItemDetailsModal: React.FC<ItemDetailsModalProps> = ({
...
               packSize={packSize}
               setPackSize={setPackSize}
             />
-            <BatchesTable
-              item={selectedItem}
-              onChange={onChangeRowQuantity}
-              register={register}
-              rows={batchRows}
-            />
+            {!isLoading && (
```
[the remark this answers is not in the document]
joshxg: yup! Also some more static dimensions maybe to reduce the moving around of the modal when things are stretchin

packages/invoices/src/OutboundShipment/DetailView/reducer.ts
joshxg: I'm not sure if we are on the same page.

Wasn't able to use outbound shipment rows as we needed the available quantity etc from the stock line

PR #419, opened 2021-11-13
#293 edit rows 2

joshxg:
#293 

- How did I get here?
- Where am I?
- Ah.. thats right..
- Turns out when I open the edit modal, the search input was having a bit of a meltdown because the value I was trying to give it (the item I selected) sometimes wasn't in the selection of items (as there was a limit of 20 items being requested from the server), so I updated the search input to filter items by code (no ID filter) if there was a code or item passed as an initial value. then there was an issue because I also added to filter/request items based on the typed values. But when I was assigning the default item, the name of the item triggered the `onInputChanged`, which triggered the filter, which was filtering on code (as we can currently only filter with ANDs, not ORs), so made some changes so that typed values are filtering by name.
- I also did some factoring out of things and made a `useItem`, `useStockLines` hooks because I am obsessed (`useStockLines` implementation likely to be changed once we have a stock lines query? Same goes for `useItem` when we have a by ID query)


packages/system/src/Item/Components/ItemSearchInput/ItemSearchInput.tsx
```diff
@@ -28,41 +28,61 @@ const renderOption = (
...
+  }, [filter]);
+
   const value =
     currentItem ??
-    (currentItemId
-      ? data?.nodes?.find(i => i.id === currentItemId)
-      : undefined);
+    (currentItemCode
+      ? data?.nodes?.find(i => i.code === currentItemCode) || null
+      : null);
```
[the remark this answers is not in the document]
mark-prins: oh. yes. that last comment - the item doesn't need to be in the list really. Though the autocomplete should look like it has the selected item, and the autocomplete should still function correctly; so it would make sense to do the dance that you have, and make it all work.

packages/system/src/Item/Components/ItemSearchInput/ItemSearchInput.tsx
```diff
@@ -28,41 +28,63 @@ const renderOption = (
...
 
   return (
     <Autocomplete
       filterOptionConfig={filterOptions}
       loading={isLoading}
+      value={value ? { ...value, label: value.name ?? '' } : null}
       noOptionsText={t('error.no-items')}
+      onInputChange={(_, value) => {
+        setFilter({ searchTerm: value, field: 'name' });
```
mark-prins, line 81: I expect that we'll need to filter by name and code; perhaps when the api supports `item.id` that might be possible?

packages/system/src/Item/Components/ItemSearchInput/ItemSearchInput.tsx
```diff
@@ -28,41 +28,63 @@ const renderOption = (
...
+      }}
       onChange={(_, item) => {
-        item && onChange(item);
+        onChange(item);
```
mark-prins, line 84: no, I don't remember why an item could be null. bit of a concern, that one. I mean, how can that be a thing?!

mark-prins (approved):
👍  looks great!

packages/system/src/Item/Components/ItemSearchInput/ItemSearchInput.tsx
joshxg: Hmm yeah. I suspect that filtering by `OR` will take awhile..

PR #420, opened 2021-11-14
#293 edit rows 3

joshxg:
Ugh.. I am unsure if this is all worthwhile at this point :)

- this updates the reducer and tests to work with the summary items rather than just rows
- Still need to fix up the actual item edit modal to work properly again!

### DIFF
--- a/packages/common/src/utils/arrays/utils.ts
+++ b/packages/common/src/utils/arrays/utils.ts
@@ -11,3 +11,12 @@ export const ifTheSameElseDefault = <
   const allTheSame = someEntities.every(entity => entity[key] === value);
   return allTheSame && value ? value : defaultValue;
 };
+
+export const arrayToRecord = <T extends { id: string }>(
+  array: T[]
+): Record<string, T> => {
+  return array.reduce((acc, value) => {
+    acc[value.id] = value;
+    return acc;
+  }, {} as Record<string, T>);
+};
--- a/packages/invoices/src/OutboundShipment/DetailView/Toolbar.tsx
+++ b/packages/invoices/src/OutboundShipment/DetailView/Toolbar.tsx
@@ -13,7 +13,7 @@ import {
   useTableStore,
 } from '@openmsupply-client/common';
 import { NameSearchInput } from '@openmsupply-client/system/src/Name';
-import { OutboundShipmentRow, OutboundShipment } from './types';
+import { OutboundShipment, OutboundShipmentSummaryItem } from './types';
 import { isInvoiceEditable } from '../utils';
 
 interface ToolbarProps {
@@ -27,13 +27,15 @@ export const Toolbar: FC<ToolbarProps> = ({ draft }) => {
   const { selectedRows } = useTableStore(state => ({
     selectedRows: Object.keys(state.rowState)
       .filter(id => state.rowState[id]?.isSelected)
-      .map(selectedId => draft.lines.find(({ id }) => selectedId === id))
-      .filter(Boolean) as OutboundShipmentRow[],
+      .map(selectedId => draft.items.find(({ id }) => selectedId === id))
+      .filter(Boolean) as OutboundShipmentSummaryItem[],
   }));
 
   const deleteAction = () => {
     if (selectedRows && selectedRows?.length > 0) {
-      selectedRows.forEach(line => draft.deleteLine?.(line));
+      selectedRows.forEach(item =>
+        Object.values(item.batches).forEach(line => draft.deleteLine?.(line))
+      );
       const successSnack = success(`Deleted ${selectedRows?.length} lines`);
       successSnack();
     } else {
--- a/packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
+++ b/packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
@@ -88,7 +88,7 @@ const useBatchRows = (summaryItem: OutboundShipmentSummaryItem | null) => {
     setBatchRows(() => {
       return data
         .map(batch => {
-          const matchingInvoiceRow = summaryItem.batches.find(
+          const matchingInvoiceRow = Object.values(summaryItem.batches).find(
             ({ stockLineId }) => stockLineId === batch.id
           );
           return { ...batch, quantity: matchingInvoiceRow?.numberOfPacks ?? 0 };
--- a/packages/invoices/src/OutboundShipment/DetailView/reducer.test.ts
+++ b/packages/invoices/src/OutboundShipment/DetailView/reducer.test.ts
@@ -1,8 +1,4 @@
-import {
-  DocumentAction,
-  Invoice,
-  DocumentActionSet,
-} from '@openmsupply-client/common';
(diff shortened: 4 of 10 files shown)


packages/common/src/utils/arrays/utils.ts
```diff
@@ -11,3 +11,12 @@ export const ifTheSameElseDefault = <
...
 };
+
+export const arrayToRecord = <T extends { id: string }>(
+  array: T[]
+): Record<string, T> => {
+  return array.reduce((acc, value) => {
+    acc[value.id] = value;
+    return acc;
+  }, {} as Record<string, T>);
+};
```
joshxg, line 22: hmm maybe could have made the key a parameter

packages/invoices/src/OutboundShipment/DetailView/Toolbar.tsx
```diff
@@ -27,13 +27,15 @@ export const Toolbar: FC<ToolbarProps> = ({ draft }) => {
...
+      .map(selectedId => draft.items.find(({ id }) => selectedId === id))
+      .filter(Boolean) as OutboundShipmentSummaryItem[],
   }));
 
   const deleteAction = () => {
     if (selectedRows && selectedRows?.length > 0) {
-      selectedRows.forEach(line => draft.deleteLine?.(line));
+      selectedRows.forEach(item =>
+        Object.values(item.batches).forEach(line => draft.deleteLine?.(line))
+      );
```
joshxg, line 38: the start of the changes needed to go from `lines` to `items`..

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
```diff
@@ -88,7 +88,7 @@ const useBatchRows = (summaryItem: OutboundShipmentSummaryItem | null) => {
     setBatchRows(() => {
       return data
         .map(batch => {
-          const matchingInvoiceRow = summaryItem.batches.find(
+          const matchingInvoiceRow = Object.values(summaryItem.batches).find(
```
joshxg, line 91: I've ended up making `batches` a lookup of id:batch as it makes it a little easier to search for a specific batch (iterate through items, and then through batches), so there's a decent amount of transforming it back to an array...

packages/invoices/src/OutboundShipment/DetailView/reducer.test.ts
```diff
@@ -1,129 +1,69 @@
-import {
-  DocumentAction,
-  Invoice,
-  createColumns,
-  DocumentActionSet,
-} from '@openmsupply-client/common';
+import { DocumentAction } from './../../../../common/src/hooks/useDocument/useDocument';
```
joshxg, line 1: obviously swapping led to a lot of test changes!

packages/invoices/src/OutboundShipment/DetailView/reducer.ts
```diff
@@ -22,37 +23,50 @@ import {
   OutboundShipmentSummaryItem,
   OutboundShipmentRow,
 } from './types';
+import { getDataSorter } from '../utils';
+
+const getExistingLine = (
```
joshxg, line 28: hmm this is where I thought maybe this idea sucked.

Basically searching for an existing batch is a bit annoying 🤷

packages/invoices/src/OutboundShipment/DetailView/reducer.ts
```diff
@@ -64,13 +78,7 @@ export const OutboundAction = {
     type: ActionType.UpdateInvoice,
     payload: { key, value },
   }),
-  updateNumberOfPacks: (
-    rowKey: string,
-    numberOfPacks: number
-  ): OutboundShipmentAction => ({
-    type: ActionType.UpdateNumberOfPacks,
-    payload: { rowKey, numberOfPacks },
-  }),
```
joshxg, line 73: never really used.. so instead of updating it.. 👋

packages/invoices/src/OutboundShipment/DetailView/reducer.ts
```diff
@@ -170,33 +161,49 @@ export const reducer = (
...
+            if (!existingSummaryItem) {
+              itemsArray.push({
+                ...summaryItem,
+                unitQuantity,
+                numberOfPacks,
+              });
+            }
+
+            return itemsArray;
+          }, draft.items);
```
joshxg, line 206: this is very similar, however we need to attach the merged lines from the server into a summary item, also

packages/invoices/src/OutboundShipment/DetailView/reducer.ts
```diff
@@ -249,61 +243,75 @@ export const reducer = (
...
+                  existingSummaryItem.batches[
+                    existingRow.id
+                  ] as OutboundShipmentRow
+                ).isDeleted = true;
+
+                const allDeleted = Object.values(
+                  existingSummaryItem.batches
+                ).every(({ isDeleted }) => isDeleted);
+                if (allDeleted) existingSummaryItem.isDeleted = true;
+              }
```
joshxg, line 314: this one was a bit more annoying because we need to check if everythings deleted and mark the summary item or potentially remove it.

Here I wondered how we're going to do deletes for a batch. Possibly put a checkbox in the batch expando?

mark-prins (approved):
👍  worth it! I think anyway. 
sounds like some of this was frustrating to sort out - it looks like you've sorted out a lot of core problems in this area all at once, which is probably why. there's a lot of work before you can see the benefits; needed doing though and I think it will be much better like this

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
joshxg: `Map` is a great idea. Wonder why that's not used more! Wonder if it is to do with serialisation .. 🤔 

The `Object` methods are pretty annoying! 

I also wondered if we could make `OutboundShipmentItem` an object with methods like `getRow` etc.. 🤔

packages/invoices/src/OutboundShipment/DetailView/reducer.ts
joshxg: oh! interesting. I don't remember/forgot about that. There is an explicit delete mutation (which seems easier than an update with 0 quantity?)

What I meant was that if we have a table of items grouped by item, from which you use the expand to view the batches, then the current checkbox would be on the summary item row, so would need some alternative for deleting a specific batch

mark-prins: ah, yes. that becomes a bit of a mess. force people to edit the row and adjust quantities? Delete the row and re-add it?
I prefer the explicit mutation to an implied delete where you set quantity to 0 and save and the row disappears. or something like that.

joshxg: I've kept the reducer action as per line for now. In the expando in the table, we could just add extra check boxes and adjust the 'table state' to account for 'sub rows'. Then we can have the summary item row checkbox check all the sub rows and delete the whole item, or just check the subrows in the expando for individual batches. Hope that makes sense 😨

Issue #422, opened 2021-11-15
Available quantity / FEFO assigning


joshxg:
In my effort to make rows editable, I have completely broken the `ItemDetailsModal`. I should probably make some issues to focus my attention better rather than making random PRs.

This one will be for
- Fix available quantity being hard coded to zero.
- Change FEFO assigning to only assign to the pack size in the drop down, and only after pressing an 'OK' button. Also add a toggle to switch between # of packs and unit quantities (where unit quantity ignores pack sizes?)

PR #424, opened 2021-11-15
#422 fefo stock issuing

joshxg:
#422 

- Has terrible UI, but think the functionality should be there

### DIFF
--- a/packages/common/src/ui/components/index.ts
+++ b/packages/common/src/ui/components/index.ts
@@ -13,6 +13,9 @@ import TextField from '@mui/material/TextField';
 import AccordionSummary from '@mui/material/AccordionSummary';
 import InputAdornment from '@mui/material/InputAdornment';
 import Tooltip from '@mui/material/Tooltip';
+import Radio from '@mui/material/Radio';
+import RadioGroup from '@mui/material/RadioGroup';
+import FormControlLabel from '@mui/material/FormControlLabel';
 
 export * from './portals';
 export * from './inputs';
@@ -43,4 +46,7 @@ export {
   Typography,
   TextField,
   Tooltip,
+  Radio,
+  RadioGroup,
+  FormControlLabel,
 };
--- a/packages/invoices/src/OutboundShipment/DetailView/modals/BatchesTable.tsx
+++ b/packages/invoices/src/OutboundShipment/DetailView/modals/BatchesTable.tsx
@@ -66,6 +66,7 @@ const BatchesRow: React.FC<BatchesRowProps> = ({ batch, label, onChange }) => {
       <BasicCell align="right">{label}</BasicCell>
       <BasicCell sx={{ width: '88px' }}>
         <ModalNumericInput
+          value={batch.quantity}
           inputProps={stockLineInputProps}
           disabled={isDisabled}
         />
@@ -160,11 +161,14 @@ export const BatchesTable: React.FC<BatchesTableProps> = ({
   const t = useTranslation();
   const onChangeValue: React.ChangeEventHandler<HTMLInputElement> = event =>
     onChange('placeholder', Number(event.target.value), 1);
+
   const placeholderInputProps = register('placeholder', {
     min: { value: 0, message: t('error.invalid-value') },
     pattern: { value: /^[0-9]+$/, message: t('error.invalid-value') },
     onChange: onChangeValue,
   });
+
+  const placeholderRow = rows.find(({ id }) => id === 'placeholder');
 
   return (
     <>
@@ -188,20 +192,25 @@ export const BatchesTable: React.FC<BatchesTableProps> = ({
             </TableRow>
           </TableHead>
           <TableBody>
-            {rows.map((batch, index) => (
-              <BatchesRow
-                batch={batch}
-                key={batch.id}
-                label={t('label.line', { number: index + 1 })}
-                onChange={onChange}
-              />
-            ))}
+            {rows
+              .filter(({ id }) => id !== 'placeholder')
+              .map((batch, index) => (
+                <BatchesRow
+                  batch={batch}
+                  key={batch.id}
+                  label={t('label.line', { number: index + 1 })}
+                  onChange={onChange}
+                />
+              ))}
             <TableRow>
               <BasicCell align="right" sx={{ paddingTop: '3px' }}>
                 {t('label.placeholder')}
               </BasicCell>
               <BasicCell sx={{ paddingTop: '3px' }}>
-                <ModalNumericInput inputProps={placeholderInputProps} />
+                <ModalNumericInput
+                  value={placeholderRow?.quantity}
+                  inputProps={placeholderInputProps}
+                />
               </BasicCell>
             </TableRow>
           </TableBody>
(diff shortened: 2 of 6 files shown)


packages/common/src/ui/components/index.ts
```diff
@@ -43,4 +46,7 @@ export {
   Typography,
   TextField,
   Tooltip,
+  Radio,
+  RadioGroup,
+  FormControlLabel,
```
joshxg, line 51: Just using the plain mui components. Shouldv'e done a styled radio before this!

packages/invoices/src/OutboundShipment/DetailView/modals/BatchesTable.tsx
```diff
@@ -66,6 +66,7 @@ const BatchesRow: React.FC<BatchesRowProps> = ({ batch, label, onChange }) => {
       <BasicCell align="right">{label}</BasicCell>
       <BasicCell sx={{ width: '88px' }}>
         <ModalNumericInput
+          value={batch.quantity}
```
joshxg, line 69: I just updated the quantity in the batch which re-rendered and meant I didn't use `setValue` so added `value` to `ModalNumericInput`

packages/invoices/src/OutboundShipment/DetailView/modals/BatchesTable.tsx
```diff
@@ -160,12 +161,15 @@ export const BatchesTable: React.FC<BatchesTableProps> = ({
...
   const onChangeValue: React.ChangeEventHandler<HTMLInputElement> = event =>
     onChange('placeholder', Number(event.target.value), 1);
+
   const placeholderInputProps = register('placeholder', {
     min: { value: 0, message: t('error.invalid-value') },
     pattern: { value: /^[0-9]+$/, message: t('error.invalid-value') },
     onChange: onChangeValue,
   });
 
+  const placeholderRow = rows.find(({ id }) => id === 'placeholder');
```
joshxg, line 171: Bit later on you see I added a `batchRow` which has the `id` `placeholder`

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsForm.tsx
```diff
@@ -14,19 +14,24 @@ import {
...
 interface ItemDetailsFormProps {
   allocatedQuantity: number;
-  quantity?: number;
   register: UseFormRegister<FieldValues>;
   summaryItem?: OutboundShipmentSummaryItem;
-  packSize: number;
   onChangeItem: (newItem: Item | null) => void;
-  onChangeQuantity: (quantity: number) => void;
-  setPackSize: (packSize: number) => void;
+  onChangeQuantity: (quantity: number, packSize: number | null) => void;
```
joshxg, line 32: calling back with the pack size selected and the quantity typed into the input. A null pack size is when we're using units.

This calls the `allocateQuantity` func which assigns a quantity to the batch rows with the right pack size

The 'quantity' only ended up being needed here so just shifted it to state here.

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsForm.tsx
```diff
@@ -40,37 +45,24 @@ export const ItemDetailsForm: React.FC<ItemDetailsFormProps> = ({
...
   summaryItem,
-  setPackSize,
-  packSize,
+  packSizeController,
+  availableQuantity,
 }) => {
   const t = useTranslation();
 
-  const quantityDescription = quantity ? (
+  const quantityDescription = allocatedQuantity ? (
```
joshxg, line 56: I was unsure about this, but I'm not sure what to do about the allocatedQuantity now?

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
```diff
@@ -166,79 +240,77 @@ export const ItemDetailsModal: React.FC<ItemDetailsModalProps> = ({
...
-
-  const allocateQuantities = () => {
+  const allocateQuantities = (
+    newValue: number,
+    issuePackSize: number | null
+  ) => {
+    setValue('quantity', String(newValue));
     // if invalid quantity entered, don't allocate
-    if (quantity < 1 || Number.isNaN(quantity) || !summaryItem) {
+    if (newValue < 1 || Number.isNaN(newValue)) {
```
joshxg, line 249: Changes here because I am bad with numbers and it took me a bit to make sure I understand, but should be the same as what you had except assigning to a `"batchRow"` rather than using `setValue`

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
```diff
@@ -166,79 +240,77 @@ export const ItemDetailsModal: React.FC<ItemDetailsModalProps> = ({
...
 
-    // allocate remainder to placeholder
-    setValue('placeholder', toAllocate);
-    setAllocated(quantity);
+    const placeholderIdx = newBatchRows.findIndex(
+      ({ id }) => id === 'placeholder'
+    );
+    const placeholder = newBatchRows[placeholderIdx];
+
+    if (!placeholder) throw new Error('No placeholder within item editing');
```
joshxg, line 295: I mean, not ideal, but shouldn't be happening. Need to get that nice logging in place!

joshxg (commented):
So my questions from here are:

- How the hell am I gonna make this look nice!
- What should the `allocated` be, now? I am thinking just a read only field similar to 'available' which just has the allocated units?
- When you press 'the button', should it *add* packs rather than wipe & assign? I think this seems more natural
- What should the copy be for the button? "Issue" doesn't seem right 🤔

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsForm.tsx
mark-prins: yeah, it doesn't really make sense now. Useful when there is a single quantity and we can show an indication that you are under the requested quantity - but that doesn't apply now.
A summary somewhere might be nice `x units allocated` and/or `x packs of y allocated` 🤷  
Easy enough to show a total number of units allocated, and maybe that's helpful.

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
```diff
@@ -86,18 +100,83 @@ const useBatchRows = (summaryItem: OutboundShipmentSummaryItem | null) => {
...
+  return { batchRows, isLoading, setBatchRows };
+};
+
+export type PackSizeController = ReturnType<typeof usePackSizeController>;
+
+const usePackSizeController = (batches: { packSize: number }[]) => {
+  // Creating a sorted array of distinct pack sizes
+  const packSizes = Array.from(
+    new Set(
+      batches
```
mark-prins, line 126: perhaps it would be good to filter batches that have no stock or are on hold. 
we aren't saving the requested pack size, so I'm wondering if there's value in multiplying quantity*pack size and popping that in the placeholder row 🤷

mark-prins (approved):
👍  I think it's already better than it was!

Just to prove yet again that naming stuff is hard - the `Issue` on the button isn't working for me.. Allocate makes more sense to me, though I'm sure others would disagree!

Couple of comments:
- The 'units' selection isn't adding anything for me - only works when I choose packs
- I think adding is more logical than replacing 
- The layout is pretty close already, I'm thinking that a few tweaks are all that is needed

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
joshxg: Ah yes good point. I suppose that's why you can add multiple placeholders in mSupply with different pack sizes? For all the effort we're putting into FEFOing with the right pack size I wonder why we aren't doing similar with supporting placeholders with the right pack size 🤔

PR #425, opened 2021-11-16
#273 upsert items

joshxg:
#273 

- finally?!
- This one makes editing lines work! It's working well against the mock server and workers
- There are still a few issues with available quantity I will point out in comments.
- I've noticed a bug with deleting (where deleting isn't deleting everything) which i'll fix elsewhere
- I've also had a bug a few times with an invalid stock line against the mock server and also once where there was an invalid quantity deduction. I checked my payload and all of the values sent seem to be correct, so I believe this is more about server errors so I am ignoring, for now (I also couldn't find a reliable replication, if you happen to find one I would like to fix the server regardless).
- At the moment, we are sending all lines per save as an update regardless of if they've been touched or not. This is for the use case of save -> edit row -> response comes back -> set all to clean or..?. Basically need a 'saving' and 'editted during save' state? Maybe silly and I should just mark them all as clean after a response 🤷

### DIFF
--- a/packages/common/src/operations.graphql
+++ b/packages/common/src/operations.graphql
@@ -92,6 +92,39 @@ query invoice($id: String!) {
             note
             locationDescription
             sellPricePerPack
+            stockLine {
+              __typename
+              ... on NodeError {
+                __typename
+                error {
+                  description
+                  ... on DatabaseError {
+                    __typename
+                    description
+                    fullError
+                  }
+                  ... on RecordNotFound {
+                    __typename
+                    description
+                  }
+                }
+              }
+              ... on StockLineNode {
+                __typename
+                availableNumberOfPacks
+                batch
+                costPricePerPack
+                expiryDate
+                id
+                itemId
+                packSize
+                sellPricePerPack
+                storeId
+                totalNumberOfPacks
+                onHold
+                note
+              }
+            }
           }
           totalCount
         }
--- a/packages/common/src/operations.graphql
+++ b/packages/common/src/operations.graphql
@@ -567,11 +567,13 @@ query stockCounts {
 mutation upsertOutboundShipment(
   $deleteOutboundShipmentLines: [DeleteOutboundShipmentLineInput!]
   $insertOutboundShipmentLines: [InsertOutboundShipmentLineInput!]
+  $updateOutboundShipmentLines: [UpdateOutboundShipmentLineInput!]
   $updateOutboundShipments: [UpdateOutboundShipmentInput!]
 ) {
   batchOutboundShipment(
     deleteOutboundShipmentLines: $deleteOutboundShipmentLines
     insertOutboundShipmentLines: $insertOutboundShipmentLines
+    updateOutboundShipmentLines: $updateOutboundShipmentLines
     updateOutboundShipments: $updateOutboundShipments
   ) {
     __typename
--- a/packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
+++ b/packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
@@ -11,7 +11,7 @@ interface DataRowProps<T extends DomainObject> {
   onClick?: (rowData: T) => void;
   rowData: T;
   rowKey: string;
-  ExpandContent?: FC;
+  ExpandContent?: FC<{ rowData: T }>;
 }
 
 export const DataRow = <T extends DomainObject>({
@@ -74,7 +74,7 @@ export const DataRow = <T extends DomainObject>({
       <tr>
         <td style={{ display: 'flex' }}>
           <Collapse sx={{ flex: 1 }} in={isExpanded}>
-            {ExpandContent ? <ExpandContent /> : null}
+            {ExpandContent ? <ExpandContent rowData={rowData} /> : null}
           </Collapse>
         </td>
       </tr>
--- a/packages/common/src/ui/layout/tables/hooks/useColumns/useColumns.tsx
+++ b/packages/common/src/ui/layout/tables/hooks/useColumns/useColumns.tsx
@@ -71,7 +71,7 @@ const getDefaultFormatter = <T extends DomainObject>(
     }
     case ColumnFormat.Currency: {
       return (value: unknown) => {
-        if (Number.isNaN(value)) return '';
+        if (Number.isNaN(Number(value))) return '';
 
         const formatNumber = useFormatNumber();
 
@@ -80,7 +80,7 @@ const getDefaultFormatter = <T extends DomainObject>(
       };
     }
     default: {
-      return (value: unknown) => String(value);
+      return (value: unknown) => String(value ?? '');
     }
   }
 };
--- a/packages/invoices/src/OutboundShipment/DetailView/SidePanel.tsx
+++ b/packages/invoices/src/OutboundShipment/DetailView/SidePanel.tsx
@@ -114,11 +114,28 @@ export const SidePanel: FC<SidePanelProps> = ({ draft }) => {
   return (
     <DetailPanelPortal
       Actions={
-        <DetailPanelAction
-          icon={<CopyIcon />}
-          titleKey="link.copy-to-clipboard"
-          onClick={copyToClipboard}
-        />
+        <>
+          {!process.env['NODE_ENV'] ||
+            (process.env['NODE_ENV'] === 'development' && (
+              <DetailPanelAction
+                icon={<CopyIcon />}
+                titleKey="dev.log-draft"
+                onClick={() =>
+                  draft.items.forEach(item => {
+                    console.table(item);
+                    Object.values(item.batches).forEach(batch => {
+                      console.table(batch);
+                    });
+                  })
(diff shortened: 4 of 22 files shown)


packages/common/src/operations.graphql
```diff
@@ -92,6 +92,39 @@ query invoice($id: String!) {
             note
             locationDescription
             sellPricePerPack
+            stockLine {
```
joshxg, line 95: I've added the stock line for the use case:

- Delete persisted line (so, flag and delete later)
- Create line for the same stock line

So in this case, reuse the 'deleted' line, which is easier if we have the stock line id (which I was already matching with, wrongly, as we didn't have the stock line)

packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
```diff
@@ -11,7 +11,7 @@ interface DataRowProps<T extends DomainObject> {
   onClick?: (rowData: T) => void;
   rowData: T;
   rowKey: string;
-  ExpandContent?: FC;
+  ExpandContent?: FC<{ rowData: T }>;
```
joshxg, line 14: Can remove this if you'd rather it not be on the demo server, but I just stringified the summary item and displayed it in the expando for something to look at

packages/common/src/ui/layout/tables/hooks/useColumns/useColumns.tsx
```diff
@@ -80,7 +80,7 @@ const getDefaultFormatter = <T extends DomainObject>(
       };
     }
     default: {
-      return (value: unknown) => String(value);
+      return (value: unknown) => String(value ?? '');
```
joshxg, line 83: fixing some of the undefined shenanigans which were displaying

packages/invoices/src/OutboundShipment/DetailView/SidePanel.tsx
```diff
@@ -114,11 +114,29 @@ export const SidePanel: FC<SidePanelProps> = ({ draft }) => {
...
+                  console.table(draft);
+                  draft.items.forEach(item => {
+                    console.table(item);
+                    Object.values(item.batches).forEach(batch => {
+                      console.table(batch);
+                    });
+                  });
+                }}
+              />
+            ))}
```
joshxg, line 133: Can also remove this if you'd like. Also, does this dev env work? 🤷 

This was the first time i've really wanted some debugging (need to set up the source maps for debugging nicely, I think) - basically just a button to log the draft, which was pretty helpful for me at the time.

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
```diff
@@ -105,7 +105,13 @@ const useBatchRows = (summaryItem: OutboundShipmentSummaryItem | null) => {
           const matchingInvoiceRow = Object.values(summaryItem.batches).find(
             ({ stockLineId }) => stockLineId === batch.id
           );
-          return { ...batch, quantity: matchingInvoiceRow?.numberOfPacks ?? 0 };
+          return {
+            ...batch,
+            quantity: matchingInvoiceRow?.numberOfPacks ?? 0,
+            availableNumberOfPacks:
+              batch.availableNumberOfPacks +
+              (matchingInvoiceRow?.numberOfPacks ?? 0),
```
joshxg, line 113: I'm not sure if this is right, need to confirm as the docs have:
image

But I don't think that makes sense for all stock lines, so assuming that the reduction in the invoice is applied to the available quantity returned for the api for the stock line, then..

- The 'total available' includes reductions from any invoice lines.
- So, to get the real 'available' amount, we need to add on what is on the invoice line (if the invoice is > draft?)

But, this gets messy of course.

Available(A) = 100, Invoice line packs (B) = 50, Real available(C) = 150
1. Open dialog to edit: A=100, B=50, so we display C=150
2. Reduce invoice line from 50 to 20 (A = 100, B=20, (C) = 170)
3. Open the form again without saving, we display 100(A)+20(B)=120(C)

So, we would need to store an 'untouched number of packs' value to fix this.

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
```diff
@@ -206,32 +212,20 @@ export const ItemDetailsModal: React.FC<ItemDetailsModalProps> = ({
...
-          'placeholder',
-          summaryItem,
-          { id: 'placeholder', expiryDate: '' },
-          placeholderValue
-        )
+    // TODO: Handle placeholder upserting.
+    const invoiceLines = batchRows
+      .filter(({ id }) => id !== 'placeholder')
+      .map(batch =>
+        getInvoiceLine(generateUUID(), summaryItem, batch, batch.quantity)
```
joshxg, line 221: I've just handled the zero quantity invoice lines in the reducer, as they might have been edited to zero quantity rather than not being touched. So, rather than track lines which have been touched on top of the quantity, just deriving if they already exist and deleting them

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
```diff
@@ -206,32 +212,20 @@ export const ItemDetailsModal: React.FC<ItemDetailsModalProps> = ({
...
-      invoiceLines.push(
-        getInvoiceLine(
-          'placeholder',
-          summaryItem,
-          { id: 'placeholder', expiryDate: '' },
-          placeholderValue
-        )
+    // TODO: Handle placeholder upserting.
+    const invoiceLines = batchRows
+      .filter(({ id }) => id !== 'placeholder')
```
joshxg, line 219: filtering out for now!

packages/invoices/src/OutboundShipment/DetailView/reducer.ts
```diff
@@ -252,19 +261,62 @@ export const reducer = (
             line
           );
 
-          if (existingSummaryItem && existingRow) {
-            existingRow.isUpdated = existingRow.isCreated ? false : true;
-            existingRow.isDeleted = false;
-            existingRow.numberOfPacks = line.numberOfPacks;
+          // If the row is being updated
+          if (existingSummaryItem) {
```
joshxg, line 265: what a mess... 😢

packages/invoices/src/api.ts
```diff
@@ -67,6 +71,14 @@ const linesGuard = (invoiceLines: InvoiceLineConnector | ConnectorError) => {
   throw new Error('Unknown');
 };
 
+const stockLineGuard = (stockLine: StockLineResponse): StockLineNode => {
+  if (stockLine.__typename === 'StockLineNode') {
+    return stockLine;
+  }
+
+  throw new Error('Unknown');
+};
```
joshxg, line 80: these are getting very tedious 🤔 

I wonder if we need to speak to the team about it? I supposes it's fine, to write this small function but if we query for an invoice line with it's stock line and we get a response that there was an error fetching the stock line, I don't know what else we would do other than say the invoice is broken, you can't edit it sorry 🤔

packages/invoices/src/OutboundShipment/DetailView/reducer.ts
mark-prins: good to get it in and then we can have a go at refactoring! though it's not bad and is clear enough to read

packages/invoices/src/api.ts
mark-prins: yeah, I'm wondering what we can do about it - they are tedious to write and have to be so specific; would be nice if the response had an error of a standard shape that could be checked

mark-prins:
I've noticed that the modal / form aren't cleared when re-opening now. Shall I make an issue for that? seems like a separate thing. Also wondered about enabling the `OK & NEXT` button too

mark-prins (approved):
😮  and 🙇 
There's a lot going on! nice work in pulling all of that together! I've had a reasonable run through and tested the site.. while I'm worried that I've missed things, it is all working and looks good!

joshxg:
@mark-prins The reopening i've just noticed also and am addressing in .. oh, I swear I made an issue 🙄 .. i'll make them now!

joshxg:
> wait! I'm re-reading the steps above and I think I see the problem :)
> Open dialog to edit: A=100, B=50 we display 150
> Reduce invoice line from 50 to 20 ( C is still 150! so A should be 130 )
> Open the form again without saving and we display 130(A) + 20(B) = 150 😄

Yes, A *should* be 130. However, A comes from the server - from a stock line available number of packs, or `batchRow.availableNumberOfPacks` and B is client side and is from the invoice line.

- When we reduce 50 to 20, we reduce the invoice line number of packs from 50 -> 20 client side.
- When we calculate C, we add the invoice line (20, client side) to the stock line (100, from the server). So we get 120 because A (on the server) hasn't been reduced. So we need to track the `30` somehow.

Either:
1. Cache all the stock lines so we can update them client side
2. Cache the 'original' invoice line value on the server

PR #428, opened 2021-11-16
#314 improved fefo UI

joshxg:
There's a bit going on here!

- Updated the FEFO UI bit. Think it looks OK?
image

image
- Made the allocation add rather than replace
- filtered out the invalid pack sizes from the pack size selection
- made placeholder in unit quantity? Though, the column header says # packs so maybe not a good idea.. or can put pack size 1? I dunno...
- added an `allocated` field - dunno if it's useful.
- changed the sort order a bit - I was getting confused when FEFO allocating 🤣  so I put the on hold rows at the top of the invalid area (and a divider between) and then the zero stock ones underneath 🤷 
- Made the headers sticky and added some heights cause the height changing so much was annoying me? Though, it's still not exactly right?
- added a loading indicator for the batches table
- Also changed the field BatchRow.quantity to number of packs to be consistent

### DIFF
--- a/packages/common/src/ui/components/index.ts
+++ b/packages/common/src/ui/components/index.ts
@@ -16,6 +16,7 @@ import Tooltip from '@mui/material/Tooltip';
 import Radio from '@mui/material/Radio';
 import RadioGroup from '@mui/material/RadioGroup';
 import FormControlLabel from '@mui/material/FormControlLabel';
+import Switch from '@mui/material/Switch';
 
 export * from './portals';
 export * from './inputs';
@@ -49,4 +50,5 @@ export {
   Radio,
   RadioGroup,
   FormControlLabel,
+  Switch,
 };
--- a/packages/invoices/src/OutboundShipment/DetailView/modals/BatchesTable.tsx
+++ b/packages/invoices/src/OutboundShipment/DetailView/modals/BatchesTable.tsx
@@ -30,6 +30,20 @@ export interface BatchesTableProps {
   rows: BatchRow[];
 }
 
+const sortByExpiry = (a: BatchRow, b: BatchRow) => {
+  const expiryA = new Date(a.expiryDate ?? '');
+  const expiryB = new Date(b.expiryDate ?? '');
+
+  if (expiryA < expiryB) {
+    return -1;
+  }
+  if (expiryA > expiryB) {
+    return 1;
+  }
+
+  return 0;
+};
+
 type BatchesRowProps = {
   batch: BatchRow;
   label: string;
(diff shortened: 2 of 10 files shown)


mark-prins:
Couple of quick comments before I go through the code
- Badass!! That's genius, a huuuuge improvement on the look and usability of the modal
- I'm getting a horizontal scroll, which I can remove by taking off the `100%` width from the table or the `auto` on the scroll-x
- the disabled rows look pretty good with a lighter font, just to show clearly that they are different; I was using `#ccc` for the color. was wondering about a different background but I think that would be too heavy
- 😁  it's so good

packages/common/src/ui/components/index.ts
```diff
@@ -16,6 +16,7 @@ import Tooltip from '@mui/material/Tooltip';
 import Radio from '@mui/material/Radio';
 import RadioGroup from '@mui/material/RadioGroup';
 import FormControlLabel from '@mui/material/FormControlLabel';
+import Switch from '@mui/material/Switch';
```
mark-prins, line 19: there is a small version of this in one of Manu's designs.. so I was hoping to add one in at some point, was going to use it for something else.
Really liked how clear the radio buttons were, though I think what you've done is better - nice and clean looking.

packages/invoices/src/OutboundShipment/DetailView/modals/BatchesTable.tsx
```diff
@@ -30,6 +30,20 @@ export interface BatchesTableProps {
   rows: BatchRow[];
 }
 
+export const sortByExpiry = (a: BatchRow, b: BatchRow) => {
```
mark-prins, line 33: I had meant to do the sorting which you mentioned ages ago. thanks for getting to it

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsForm.tsx
```diff
@@ -84,99 +68,113 @@ export const ItemDetailsForm: React.FC<ItemDetailsFormProps> = ({
...
+          <Box display="flex" flex={1} flexDirection="column">
+            <ModalRow>
+              <FormControlLabel
+                onChange={(_, checked) =>
+                  setIssueType(checked ? 'packs' : 'units')
+                }
+                control={
+                  <Switch defaultChecked color="secondary" size="small" />
+                }
+                label={<Box fontSize={10}>Allocate to Packs?</Box>}
```
mark-prins, line 110: could you add to the localisation please?

mark-prins (approved):
👍  not much more to say.. it's awesome!

joshxg:
Awesome. I'm just about to PR the ok & next so will be lazy and make the changes in that branch 😁

PR #431, opened 2021-11-17
#423 localisation 42

### LINKED ISSUE #79: Detect user language
Would be good to have the language default to the user's current environment settings.

Suggest using https://github.com/i18next/i18next-browser-languageDetector from this comment:

_Originally posted by @joshxg in https://github.com/openmsupply/openmsupply-client/pull/74#discussion_r697808032_

### LINKED ISSUE #423: Localisation maintenance
There's a few minor annoyances with the localisation implementation that would be nice to remove.
Currently, in order to support typing of `LocaleKey` all of the localisation files have to include all of the keys and values, which means that when you add a key, it has to be added to four files, manually.

I'd like to:
- find a nice way to organise localisation entries
- have an easy way to add new entries
- get a list of what is translated for each language
- use English as a fallback and not require all keys in all language files

mark-prins:
Fixes #423 
Changes from `reactl-intl` to `react-i18next` as this has browser language detection, async loading and support for namespaces which allows partitioning of language files.

Given that this has been such a nightmare, I've left off a couple of parts - so this PR will
- allow for a single file to be updated when adding a translation 
- and has fallback for other languages
- partitions language files into functional areas ( or whatever we like )
- browser language detection ( fixes #79 )
- async loading of language files

and.. coming soon in other PRs:
- a nice way to edit translations ( maybe.. not sure if I'll do that anytime soon )
- reorganising the language files
- updating components so that we can use them in a specific language-supported area
- fix all the tests

there is still something very strange with the import of modules, which this PR avoids now by including an unused method in `intlHelpers` which seems to alter the import order in a way that makes it all hang together. 😬


joshxg (approved):
It's great, thanks so much for handling this. Good to get something more solid in place!

I'm not 100% sure how the backend stuff works but I think I don't need to! 😁  I can see http requests going out when they're not in local storage so.. cool! I suppose the request is customisable/standard so the backend team just needs to host the json files?

When I `useTranslation("common")` in `Invoices` (only) then load `/dashboard`, the `common` json hasn't been downloaded, so when I navigate to `/outbound-shipments`, I get a flash of un translated strings, which after a re-render and the json having been downloaded, are correctly translated.

I was very interested in this so I looked into it a .. little!

We can 'pre-download' a bunch of name space jsons if we use `ns: ['common', 'app']`  etc in the i18next config options in the provider which fixes the flash of untranslated text

Though, I think a big benefit of name spaces is to be able to dynamically load them so you don't need to take the localisation file hit when first loading the page. Also though, they're cached so who cares? But then also.. dynamically loading name spaces would be good for plugins?

Well, it looked fun so I went down the rabbit hole and here is some information no one asked for:

- loading a json during runtime doesn't cause a re-render and [doesn't pause rendering](https://github.com/i18next/react-i18next/pull/523/files#diff-6cc2a4b04b8f73c37303657262b2e2827c7feabcdda33b3ba5885d06cad99cdcR57). So the flash of untranslated text I'm seeing is 'fixed' by our components re-rendering naturally as they fetch data. If there was no re-renders, the download of the bundle wouldn't cause a re-render and we'd still be stuck with the untranslated text
- I thought to create a component to be able to check if something was loaded using   if (i18n.hasResourceBundle(lng, ns)) return true; - but we'd have to poll to keep checking this which is not ideal..
- Turns out i18next emits events https://github.com/i18next/i18next/blob/e03dd9b9b50541e2395da52e27ccae026ca2c0f2/src/i18next.js#L131 - so we could set up a handler.. something like a component which uses a hook similar to this and renders `null` if the bundle hasn't loaded yet?


```
export const useIsBundleLoaded = (lng: string, ns: string) => {
  const [isBundleLoaded, setIsBundleLoaded] = React.useState(
    i18next.hasResourceBundle(lng, ns)
  );

  const handler = useCallback(e => {
    if (e === 'loaded') {
      setIsBundleLoaded(i18next.hasResourceBundle(lng, ns));
    }
  }, []);

  useEffect(() => {
    i18next.on('*', handler);
    return () => i18next.off('*', handler);
  }, []);

  return isBundleLoaded;
};

```

Using this we can ensure a json is loaded when navigating to a page.. which made me wonder if the partioning of localisation files should be something like `common` (and `app`?) live in the common package, and every other package just has their own json file and can inject it at runtime (seems just calling `useTranslation("some namespace")` triggers a request to `http://localhost:3003/locales/en/some%20namespace.json` to load it in 🤔 )

joshxg:
Hopefully that merge was right - was just deleting the old language json files 🤔

mark-prins:
awesome - thanks!

PR #432, opened 2021-11-18
Add update to mock server to create inbound and outbound shipments

### LINKED ISSUE #430: setup mock server to handle requests for inbound shipments
currently the invoices are hard coded customer invoices and the invoices query just returns all the invoices. So would be good to not hard code the type and use a filter to differentiate between the invoice types.

Also, the stock lines are created with some random stock and then invoice lines are created and take packs away. Would be better if we instead, create inbound invoice lines which can create the stock lines, then outbound invoice lines to take stock away like in reality. Although it's mock data and we can just put anything, it's easier when the data make sense!

joshxg:
Fixes #430 

- Made it so stock lines are created with zero quantity and then the invoices are created, then we take all the inbound invoices, a random subset of stocklines and create some invoice lines - adding to the stock as we go. Then the same for the outbounds, taking a random subset of stock lines and some random percentage of the stock lines available quantity and adjusting stock as we go through.
- This actually ends up being simpler creation code, of course 🙄 
- feel free to change numbers if you want more/less data to be committed!

### DIFF
--- a/packages/invoices/src/OutboundShipment/ListView/ListView.tsx
+++ b/packages/invoices/src/OutboundShipment/ListView/ListView.tsx
@@ -39,7 +39,10 @@ export const OutboundShipmentListViewComponent: FC = () => {
     filter,
     invalidate,
   } = useListData(
-    { initialSortBy: { key: 'otherPartyName' } },
+    {
+      initialSortBy: { key: 'otherPartyName' },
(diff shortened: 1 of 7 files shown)


mark-prins (approved):
👍  that looks great! much better to have some valid-ish data

Issue #410, opened 2021-11-10
Add inbound shipment detail view


joshxg:
Assuming this is at least similar to the outbound shipment view

unsure about:
- detail panel/side panel - what to show
- what columns to show in the primary list view
- what to display in the add/edit item view

Need to find out and add to: https://docs.google.com/spreadsheets/d/1H0Rv1DEOgW0cE07wpfc_KivLf8TcNhYDqlwRp92x-To/edit#gid=0

PR #439, opened 2021-11-19
#410 inbound detail view 2

joshxg:
More of #410

- This one moves around a bunch of types.
- Hopefully it's a good idea - I moved all the invoice related types from common to the `invoices` package. That's the only place they're used. If they're going to be used elsewhere my thinking is that it's likely that it would be a different query and might need a different type anyway, rather than messing with these ones 🤷 
- Additionally made the `Invoice`/`InvoiceRow` types extend the gql `InvoiceNode`
- I also noticed that the `operations.graphql` file is going to be ridiculous. Same with the `schema.graphql`. I'm thinking that at least the `operations.graphql` file should live next to the `api.ts` file in each submodule which those operations are for.. though there will likely be a bunch of duplicates... probably. But, I do think we need to partition this file somehow as we edit it manually. `schema.graphql` is a copy paste of the graphql schema file from the api which is not ideal but at least somewhat automated.. i'd really like to split that out also, but it would make the process very tedious and annoying. Any ideas?


packages/common/src/operations.graphql
```diff
@@ -27,6 +27,7 @@ query invoice($id: String!) {
       shippedDatetime
       deliveredDatetime
       enteredByName
+      donorName
```
joshxg, line 30: Ah I just realised this was my addition to the type, but the generated code was melting down

packages/common/src/types/index.ts
```diff
@@ -52,78 +51,3 @@ export type Store = {
   id: string;
   name: string;
 };
-
-export interface InvoiceLine extends InvoiceLineNode, DomainObject {
```
joshxg, line 56: So all of this moved to `invoices` package.. Think the other ones should go in their packages also - i.e. items/name into system. Stock lines.. uhh.. system also?

packages/common/src/ui/components/domain/StatusCrumbs/StatusCrumbs.stories.tsx
```diff
@@ -1,36 +1,30 @@
...
 import Radio from '@mui/material/Radio';
 import RadioGroup from '@mui/material/RadioGroup';
 import FormControlLabel from '@mui/material/FormControlLabel';
 import FormControl from '@mui/material/FormControl';
 import FormLabel from '@mui/material/FormLabel';
+import { StatusCrumbs } from './StatusCrumbs';
+import { LocaleKey, useTranslation } from '../../../../intl';
 
-const StatusTranslation: Record<OutboundShipmentStatus, LocaleKey> = {
+const statusTranslation: Record<string, LocaleKey> = {
```
joshxg, line 12: did some status changes a little further down so just changed this story up a bit..

packages/invoices/src/InboundShipment/ListView/ListView.tsx
```diff
@@ -48,13 +48,11 @@ export const InboundListView: FC = () => {
...
-    onUpdate({ ...row, color: color.hex });
-  };
-
-  const columns = useColumns<InvoiceRow>(
+  const columns = useColumns(
     [
-      getNameAndColorColumn(onColorUpdate),
+      getNameAndColorColumn((row: InvoiceRow, color: Color) => {
+        onUpdate({ ...row, color: color.hex });
+      }),
```
joshxg, line 55: not sure why 🤔

packages/invoices/src/InboundShipment/ListView/api.ts
```diff
@@ -56,7 +55,7 @@ const getSortKey = (sortBy: SortBy<Invoice>): InvoiceSortFieldInput => {
   }
 };
 
-const getSortDesc = (sortBy: SortBy<Invoice>): boolean => {
+const getSortDesc = (sortBy: SortBy<InvoiceRow>): boolean => {
```
joshxg, line 58: this is the list view, so uses invoice row not invoice!

packages/invoices/src/OutboundShipment/DetailView/Footer.tsx
```diff
@@ -52,79 +52,81 @@ export const Footer: FC<OutboundDetailFooterProps> = ({ draft, save }) => {
...
-            }}
-            labelKey="label.hold"
-          />
-
-          <StatusCrumbs
-            statuses={outboundStatuses}
-            statusLog={createStatusLog(draft)}
-            statusFormatter={getStatusTranslation}
-          />
+        draft && (
```
joshxg, line 55: hmm I guess I dislike the optional chaining

joshxg: nothing else really changed here - just the calls for statuses as the draft is now definitely a draft.. 🙄

packages/invoices/src/OutboundShipment/DetailView/index.ts
```diff
@@ -1,38 +1 @@
-import { OutboundShipment } from '../../types';
 export { DetailView } from './DetailView';
-
-export const placeholderInvoice: OutboundShipment = {
```
joshxg, line 4: put this in `invoices/utils`. Though, now that I think about it. Why?

packages/common/src/types/index.ts
mark-prins: hmm, yeah. stock lines are shared across a number of areas.. might be best to have them in system

packages/invoices/src/OutboundShipment/DetailView/Footer.tsx
mark-prins: fair enough! that was a lot of chaining

packages/invoices/src/OutboundShipment/DetailView/index.ts
mark-prins: either place is fine - it is a utils kind of thing 🤷

packages/invoices/src/types.ts
```diff
@@ -86,3 +134,45 @@ export type OutboundShipmentSummaryItem = {
...
+  itemId: string;
+  itemCode: string;
+  itemName: string;
+  unitQuantity: number;
+  numberOfPacks: number;
+  locationDescription?: string | null;
+  itemUnit?: string;
+  batch?: string | null;
+  batches: Record<string, OutboundShipmentRow>;
+  sellPrice?: number | undefined;
```
mark-prins, line 166: I was wondering what the differences are between In & Out BoundShipments, and here's one! Outbound will show `sellPrice` and Inbound should show `costPrice`.

All the same, both types can have both props - it's only the column list which needs to differ!

mark-prins (approved):
🎉  Nice! very sensible moving of stuff around generally.

I agree about the `operations.graphql` getting too large! not sure how to handle the splitting of the `schema.graphql` though.. if we are taking it directly from the API then it would be easiest to simply copy, regardless of size.. though I am worried about how big it will get.

packages/invoices/src/types.ts
joshxg: Yeah there's really not much differences right now and maybe not ever, maybe it's backwards but if it turns out there's no difference we can consolidate

PR #443, opened 2021-11-19
Add inbound detail view reducer

joshxg:
#410 

- The reducer for the detail view
- You guessed it!
- I think if we stick with the summary items we will be able to use the same reducer for inbound and outbounds

### DIFF
--- a/packages/invoices/src/InboundShipment/DetailView/reducer/reducer.test.ts
+++ b/packages/invoices/src/InboundShipment/DetailView/reducer/reducer.test.ts
@@ -1,0 +1,3 @@
+describe('Inbound shipment DetailView Reducer', () => {
+  it('is a placeholder', () => {});
+});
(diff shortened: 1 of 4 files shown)


packages/invoices/src/InboundShipment/DetailView/reducer/reducer.test.ts
```diff
@@ -0,0 +1,3 @@
+describe('Inbound shipment DetailView Reducer', () => {
+  it('is a placeholder', () => {});
+});
```
joshxg, line 3: I'm keen to wait and see if changes are needed to the reducer. currently it's a copy and paste which already has tests

packages/invoices/src/InboundShipment/DetailView/reducer/reducer.ts
```diff
@@ -0,0 +1,231 @@
...
+  ).reduce(getUnitQuantity, 0);
+
+  const numberOfPacks = Object.values<InboundShipmentRow>(
+    summaryItem.batches
+  ).reduce(getSumOfKeyReducer('numberOfPacks'), 0);
+
+  return { unitQuantity, numberOfPacks };
+};
+
+export const InboundAction = {
```
joshxg, line 62: Using the outbound actions at the moment.. the boilerplate for reducers is too damn high!

mark-prins (approved):
👍  cool, that looks good! Would be good to see if we can re-use for in & out bound

Issue #454, opened 2021-11-20
revert all changes to graphql schema making it incompatible with the current backend


joshxg:
- *I* want to work on inbound shipment mutations and also go back to error handling and polishing outbounds.
- *I* want to ensure that that we stay compatible with the backend because:
    - Implementing the error system in the mock server is likely not worth the time.
    - It's taking so long for changes that I think a better system would be to move on to a *new* area rather than cause incompatibilities. For example, move onto other areas where there are not already GQL types defined which cause the incompatibilities - defining the requisition types or implementing a 'log' type.
    - This is in contradiction to wanting to work depth rather than breadth, but that is, I think evidently, not working well.
    - We can have the service worker capture any requests which aren't implemented on the backend - i.e. for requisitions or updating requisitions, but we can't 'change' queries that are incompatible with the backend - or, at least, it would be more difficult. Ideally for me, we still have the exisiting ones implemented so that we can use the service worker for tests and for when we are developing normally so we don't have to run the rust server *all* the time, just when implementing new features.
- And on the front-end-fun branch: It requires other changes, I think? It's not working for me - I think it requires schema changes and I think it'll take less time, though a bit messier, to just revert all the incompatible changes.

To do this, I want to:
- Comment out all the changes in the schema which is incompatible with the backend
- Comment out all the changes in the gql operations which is incompatible with the backend
- Comment out all the changes in the code base using the fields/types which get removed from the schema is incompatible with the backend

Not ideal but I've checked and there aren't really many changes - mainly sorting, colour, statuses/dates and some fields on the lines. Commenting out so they can be re-added, eventually.

An issue right now is that graphiql, where I've been getting a pretty print of the gql schema, doesn't work on the current remote server develop branch. It appears to be from the changes made recently to using workspaces, but I have no idea. I've found a commit which has graphiql working from which I don't *think* there are any other schema changes since. So, I've made a branch from that older commit https://github.com/openmsupply/remote-server/tree/%23check-graphiql which has the same schema as we do currently. Other changes are in PRs but I don't want to bother spending time on that.. as even with those changes, we still won't be compatible.

This could be a really bad idea. Hope not.

mark-prins:
I agree - not ideal, but something needs to be done. I don't want to maintain a separate API branch and since we can't have queries asking for fields that aren't there, I think we have to try new areas rather than depth within the areas done.

[...]

PR #455, opened 2021-11-20
Revert all incompatible changes

joshxg:
#454 

Well I hope this isn't a bad idea. 

 I think this will be all round better as we can implement the more complex things like handling errors

all up this took about 30 minutes. Should only take less to revert things as they come in to the server

### DIFF
--- a/packages/host/public/config.js
+++ b/packages/host/public/config.js
@@ -1,4 +1,4 @@
 window.env = {
-  API_URL: 'http://localhost:4000',
-  // API_URL: 'http://localhost:8000/graphql',
+  // API_URL: 'http://localhost:4000',
+  API_URL: 'http://localhost:8000/graphql',
 };
--- a/packages/host/public/config.js
+++ b/packages/host/public/config.js
@@ -1,5 +1,5 @@
 window.env = {
-  // API_URL: 'http://localhost:4000', - default URL for the mock graphql server
-  // API_URL: 'http://localhost:8000/graphql' - default URL for the backend graphql server
+  // API_URL: 'http://localhost:4000', // - default URL for the mock graphql server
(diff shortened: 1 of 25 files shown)


packages/host/public/config.js
```diff
@@ -1,4 +1,4 @@
 window.env = {
-  API_URL: 'http://localhost:4000',
-  // API_URL: 'http://localhost:8000/graphql',
+  // API_URL: 'http://localhost:4000',
+  API_URL: 'http://localhost:8000/graphql',
```
mark-prins, line 3: I can update the version on demo-open.msupply.org and we could commit that URL rather than requiring a locally running server

packages/host/src/bootstrap.tsx
```diff
@@ -9,7 +9,7 @@ if (process.env.NODE_ENV === 'development') {
   } = require('@openmsupply-client/mock-server/src/worker/client');
   const worker = setupMockWorker();
 
-  worker.start();
+  // worker.start();
```
mark-prins, line 12: would the msw be permanently disabled?

mark-prins (approved):
😢  bye bye all you beautiful unsupported things. I'll see you again one day..

one final thing - I'm getting an error when trying to add items - might be just me? I think it could be the dates in my stock lines.. just check that it's ok for you?
```
"availableBatches": {
                        "__typename": "ConnectorError",
                        "error": {
                            "description": "Database Error",
                            "__typename": "DatabaseError",
                            "fullError": "\"DIESEL_DESERIALIZATION_ERROR\" (\"ParseError(Invalid)\")"
                        }
                    },
```

joshxg:
Changed the URL and reverted the disabling of the msw. Also, not getting an error for adding 🤔  Could very well be an issue that is on my end, not yours!Will move forward for now though

Issue #448, opened 2021-11-19
Add item adding and editing functionality to an inbound shipment


joshxg:
A lot of sub tasks for this one, but for now

- Add an invoice item details modal
- Functionality to add an item
- Functionality to edit an item

joshxg:
details that should be in the modal: https://docs.google.com/spreadsheets/d/1H0Rv1DEOgW0cE07wpfc_KivLf8TcNhYDqlwRp92x-To/edit#gid=320810900&range=A116:S116

Though, some are not supported by the API so should hold off on those, I think.

joshxg:
Am assuming this modal is similar to the outbound one, without the FEFO. Though, maybe that's a mistake. There is a lot more fields on each batch.

I count 11 editable fields and 4 read only fields. The read only could go in the 'info' popover, but that is a lot of editable columns in a small table.

Currently we have 12 columns. 1 editable, so maybe it's not too many? 🤷

PR #461, opened 2021-11-20
#448 invoice line edit 1

joshxg:
Starting to work on #448 

- This is adding a modal to add an item.
- Offroad switch of the slide animation into `useDialog` which implicitly animates when you click the next button. Mainly this was just fun so I can remove if not useful..!

### DIFF
--- a/packages/common/src/hooks/useDialog/useDialog.tsx
+++ b/packages/common/src/hooks/useDialog/useDialog.tsx
@@ -102,7 +102,7 @@ export const useDialog = (dialogProps?: DialogProps): DialogState => {
         height={height}
       >
         <ModalTitle title={title} />
-        <DialogContent>
+        <DialogContent sx={{ overflowX: 'hidden' }}>
           <Slide in={slideConfig.in} direction={slideConfig.direction}>
             <div> {children}</div>
           </Slide>
--- a/packages/common/src/hooks/useDialog/useDialog.tsx
+++ b/packages/common/src/hooks/useDialog/useDialog.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import DialogActions from '@mui/material/DialogActions';
 import DialogContent from '@mui/material/DialogContent';
+import { Slide } from '../../ui/animations';
 import { BasicModal } from '../../ui/components/modals/BasicModal';
 import { ModalTitle } from '../../ui/components/modals/ModalTitle';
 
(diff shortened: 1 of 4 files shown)


packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
```diff
@@ -82,6 +88,17 @@ export const DetailView: FC = () => {
...
+      <Modal
+        title={t('heading.add-item')}
+        cancelButton={<DialogButton variant="cancel" onClick={hideDialog} />}
+        nextButton={<DialogButton variant="next" onClick={() => {}} />}
+        okButton={<DialogButton variant="ok" onClick={hideDialog} />}
+        height={600}
+        width={900}
+      >
+        <InboundLineEdit />
+      </Modal>
```
joshxg, line 101: This could be silly but thought i'd throw this into the detail view for now. It might make disabling the buttons more difficult - but we're only checking if quantities are zero to disable it. And actually.. maybe we should be allowing that? What's the point of disabling it 🤔

joshxg (commented):
Bit of a small one. Though I can see it might be silly and I should revert it I wanted to keep the diff seperate!

packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
mark-prins: happy with either option. the outbound was moved out in order to control the validation and disable the OK, but that process has changed now.
Might not be required - I just wonder if it will be strange if the server strips out or deletes 0 quantity lines, which I think it was going to do? I don't see any value in adding a zero quantity item; it provides another way to remove an existing line which I think is more intuitive than the bulk action - but the comment has also been made that we don't want multiple ways of doing the same thing.
Overall I'd prefer to keep the validation - would be interested what others think though

mark-prins (approved):
😁  awesome!

packages/common/src/hooks/useDialog/useDialog.tsx
```diff
@@ -12,9 +13,15 @@ export interface ButtonProps {
...
 export interface ModalProps {
+  children: React.ReactElement<any, any>;
   cancelButton?: JSX.Element;
   height?: number;
-  nextButton?: JSX.Element;
+  nextButton?: React.ReactElement<{
+    onClick: () => void;
+    variant: 'next';
+    disabled: boolean;
+  }>;
```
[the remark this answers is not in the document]
joshxg: Yeah you're right that it's more flexible just trying to ensure there's an onclick! 
6e5dc22

It does stop me from example passing `null`, but not `<div/>` so.. relatively useless anyway 🤷

packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
joshxg: I think I heard someone mention that zero stock lines should be removed 🤔  so.. should be nothing to prune

PR #462, opened 2021-11-21
#448 inbound line adding 2

joshxg:
This is adding a search input, some item fields and a table to the inbound edit item modal.

It's very loosey goosey right now. I wanted to get it in to have something concrete to show so we can determine whats best to do with the large amount of editable fields on an inbound row v the outbound rows.

Some options I can see:
- Just Scroll
- Tabs - i.e. `general`, `pricing`, `weights`, `discrepancies` `totals`
- 'Tabs' but actually they just scroll to the columns related to the above
- expando
- keep only general details in the table, click on a row and some 'detail' area shows up elsewhere to edit other things
- 1 row per modal and don't use the table [I find this to be least preferable so we can be more consistent 🤷 ]
- use some sort of 'card' in the area for each row, and have the modal vertically scroll - edit: apparently this is a 'responsive table'
image

### DIFF
--- a/packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
+++ b/packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
@@ -45,7 +45,7 @@ const useDraftInbound = () => {
 };
 
 export const DetailView: FC = () => {
-  const t = useTranslation('common');
+  const t = useTranslation('outbound-shipment');
   const { hideDialog, showDialog, Modal } = useDialog();
   const { draft, save, onChangeSortBy, sortBy } = useDraftInbound();
 
@@ -66,6 +66,8 @@ export const DetailView: FC = () => {
     setSelectedItem({ item: null, editing: false });
     showDialog();
   };
+
+  const isEditMode = selectedItem.editing;
 
   const columns = useColumns(
     [
@@ -106,7 +108,7 @@ export const DetailView: FC = () => {
       <SidePanel draft={draft} />
 
       <Modal
-        title={t('heading.add-item')}
+        title={!isEditMode ? t('heading.add-item') : t('heading.edit-item')}
         cancelButton={<DialogButton variant="cancel" onClick={hideDialog} />}
         nextButton={<DialogButton variant="next" onClick={() => {}} />}
         okButton={<DialogButton variant="ok" onClick={hideDialog} />}
--- a/packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
+++ b/packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
@@ -16,7 +16,7 @@ import {
   Item,
 } from '@openmsupply-client/common';
 
-import { reducer } from './reducer';
+import { InboundAction, reducer } from './reducer';
 import { getInboundShipmentDetailViewApi } from './api';
 import { Toolbar } from './Toolbar';
 import { Footer } from './Footer';
@@ -27,7 +27,6 @@ import { InboundLineEdit } from './modals/InboundLineEdit';
 
 import { isInboundEditable } from '../../utils';
 import { InboundShipmentItem } from '../../types';
-import { OutboundAction } from '../../OutboundShipment/DetailView/reducer';
 
 const useDraftInbound = () => {
   const { id } = useParams();
@@ -40,7 +39,7 @@ const useDraftInbound = () => {
   );
 
   const onChangeSortBy = (column: Column<InboundShipmentItem>) => {
-    dispatch(OutboundAction.onSortBy(column));
+    dispatch(InboundAction.onSortBy(column));
   };
 
   return { draft, save, dispatch, onChangeSortBy, sortBy: state.sortBy };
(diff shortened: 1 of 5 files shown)


packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit.tsx
```diff
@@ -1,5 +1,302 @@
...
+import { flattenInboundItems } from '../../../utils';
+interface InboundLineEditProps {
+  item: InboundShipmentItem | null;
+  onUpsert: (item: InboundShipmentItem) => void;
+  onChangeItem: (item: Item | null) => void;
+}
 
-export const InboundLineEdit: FC = () => {
-  return null;
+const BasicCell: React.FC<TableCellProps> = ({ sx, ...props }) => (
```
mark-prins, line 31: I'm thinking that a bunch of these components will be useful elsewhere and we may have to lift them out to `common`

packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit.tsx
```diff
@@ -1,5 +1,302 @@
...
+                  Sent Pack <br /> Size
+                </HeaderCell>
+              </TableRow>
+            </TableHead>
+
+            {flattenInboundItems([inboundItem]).map(batch => (
+              <BatchRow key={batch.id} batch={batch} />
+            ))}
+
+            <Fab
```
mark-prins, line 280: The rectangular drop shadow on the FAB troubles me:

Screen Shot 2021-11-22 at 12 01 49 PM

Changing the viewbox of `PlusCircleIcon` (almost) fixes the alignment, and then you can use the `40px` version of the fab, which I think looks good:

`viewBox="0 0 21 20.25"`

Screen Shot 2021-11-22 at 12 18 33 PM

that's with this on the fab: 
```
              sx={{
                alignSelf: 'flex-end',
                margin: '10px',
                alignItems: 'center',
                justifyContent: 'center',
              }}
```

mark-prins (approved):
👍  that's great! 
adding rows is very neat - like what you've done with that!

Couple of thoughts for later
- it did come up in the UI discussions.. we'll need a date picker now 😃 
- ok & next keeps the same item selected, we'd need to clear that out

current approach of having lots of columns..? Fine for now and we can work that through with Manu. I'm aware that a very wide table is just taking the msupply approach through, and we can probably do better, or at least different 😜  The card example you have could work nicely.. maybe expando though I worry about having things hidden

packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
```diff
@@ -38,18 +39,70 @@ const useDraftInbound = () => {
...
+  });
+
+  console.log('-------------------------------------------');
+  console.log('selectedItem', selectedItem);
+  console.log('-------------------------------------------');
+
+  const onChangeSelectedItem = (newItem: Item | null) => {
+    if (!newItem) return setSelectedItem({ item: newItem, editing: false });
+
+    // Try and find the outbound summary row that matches the new item
```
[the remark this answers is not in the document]
joshxg: this is why no one should write comments!!!

packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
```diff
@@ -38,18 +39,70 @@ const useDraftInbound = () => {
...
+    batches: {},
+    unitQuantity: 0,
+    numberOfPacks: 0,
+  };
+};
+
 export const DetailView: FC = () => {
-  const t = useTranslation('common');
-  const { hideDialog, showDialog, Modal } = useDialog();
+  const t = useTranslation('outbound-shipment');
```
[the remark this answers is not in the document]
joshxg: I've made a bit of a mess without thinking actually. Sorry 😁 

I'll do a PR with distribution?

packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit.tsx
joshxg: yup for sure!

I was actually wondering if we could try and use the `DataTable` also..

packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit.tsx
joshxg: much better! Thanks

joshxg:
Awesome thanks.

Yes, a date picker for expiry and also a drop down or autocomplete for the location.

The annoying thing about date pickers for expiries is most date pickers need a day picked, where as the expiry is usually just a month and year (I think!). Though, I see mui expose a month and year picker, so hopefully we can do something nice with those - that'd be a bit of fun

My main concern with moving forward with this modal is that if we move into a 'by batch' implementation then it will be a lot of wasted work. Do you think we can lock in the 'by item' (as much as we can, I suppose)? Then i'd be happy to continue on with getting the features in!

packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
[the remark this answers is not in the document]
mark-prins: I'll make an issue to rename the `outbound-shipment` one to tidy that up.

PR #463, opened 2021-11-21
#448 Very rough tabbed implementation

joshxg:
Just a very rough tab implementation to give an idea. I thought I wouldn't like tabs but turns out I quite like it. 

Would be good to determine if people actually use all the tabs or not. I imagine they don't?

Adding keyboard shortcuts to navigate between tabs would make it fine to use, I think. Could also add a 'summary' or 'totals' tab.

### DIFF
--- a/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit.tsx
+++ b/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit.tsx
@@ -18,6 +18,10 @@ import {
   generateUUID,
   NumericTextInput,
   formatDate,
+  TabContext,
+  TabList,
+  Tab,
+  TabPanel,
 } from '@openmsupply-client/common';
 import { InboundShipmentItem, InboundShipmentRow } from '../../../types';
 import { ItemSearchInput } from '@openmsupply-client/system';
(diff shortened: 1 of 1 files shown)


packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit.tsx
```diff
@@ -86,72 +102,270 @@ const BatchRow: FC<{ batch: InboundShipmentRow }> = ({ batch }) => {
...
+          key={batch.id}
+          batch={batch}
+        />
+      ))}
+    </Table>
+  );
+};
+
+enum Tabs {
+  General = 'General',
```
mark-prins, line 362: Batch? Quantity? Just wondering if we can get away from general :)

Guessing that there will be some opinions about names 😉  What you've done is sensible and clear to me.
Translations, when we go ahead with it.

mark-prins (approved):
👍  that goes to show the value of seeing something in action. I didn't think I'd like the tabs, thought that they'd be confusing to use, but I really like it. 

- You can happily ignore the bits you don't care about, and focus on the important fields
- The horizontal scroll is generally considered terrible usability, so good to avoid if possible
- The grouping of columns is helpful I think

joshxg:
I know!

I did a quick mock up in a codesandbox to experiment and it was surprisingly not terrible so I thought to give this a go. I thought I would be weirded out that the context was so similar after pressing a tab, but it seems OK.

I've renamed to `batch` as there is also the `quantity` and `location` to be assigned in that tab, though batch doesn't quite work and maybe that means the grouping is not good? 🤷  We'll see, I suppose

Issue #444, opened 2021-11-19
Add Replenishments top level navigation unit and shift inbound shipments


joshxg:
Just realised inbound shipments should be in a section called replenishments, not distribution.

We've spoken before about removing the 'container' packages - at the moment `catalogue` and `distribution`. Think this would be a good time to do it.

So

- Remove `catalogue` package, shift code into host
- Remove `distribution` package, shift code into host
- add `replenishments` section
- shift inbound shipments into `replenishments`

PR #466, opened 2021-11-22
#444 Shifting around host files to make me feel tidy

joshxg:
Just shifting around some files as I want to move the distribution and catalogue packages into the host and make some more folders!


mark-prins (approved):
👍  tidy = good 😄

PR #467, opened 2021-11-22
#444 replenishments top level nav 2

joshxg:
This one moves the catalogue and distribution nav components into the host package


mark-prins (approved):
👍  thanks! that's much clearer

PR #468, opened 2021-11-22
#444 replenishments top level nav 3

### LINKED ISSUE #444: Add Replenishments top level navigation unit and shift inbound shipments

joshxg:
Fixes #444 

Finally this removes the distribution/catalogue packages and puts the routers in `host` - also adds the top level `replenishment` nav item.

p.s. is the navigation hovering a bit weird if you try to go from distribution -> replenishments?


mark-prins:
> is the navigation hovering a bit weird if you try to go from distribution -> replenishments?

the navigation hovering troubles me anyway - distribution to catalogue was funky; the replenishment just falls into the same swamp.

mark-prins (approved):
👍  very nice! works beautifully ( yes, apart from the hover on the nav.. separate problem! )

PR #473, opened 2021-11-22
#472 suspense loading spinner

### LINKED ISSUE #472: The main suspense fallback should be a better spinner
The main loading component when the app is loading is just `"Loading..."` which is annoying me.

I also found a funny spinner so want to add it 😁

joshxg:
Fixes #472 

It would be best if you load this and take a look before you look at the code..

It's the spinner when you first load the page!

### DIFF
--- a/packages/common/src/ui/components/loading/Biker/Biker.tsx
+++ b/packages/common/src/ui/components/loading/Biker/Biker.tsx
@@ -1,0 +1,87 @@
+import React, { FC } from 'react';
+import { GlobalStyles, Typography, Box } from '@mui/material';
+
+const styles = {
+  '.biker': {
+    width: '106px',
+    height: '56px',
+    display: 'block',
+    margin: '30px auto',
+    backgroundImage:
+      'linear-gradient(#000 50px, transparent 0), linear-gradient(#000 50px, transparent 0), linear-gradient(#000 50px, transparent 0), linear-gradient(#000 50px, transparent 0), radial-gradient(circle 14px, #000 100%, transparent 0)',
+    backgroundSize: '48px 15px , 15px 35px, 15px 35px, 25px 15px, 28px 28px',
+    backgroundPosition: '25px 5px, 58px 20px, 25px 17px, 2px 37px, 76px 0px',
+    backgroundRepeat: 'no-repeat',
+    position: 'relative',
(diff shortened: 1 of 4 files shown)


mark-prins:
ha! you beat me to it! I was going to do the same thing - the `loading` suspense fallback bothered me too 😁

packages/common/src/ui/components/loading/Biker/Biker.tsx
```diff
@@ -0,0 +1,87 @@
...
+      transform: 'rotate(0deg)',
+    },
+    '100%': {
+      transform: 'rotate(360deg)',
+    },
+  },
+} as const;
+
+const text = [
+  'One sec, I got hit by a bus...',
```
mark-prins, line 52: I think if you put this out there, and ask for suggestions on telegram, we'd get 1,000 odd posts

mark-prins: As it is, I was hitting refresh a dozen times trying to read them all 🤣

mark-prins (approved):
🥇  legend! that is just too perfect!

I had thought of putting in an animated something and really liked the random message idea that videostream uses; once again you have done it better than I could have! 🤣

packages/common/src/ui/components/loading/Biker/Biker.tsx
joshxg: That message is paying homage to our falling comrade William!

Look, the implementation isn't ideal but .. whatever!

joshxg: Oh. I did forget to mention that I tried using `NODE_ENV` and it doesn't work (Was going to put a boring spinner when not in dev!) so we'll have to setup some build thing for some dev variable, I think. But, I thought not to use translations as we'd use a 'real' (🙄 ) spinner in prod (Or, maybe just no text 🤣 )

mark-prins: it's not like the translations work in prod environment anyway! 🤣 
I was going to look at `NODE_ENV` - had a feeling that wasn't working. I'll make an issue!

joshxg:
Thanks! 😁

PR #476, opened 2021-11-22
add navigation for Inventory

mark-prins:
First step of the `Inventory` section. Moving the nav and router components out, as per `Distribution` etc.


joshxg (approved):
Nice! Certainly a lot easier with a lot less boiler plate 🎉

Issue #470, opened 2021-11-22
SummaryItems have a `lines` key as well as an `items` key which is confusing


joshxg:
Current `OutboundShipment` type extends `Invoice`. `Invoice` has `lines` which is an array of `InvoiceLines` and `OutboundShipment` has `items` which is an array of `OutboundShipmentItem`s.

`OutboundShipmentItem`s contain `InvoiceLines` - so the `InvoiceLines` against the `Lines` key become out dated. 

Replace `lines` with the `OutboundShipmentItem` array?

mark-prins:
Sounds good, thanks!

Issue #469, opened 2021-11-22
Sort by key for columns should be either a key of or a function


joshxg:
We should add a property to `Column` and `ColumnDefinition` - `sortKey` which is either a `keyof`, or is a function the revolves to .. something.

This is so we can write a generic sorting function which is either a key on the object, or calls a function to which resolves to the value to use to compare

mark-prins:
fixed-ish with #687

PR #477, opened 2021-11-23
Add remove objectwithstrings type

### LINKED ISSUE #459: The 'RecordWithStringKeys'  utility type causes issues with asserting keys
The `RecordWithStringKeys` type: `Record<string, unknown>` causes issues with using `keyof`.

I.e.

```
const someKey: keyof Invoice = "nope"
```

is fine.

This means trying to type something like

```
type updater = <K extends keyof Invoice>(invoiceKey: K, value: Invoice[K]) => void
```

Which is trying to ensure that the value passed is of the right value for the key passed, doesn't work and you can pass any string key and any value .. oops! 🙄

joshxg:
Fixes #459 and https://github.com/openmsupply/openmsupply-client/issues/470

- Fixes it.. after I make a bunch more issues for problems this also causes :)
- I have now learned a strong lesson about extending a type everywhere and making sure it is correct when I do... 🙄 
- Also lots of TS lessons in general ...
- Basically this was not restricting the 'domain objects' to their actual fields, so invoice.xsdfdsf was valid..


packages/common/src/hooks/useFilterBy/useFilterBy.tsx
```diff
@@ -11,39 +11,37 @@ type FilterRule = {
     | FilterByConditionByType['date']]?: unknown;
 };
 
-export type FilterBy<T> = Partial<Record<Partial<keyof T>, FilterRule>>;
+export type FilterBy = Record<string, FilterRule | null>;
```
joshxg, line 14: removing the `ObjectWithStringKeys` causes a whole bunch of issues everywhere.. This was taking a generic and then using a key to filter things by. 

I had some type issues with this, then I realised that we shouldn't be trying to restrict filtering to only keys on the object, otherwise things will be annoying. For example on related entity fields. 

So, issue to make #1 Make filtering somehow type safe..?  We can use the GQL types, maybe, so that type could be the generic here. But in general this filtering hook I don't really like and is quite difficult to use..

packages/common/src/hooks/useSortBy/useSortBy.ts
```diff
@@ -1,28 +1,26 @@
-import { ObjectWithStringKeys } from './../../types/utility';
 import { useCallback, useState } from 'react';
 
-export interface SortRule<T extends ObjectWithStringKeys> {
+export interface SortRule<T> {
   key: keyof T | string;
```
joshxg, line 4: also realising for the same reason as filtering, that sorting shouldn't be restricted to the keys on an object. I've made #469 previously to try and sort this out.

packages/common/src/hooks/useSortedData/useSortedData.ts
```diff
@@ -1,9 +1,9 @@
 import { SortRule, SortBy, getDataSorter } from '@openmsupply-client/common';
-import { ObjectWithStringKeys } from './../../types/utility';
+
 import { useEffect, useState } from 'react';
 import { useSortBy } from '../useSortBy';
 
-interface SortedDataState<T extends ObjectWithStringKeys> {
+interface SortedDataState<T> {
```
joshxg, line 6: To be honest this isn't even used

packages/common/src/ui/components/inputs/Autocomplete/utils.tsx
```diff
@@ -9,12 +9,12 @@ export const DefaultAutocompleteItemOption = styled('li')(({ theme }) => ({
...
 type Option<T> = { label: string } & T;
 
-export type OptionMapper<T extends Record<keyof T, unknown>> = (
+export type OptionMapper<T extends { [K in keyof T]: T[K] }> = (
   options: T[],
   key: keyof T
 ) => Option<T>[];
 
-export const defaultOptionMapper = <T extends Record<keyof T, unknown>>(
+export const defaultOptionMapper = <T extends { [K in keyof T]: T[K] }>(
```
joshxg, line 17: this, after removing the `ObjectWithStringKeys` was having a few problems with things being passed in. This type is 'copying' over the same type from T, so optionals are carried over, where as `Record<keyof T, unknown>` was making it so every key was non optional

packages/common/src/ui/layout/tables/columns/types.ts
```diff
@@ -76,5 +76,5 @@ export interface Column<T extends DomainObject> {
 
 export interface ColumnDefinition<T extends DomainObject>
   extends Partial<Omit<Column<T>, 'key'>> {
-  key: keyof T | GenericColumnKey;
+  key: keyof T | GenericColumnKey | string;
```
joshxg, line 79: and again. Realising that restricting a column to a key that must be on the object is not ideal. Related entities, the `Select` column and any other 'random' cells we need to add will break this rule and there's no real reason for it, other than the `DataAccessor` returning a 'valid' value as a default. Maybe this could be extended to be something like.. either the `key` is a `keyof T` OR, `DataAccessor` must be supplied?

packages/invoices/src/InboundShipment/DetailView/reducer/reducer.ts
```diff
@@ -112,15 +112,15 @@ export const createSummaryItem = (
...
-    locationDescription: ifTheSameElseDefault(
-      batches,
-      'locationDescription',
-      undefined
-    ),
+    // locationDescription: ifTheSameElseDefault(
+    //   batches,
+    //   'locationDescription',
+    //   undefined
+    // ),
```
joshxg, line 123: once I made the keys of the domain objects actually start getting checked, a decent amount of compiler errors came up, using fields which weren't actually on the types.. 🙄

packages/invoices/src/InboundShipment/DetailView/reducer/reducer.ts
```diff
@@ -155,20 +155,18 @@ export const reducer = (
...
-          Object.keys(draft).forEach(key => {
-            // TODO: Sometimes we want to keep the user entered values?
-            if (key === 'items' || key === 'lines') return;
-            draft[key] = data[key];
-          });
+          state.draft = {
+            ...state.draft,
+            ...data,
+            status: state.draft.status,
+          };
```
joshxg, line 162: This is equivalent (pretty sure 😁 )

Taking the `data` fields over `draft`, and then later on, `items` gets assigned correctly.

The `status` was just because of type annoyances..

packages/invoices/src/InboundShipment/DetailView/reducer/reducer.ts
```diff
@@ -242,15 +240,8 @@ export const reducer = (
...
-
-          break;
-        }
-
-        case ActionType.UpdateInvoice: {
-          const { payload } = action;
-          const { key, value } = payload;
-
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore
```
joshxg, line 244: issue #2 ... this (and in the other reducer) having issues with these types

packages/invoices/src/OutboundShipment/DetailView/Toolbar.tsx
```diff
@@ -64,7 +64,7 @@ export const Toolbar: FC<ToolbarProps> = ({ draft }) => {
                     disabled={!isInvoiceEditable(draft)}
                     value={draft.otherParty}
                     onChange={name => {
-                      draft.update?.('name', name);
+                      draft.update?.('otherParty', name);
```
joshxg, line 67: more bugs from being not typed properly!

packages/invoices/src/InboundShipment/ListView/api.ts
```diff
@@ -132,7 +132,7 @@ export const getInboundShipmentListViewApi = (
   onCreate: async (invoice: Partial<Invoice>): Promise<string> => {
     const result = await api.insertInboundShipment({
       id: invoice.id ?? '',
-      otherPartyId: String(invoice['nameId']) ?? '',
+      otherPartyId: invoice?.otherPartyId ?? '',
```
joshxg, line 135: more bugs from being not typed properly!

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsModal.tsx
```diff
@@ -44,9 +44,9 @@ export const getInvoiceLine = (
   itemId: summaryItem.itemId,
   itemName: summaryItem.itemName,
   itemCode: summaryItem.itemCode,
-  itemUnit: summaryItem.itemUnit ?? '',
+  // itemUnit: summaryItem.itemUnit ?? '',
   batch: stockLineOrPlaceholder.batch ?? '',
-  locationDescription: stockLineOrPlaceholder.locationDescription ?? '',
+  // locationDescription: stockLineOrPlaceholder.locationDescription ?? '',
```
joshxg, line 49: more bugs from being not typed properly!

packages/invoices/src/OutboundShipment/DetailView/reducer.ts
```diff
@@ -116,15 +120,15 @@ export const createSummaryItem = (
...
-    locationDescription: ifTheSameElseDefault(
-      batches,
-      'locationDescription',
-      undefined
-    ),
+    // locationDescription: ifTheSameElseDefault(
+    //   batches,
+    //   'locationDescription',
+    //   undefined
+    // ),
```
joshxg, line 131: more bugs from being not typed properly!

packages/invoices/src/OutboundShipment/DetailView/reducer.ts
```diff
@@ -159,26 +163,24 @@ export const reducer = (
...
-          Object.keys(draft).forEach(key => {
-            // TODO: Sometimes we want to keep the user entered values?
-            if (key === 'items') return;
-            draft[key] = data[key];
-          });
+          state.draft = {
+            ...state.draft,
+            ...data,
+            status: state.draft.status,
+          };
```
joshxg, line 170: same as the other reducer 🤣

packages/invoices/src/utils.ts
```diff
@@ -18,85 +18,72 @@ import {
 export const placeholderInbound: InboundShipment = {
   id: '',
   otherPartyName: '',
-  total: '',
   comment: '',
   theirReference: '',
-  color: 'grey',
+  otherParty: undefined,
+  otherPartyId: '',
+  items: [],
```
joshxg, line 25: more bugs from being not typed properly!

joshxg (commented):
Oops just noticed some failing tests.

joshxg:
passing now

joshxg:
I've merged in and changed the base to your list-view-2 branchso I don't have to do it later!

joshxg:
Ah.. oops! I've accidentally pushed a commit for https://github.com/openmsupply/openmsupply-client/issues/470 

Hope that's nottoo confusing 😁

packages/common/src/hooks/useDocument/types.ts
```diff
@@ -2,7 +2,7 @@ import { Dispatch } from 'react';
 
 export interface Api<ServerData, Document> {
   onRead: (id: string) => Promise<ServerData>;
-  onUpdate: (val: Document) => Promise<ServerData>;
+  onUpdate: (val: Document) => Promise<Document>;
```
joshxg, line 5: this helped with the typings for now 😕

packages/host/src/components/AppDrawer/AppDrawer.test.tsx
```diff
@@ -66,7 +66,7 @@ describe('AppDrawer', () => {
       expect(getByText(/distribution/i)).toBeVisible();
       expect(getByText(/dashboard/i)).toBeVisible();
       // expect(getByText(/suppliers/i)).toBeVisible();
-      expect(getByText(/stock/i)).toBeVisible();
+      expect(getByText(/inventory/i)).toBeVisible();
```
joshxg, line 69: offroad

packages/common/src/hooks/useDocument/types.ts
mark-prins: oh. bother.. still, better than nothing

packages/common/src/hooks/useFilterBy/useFilterBy.tsx
mark-prins: I'm not seeing any use case where you'd filter by a field that isn't in the resulting data.. but there could well be something. in which case, making it type safe is an interesting challenge

packages/common/src/hooks/useSortBy/useSortBy.ts
mark-prins: same as filtering - there's a UI problem when sorting by something that isn't shown; that can be confusing for users. maybe if we have complex sorting options then we will need something like that

packages/common/src/ui/components/inputs/Autocomplete/utils.tsx
mark-prins: that's very neat!

packages/common/src/ui/layout/tables/columns/types.ts
mark-prins: the `DataAccessor` would be a good option, though I can't quite see how that is possible. `string` good for now..

packages/host/src/components/AppDrawer/AppDrawer.test.tsx
mark-prins: oopsie

packages/invoices/src/InboundShipment/DetailView/reducer/reducer.ts
mark-prins: it works! thanks typescript 😃

mark-prins (approved):
👍  wow, that's a lot of work! great to have some typing errors show up, thanks for picking up on that original problem and then working it through so well

packages/common/src/hooks/useFilterBy/useFilterBy.tsx
joshxg: For example, the `InvoiceNode` - the total of the invoice becomes `invoice.pricing.totalAfterTax` - so if you wanted to do a filter for all invoices greater than X - or sort. Or filter by a transaction category which will also probably be nested 🤔

Issue #475, opened 2021-11-22
Add Stock list


mark-prins:
No design! Thought to pop in a table to show current stock 🤷

mark-prins:
oops! #416

PR #478, opened 2021-11-23
#475 stock list 2

### LINKED ISSUE #416: Stock list
For the `Stock` menu item, what about implementing the stock list, as per desktop?
There are no designs, so I propose this:

- Create a list view 
- An action buttons to export the list ( save as excel or csv )
- List displays
  - code
  - name
  - batch
  - expiry
  - location
  - pack size
  - stock on hand
- input to allow searching by item name or code

See the [requirements spreadsheet](https://docs.google.com/spreadsheets/d/1H0Rv1DEOgW0cE07wpfc_KivLf8TcNhYDqlwRp92x-To/edit?pli=1#gid=0&range=A296) for detail

mark-prins:
Fixes #416 

Decided not to export to excel at this stage. Also the query endpoint isn't ideal since I have to filter out for zero stock items. for now, I've reused the `getStockListViewApi` call and have just asked for 1000 rows, then filtered and sorted manually in the `ListView`. Means that it only has to query once and still supports pagination and sorting.

### DIFF
--- a/packages/system/src/Stock/ListView/api.ts
+++ b/packages/system/src/Stock/ListView/api.ts
@@ -1,0 +1,63 @@
+import {
+  ListApi,
+  SortBy,
+  ItemSortFieldInput,
+  OmSupplyApi,
+  FilterBy,
+} from '@openmsupply-client/common';
+import { availableBatchesGuard, itemsGuard } from '../../Item/utils';
+import { StockRow } from '../types';
+const onRead =
+  (api: OmSupplyApi) =>
+  async ({
+    sortBy,
+    filterBy,
+  }: {
+    first: number;
+    offset: number;
+    sortBy: SortBy<StockRow>;
+    filterBy: FilterBy<StockRow> | null;
+  }): Promise<{
+    nodes: StockRow[];
+    totalCount: number;
+  }> => {
+    const result = await api.itemsWithStockLines({
+      first: 1000,
+      offset: 0,
+      key: ItemSortFieldInput.Name,
+      desc: sortBy.isDesc,
+      filter: filterBy,
+    });
+
+    const items = itemsGuard(result);
+    const nodes: StockRow[] = [];
+    items.nodes.forEach(item => {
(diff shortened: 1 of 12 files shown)


packages/system/src/Stock/ListView/api.ts
```diff
@@ -0,0 +1,63 @@
...
+      filter: filterBy,
+    });
+
+    const items = itemsGuard(result);
+    const nodes: StockRow[] = [];
+    items.nodes.forEach(item => {
+      const availableBatches = availableBatchesGuard(item.availableBatches);
+      availableBatches
+        .filter(batch => batch.totalNumberOfPacks > 0)
+        .forEach(batch =>
```
joshxg, line 38: just for the future.. I think showing items with no stock is *probably* helpful?

joshxg (approved):
fantastic.

It is really helpful to be able to see stock 😁

packages/system/src/Stock/ListView/api.ts
```diff
@@ -0,0 +1,63 @@
...
+  };
+
+export const getStockListViewApi = (api: OmSupplyApi): ListApi<StockRow> => ({
+  onRead:
+    ({ first, offset, sortBy, filterBy }) =>
+    () =>
+      onRead(api)({ first, offset, sortBy, filterBy }),
+  onDelete: async () => [''],
+  onUpdate: async () => '',
+  onCreate: async () => '',
```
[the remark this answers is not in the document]
mark-prins: yeah - it may change for names, but some things will always be central data and therefore read only lists on the client.. maybe?

packages/system/src/Stock/ListView/api.ts
mark-prins: probably is - we don't currently, you have to go through items and check the stock on hand there. really, this whole list could be replaced by a toggle or filter on the item list 'only show items with stock on hand' and maybe a view option which shows a row per batch. but hey.. this is useful for now

PR #480, opened 2021-11-23
#448 inbound line editing 5

joshxg:
Thought I would float this idea - using `DataTable` as the table in the `InvoiceLineEdit` modal.

### DIFF
--- a/packages/common/src/ui/layout/tables/components/Cells/TextInputCell/TextInputCell.tsx
+++ b/packages/common/src/ui/layout/tables/components/Cells/TextInputCell/TextInputCell.tsx
@@ -1,0 +1,31 @@
+import React from 'react';
+import { CellProps } from '../../..';
+import { BasicTextInput } from '../../../../../components/inputs/TextInput';
+import { DomainObject } from '../../../../../../types';
+import { useDebounceCallback } from '../../../../../..';
+
+type R<T> = T &
+  DomainObject & { update?: (key: string, value: string) => void };
+
+type CP<T> = CellProps<R<T>>;
+
+export const TextInputCell = <T extends DomainObject>({
+  rowData,
+  column,
+}: CP<T>): React.ReactElement<CP<T>> => {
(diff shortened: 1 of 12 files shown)


packages/common/src/ui/layout/tables/context/TableContext.ts
```diff
@@ -17,6 +18,7 @@ export interface TableStore {
   toggleSelected: (id: string) => void;
   toggleAll: () => void;
   setActiveRows: (id: string[]) => void;
+  setDisabledRows: (id: string[]) => void;
```
joshxg, line 21: I added this mainly because I wanted to check that setting a row as disabled wouldn't be too difficult (i.e. if we wanted to use it for `ItemDetailsModal`)

packages/common/src/ui/layout/tables/components/Cells/TextInputCell/TextInputCell.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { CellProps } from '../../..';
+import { BasicTextInput } from '../../../../../components/inputs/TextInput';
+import { DomainObject } from '../../../../../../types';
+import { useDebounceCallback } from '../../../../../..';
```
mark-prins, line 5: I think it doesn't matter - but given the issues with imports, I'd feel safer if these were referencing explicitly `columns` and `hooks`. helps a little with readability, knowing where things are sourced too

mark-prins (approved):
👍  Nice! that works well and I really like the idea of reusing the table. impressed that you managed to combine the functionality! looks good.

packages/common/src/ui/layout/tables/components/Cells/TextInputCell/TextInputCell.tsx
joshxg: Yes! good catch, thanks

joshxg:
Awesome. Maybe i'll put off continuing through with this until we get a bit of feedback on the ui

PR #484, opened 2021-11-23
Add updated/improved item search input behaviour

### LINKED ISSUE #458: Item search input behaviour is a bit weird!
it's a little weird searching for an item.  For example when I go to edit a name of an existing item I have to backspace twice to clear the input, and have to clear the whole thing. 

Also, it's making extra requests on loading to try and 'find' the item it's loaded with and things.

I think trying to do something closer to https://codesandbox.io/s/2ops5?file=/demo.js mixed with https://codesandbox.io/s/48q8z?file=/demo.js would be good

joshxg:
Fixes #458 

- improves the behaviour I think
- I wanted to put this logic into a component .. "AutocompleteWithDefaultValue" ... and now i'm realising maybe there was a better way to do this with the default value prop.. crap

### DIFF
--- a/packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
+++ b/packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
@@ -58,21 +58,27 @@ export const itemToSummaryItem = (item: Item): InboundShipmentItem => {
   };
 };
 
+export enum ModalMode {
+  Create,
+  Update,
+}
+
 export const DetailView: FC = () => {
   const t = useTranslation('outbound-shipment');
 
   const { draft, save, onChangeSortBy, sortBy } = useDraftInbound();
 
-  const [selectedItem, setSelectedItem] = React.useState<{
+  const [modalState, setModalState] = React.useState<{
     item: InboundShipmentItem | null;
-    editing: boolean;
-  }>({ item: null, editing: false });
+    mode: ModalMode;
+  }>({ item: null, mode: ModalMode.Create });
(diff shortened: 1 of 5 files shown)


packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
```diff
@@ -58,21 +58,27 @@ export const itemToSummaryItem = (item: Item): InboundShipmentItem => {
   };
 };
 
+export enum ModalMode {
+  Create,
+  Update,
+}
+
```
joshxg, line 65: I haven't made this change in the detail view for outbounds, because i'm not sure that the behaviour of disabling the item input when youre in update mode is wanted?

packages/system/src/Item/Components/ItemSearchInput/ItemSearchInput.tsx
```diff
@@ -31,59 +31,62 @@ interface ItemSearchInputProps {
...
+    if (value && buffer && open) {
+      setBuffer(null);
+      setFilter({
+        searchTerm: '',
+        field: 'name',
+      });
+    } else if (!open) {
+      setBuffer(value);
+    }
+  }, [open, value, buffer]);
```
joshxg, line 71: all up this component generally seems easier to follow. Also does less network requests thankfully!

mark-prins (approved):
🎉  no helpful comments sorry! that's very slick and works beautifully 😁

PR #486, opened 2021-11-24
#406 filter out items

### LINKED ISSUE #406: Should filter any items which are already in the invoice in the item details modal
Currently all items are showing in the drop down, not just those which are not in the invoice already, which results in being able to add multiple lines for the same batch

Two options I can think of:
1. Filter out the items which are already in the invoice
2. When selecting an item which is already in the invoice, essentially acts as you having clicked on the row for the item, showing the invoice lines for that item (I can see this could be cool. But also it could just be confusing, so I think 1)

joshxg:
Fixes #406 

- Can't do `not` on a filter, so doing it in JS
- I quite like the added `extraFilter` prop anyway. Think it'll come in handy.
- Don't like the prop drilling of the draft

### DIFF
--- a/packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsForm.tsx
+++ b/packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsForm.tsx
@@ -53,7 +53,14 @@ export const ItemDetailsForm: React.FC<ItemDetailsFormProps> = ({
         <ModalLabel label={t('label.item')} />
         <Grid item flex={1}>
           <ItemSearchInput
-            currentItemName={summaryItem?.itemName}
+            currentItem={{
+              name: summaryItem?.itemName ?? '',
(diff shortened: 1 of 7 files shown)


packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsForm.tsx
```diff
@@ -20,7 +20,7 @@ import {
   Box,
 } from '@openmsupply-client/common';
 import { ItemSearchInput } from '@openmsupply-client/system/src/Item';
-import { OutboundShipmentSummaryItem } from '../../../types';
+import { OutboundShipment, OutboundShipmentSummaryItem } from '../../../types';
```
mark-prins, line 23: maybe time to rename this file to `OutboundLineEdit` 🤔

mark-prins (approved):
👍  Yeah! very nicely done 😁

PR #488, opened 2021-11-24
#456 ok next weird

### LINKED ISSUE #456: OK & Next with one item should close the modal
Currently OK & Next with one item just loads the same item again. Should close if there's just one

mark-prins:
Fixes #456 

Quick one to disable OK & Next if the user is editing the only item. I know you said to make the button act as OK, but thought that was misleading, since there is no 'next' to go to 🤷

### DIFF
--- a/packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
+++ b/packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
@@ -152,7 +152,15 @@ export const DetailView: FC = () => {
             : t('heading.edit-item')
         }
         cancelButton={<DialogButton variant="cancel" onClick={hideDialog} />}
-        nextButton={<DialogButton variant="next" onClick={() => {}} />}
+        nextButton={
+          <DialogButton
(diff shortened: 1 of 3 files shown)


packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
```diff
@@ -152,7 +152,15 @@ export const DetailView: FC = () => {
             : t('heading.edit-item')
         }
         cancelButton={<DialogButton variant="cancel" onClick={hideDialog} />}
-        nextButton={<DialogButton variant="next" onClick={() => {}} />}
+        nextButton={
+          <DialogButton
+            variant="next"
+            onClick={() => {}}
+            disabled={
+              modalState.mode === ModalMode.Update && draft.items.length === 3
```
mark-prins, line 160: branched off yours which has `ModalMode` because that's much nicer to use 😄

joshxg (approved):
Yup that makes more sense I think

PR #490, opened 2021-11-24
Add searchbar component

### LINKED ISSUE #408: Debounced search filter for list view
Currently the list view has a small text field filter. It should be debounced.

Currently, the callback called sets the state with the new values, which is then fed into the filter, which then triggers the api call on state change. Currently everything is .. nicely? in sync! But, would be good to have the search text field ahead of the filter, which means duplicating some state somewhere.

Suggest making a `search bar` component, which is uncontrolled and holds it's own state, and debounces the setting of the filter state.

joshxg:
Fixes #408 

Wanted to fix the lag when typing into the search input on the list views - thought to go with a search bar component to handle it.

I haven't quite got the icon looking exactly right, I tried messing with the view box but I just made it worst.. was hoping you could help me out on that one as I seem to just not understand 🙄

### DIFF
--- a/packages/common/src/ui/icons/Search.tsx
+++ b/packages/common/src/ui/icons/Search.tsx
@@ -1,0 +1,21 @@
+import React from 'react';
+import SvgIcon, { SvgIconProps } from '@mui/material/SvgIcon';
+
+export const SearchIcon = (props: SvgIconProps): JSX.Element => {
+  return (
+    <SvgIcon
+      {...props}
+      sx={{
+        strokeWidth: '2px',
+        strokeLinecap: 'round',
+        strokeLinejoin: 'round',
+        fill: 'none',
+        stroke: 'currentColor',
+      }}
+      viewBox="0 0 20 21"
+    >
+      <circle cx="11" cy="11" r="8"></circle>
+      <line x1="21" y1="21" x2="16.65" y2="16.65"></line>
+    </SvgIcon>
+  );
+};
--- a/packages/common/src/ui/icons/Search.tsx
+++ b/packages/common/src/ui/icons/Search.tsx
@@ -3,19 +3,8 @@ import SvgIcon, { SvgIconProps } from '@mui/material/SvgIcon';
 
 export const SearchIcon = (props: SvgIconProps): JSX.Element => {
   return (
-    <SvgIcon
-      {...props}
-      sx={{
-        strokeWidth: '2px',
-        strokeLinecap: 'round',
-        strokeLinejoin: 'round',
-        fill: 'none',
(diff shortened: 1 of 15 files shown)


packages/invoices/src/OutboundShipment/ListView/Toolbar.tsx
```diff
@@ -56,23 +55,14 @@ export const Toolbar: FC<{
...
-              filter.onChangeStringFilterRule(
-                'otherPartyName',
-                'like',
-                e.target.value
-              )
-            }
-          />
-        }
+      <SearchBar
+        placeholder="Search by comment..."
```
joshxg, line 61: the api doesn't support filtering by other party name - have changed to comment for now. Didn't add a translation (and so it begins?) because no one wants to search by comment, normally, right!

joshxg:
Ah nuts. Just noticed the search bar is spinning now on the first initial load and also on sorting - of course!

Thinking I will remove passing in the loading prop as I can't think of a different way to fix it properly 🤔

packages/common/src/ui/icons/Search.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
```
mark-prins, line 1: I've committed a suggestion cf400c3
Have taken the icon from [the closest I could find in zeplin](https://app.zeplin.io/project/60e44fff6fa942168092f1ff/screen/611ce286e8a19f15c3bc56c2)
and the colour from the [suggestion for the date picker icon](https://app.zeplin.io/project/60e44fff6fa942168092f1ff/screen/619a9538c40639a9f6a246cc)

I also thought it looked better smaller

mark-prins (approved):
👍  Wicked! Really like the debounced search bar and its working nicely 😁 

Have added a suggestion for the search icon.. and think it looks better a little more subtle, so went smaller and greyer ( or grayer? 🤷  ). Not entirely sold on the search spinner spinning when the page first loads, along with the one in the body of the page.. but happy to go with it!

joshxg:
Ah thanks a lot that's awesome. I've removed the search bar loading indicator when a query is flying, and just a loading indicator when the 'debounce' is happening.. I think it seems OK and seems as though something is happening while you're typing, then when you stop the query returns quickly even with a very slow throttle on.. so maybe it'll be ok.. if not, removing the spinner completely is always a good option 😁

packages/invoices/src/OutboundShipment/ListView/Toolbar.tsx
joshxg: Fixed the inbound not creating - the search issues I think are resolved with backend fix

PR #491, opened 2021-11-24
#474 outbound shipment namespace

### LINKED ISSUE #474: Rename the outbound-shipment language namespace
I think the language namespaces should reflect functional areas; and also not be too specific. The `outbound-shipment` is more targeted than necessary.

joshxg:
Fixes #474 

Was I meant to swap some translations around also? 🤔


mark-prins (approved):
🎉  perfect! I wasn't thinking to move translations around - mostly they're sensible where they are.. I think. should probably review at some point, but not feeling the need to do it right now

PR #493, opened 2021-11-25
#448 inbound line edit 7

joshxg:
Was getting a bit big. Thought I'd stop here!

- I removed all the unsupported fields and therefore tabs - only leaving the batch and pricing tabs since they needed rewriting anyway, no point having commented & also wrong code
- Added a numeric input and currency input. For the currency I used a currency input package. This is wrapped and exported from common so if we want to remove it we can pretty easily (though maybe I should have no extended the props of the packaged component) - I know how annoying currency can be and so it saved a little bit of time and the result I think is pretty nice. 
- Made the inputs inputable - though they're only editable for newly added rows, for now (the new rows are given the updater function, passed in rows aren't, yet).
- I played a bit with some animations for switching between tabs. Thought they were OK, but a bit distracting and annoying.

### DIFF
--- a/packages/common/src/ui/components/inputs/CurrencyInput/CurrencyInput.tsx
+++ b/packages/common/src/ui/components/inputs/CurrencyInput/CurrencyInput.tsx
@@ -54,7 +54,7 @@ export const CurrencyInput: FC<CurrencyInputProps> = ({
   }, [value]);
 
   // Call the onChangeNumber prop when the buffer changes to a valid number, which
-  // is essentially in ever case except when the number ends in a decimal. If the user
+  // is essentially in every case except when the number ends in a decimal. If the user
   // loses focus when the buffer is in this invalid state, the currency input will
   // append two zeros so will always result in a valid number.
   useEffect(() => {
--- a/packages/common/src/ui/components/inputs/CurrencyInput/CurrencyInput.tsx
+++ b/packages/common/src/ui/components/inputs/CurrencyInput/CurrencyInput.tsx
@@ -1,0 +1,79 @@
+import React, { FC, useEffect } from 'react';
+import { styled } from '../../../../styles';
+import RCInput, {
+  CurrencyInputProps as RCInputProps,
+} from 'react-currency-input-field';
+
+interface CurrencyInputProps extends RCInputProps {
+  onChangeNumber: (value: number) => void;
+  maxWidth?: number;
+}
+
+const StyledCurrencyInput = styled(RCInput)(({ theme }) => ({
+  fontFamily: theme.typography.fontFamily,
+  height: 34.125,
+  borderRadius: '8px',
+  padding: '4px 8px',
+  backgroundColor: '#f2f2f5',
+  textAlign: 'right',
+  border: 'none',
+  '&:focus': {
+    outline: 'none',
+  },
+}));
+
+export const CurrencyInput: FC<CurrencyInputProps> = ({
+  allowNegativeValue = false,
+  prefix = '$',
+  decimalSeparator = '.',
+  decimalsLimit = 2,
+  allowDecimals = true,
+  decimalScale = 2,
+  value = 0,
+  onChangeNumber,
+  maxWidth,
+  ...restOfProps
+}) => {
+  // Buffer the internal input value so we can account for the decimal separator
+  // which, when input will result in the input being NaN (e.g. '42.') when cast
+  // to a number. In this case, we want to keep the value as '42.' in the input
+  // but not persisted. Though, whenever I have an uncontrolled component I
+  // generally will always regret it, so sync the passed value up with the buffer.
+  const [buffer, setBuffer] = React.useState<string | undefined>(String(value));
+  useEffect(() => {
+    if (value !== Number(buffer)) {
+      setBuffer(String(value));
+    }
+  }, [value]);
+
+  // Call the onChangeNumber prop when the buffer changes to a valid number, which
+  // is essentially in ever case except when the number ends in a decimal. If the user
+  // loses focus when the buffer is in this invalid state, the currency input will
+  // append two zeros so will always result in a valid number.
+  useEffect(() => {
+    if (buffer == null) {
+      return onChangeNumber(0);
+    }
+    if (!Number.isNaN(buffer)) {
+      return onChangeNumber(Number(buffer));
+    }
+  }, [buffer]);
+
+  return (
+    <StyledCurrencyInput
+      sx={{ maxWidth }}
+      value={buffer}
+      onValueChange={newValue => {
+        setBuffer(newValue);
+      }}
+      allowNegativeValue={allowNegativeValue}
+      prefix={prefix}
+      decimalSeparator={decimalSeparator}
+      decimalsLimit={decimalsLimit}
+      allowDecimals={allowDecimals}
+      decimalScale={decimalScale}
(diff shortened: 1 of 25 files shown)


packages/common/src/ui/layout/tables/components/Cells/CurrencyInputCell/CurrencyInputCell.tsx
```diff
@@ -0,0 +1,31 @@
...
+type CP<T> = CellProps<R<T>>;
+
+export const CurrencyInputCell = <T extends DomainObject>({
+  rowData,
+  column,
+}: CP<T>): React.ReactElement<CP<T>> => {
+  const [buffer, setBuffer] = React.useState(Number(column.accessor(rowData)));
+
+  const noop = () => {};
+  const updater = useDebounceCallback(rowData.update ?? noop, [rowData], 250);
```
joshxg, line 19: im not sure about the debounce. It's helpful for not going crazy on the rendering but does it actually make things seem laggier?

packages/common/src/ui/layout/tables/components/Cells/CurrencyInputCell/CurrencyInputCell.tsx
```diff
@@ -0,0 +1,31 @@
...
+  return (
+    <CurrencyInput
+      maxWidth={column.width}
+      value={buffer}
+      onChangeNumber={newNumber => {
+        setBuffer(newNumber);
+        updater(String(column.key), String(newNumber));
+      }}
+    />
+  );
```
joshxg, line 30: maybe it's preferrable that there is no $ prefix in a cell in a table?

packages/common/src/ui/components/inputs/CurrencyInput/CurrencyInput.tsx
```diff
@@ -0,0 +1,86 @@
...
+  // generally will always regret it, so sync the passed value up with the buffer.
+  const [buffer, setBuffer] = React.useState<string | undefined>(String(value));
+  useEffect(() => {
+    if (value !== Number(buffer)) {
+      setBuffer(String(value));
+    }
+  }, [value]);
+
+  // Call the onChangeNumber prop when the buffer changes to a valid number, which
+  // is essentially in ever case except when the number ends in a decimal. If the user
```
mark-prins, line 57: :nit no need to change! 
```suggestion
  // is essentially in every case except when the number ends in a decimal. If the user
```

packages/common/src/ui/components/inputs/CurrencyInput/CurrencyInput.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+interface CurrencyInputProps extends RCInputProps {
+  onChangeNumber: (value: number) => void;
+  maxWidth?: number;
+}
+
+// TODO: It would be nice if we were to just use the BasicTextInput or
+// another MUI text input rather than trying to recreate the style so that
+// it could stay in sync with style updated.
+const StyledCurrencyInput = styled(RCInput)(({ theme }) => ({
```
mark-prins, line 15: given the idea of instead using `currency.js` or similar, maybe not an issue. One thought was to have these styles in a common location and share them across various components

packages/common/src/ui/components/inputs/CurrencyInput/CurrencyInput.tsx
```diff
@@ -0,0 +1,86 @@
...
+  },
+}));
+
+// TODO: I think the implementation of this could be swapped out by making a
+// headless component that uses something like currency.js in a custom hook
+// to manage the currency value. Then you could use whatever input component
+// you liked. That would just take a little more time than I have right now!
+export const CurrencyInput: FC<CurrencyInputProps> = ({
+  allowNegativeValue = false,
+  prefix = '$',
```
mark-prins, line 34: I was wondering about the `$` sign somewhere else and was wondering how to get the correct locale for `intl` to format the currency. We are only using `en` which isn't enough. Perhaps though, we could just pull it from browser settings - the `i18n` returns full browser locale (such as `en-NZ`) and that would work ( although IIRC `en-NZ` wasn't supported.. there were a bunch of locales that intl falls down on. that aside, the approach might work 🤔  )

packages/common/src/ui/layout/tables/components/Cells/CurrencyInputCell/CurrencyInputCell.tsx
mark-prins: I'm not noticing any lag myself

packages/common/src/ui/layout/tables/components/Cells/CurrencyInputCell/CurrencyInputCell.tsx
mark-prins: I think it looks great - might not suit every client though: Ivory Coast are using `XOF` which takes up a lot of space and I've had to use interesting fonts to get rupees and euros working elsewhere

packages/host/package.json
```diff
@@ -29,11 +29,9 @@
   },
   "dependencies": {
     "@fontsource/inter": "^4.5.0",
-    "@openmsupply-client/catalogue": "^0.0.0",
```
mark-prins, line 32: Not sure why these are removed

mark-prins (approved):
👍  Great!

Like the currency inputs - and while I don't like having stuff removed.. simplifying the tabs probably helps people reviewing this modal focus on general functionality so is helpful.

Side note: wondering if a background colour might help group the tab content and show that batch & line are constant across all the tabs?

packages/common/src/ui/layout/tables/components/Cells/CurrencyInputCell/CurrencyInputCell.tsx
joshxg: Yeah I did think that XOF would be a bit big. Thinking in the column header might be better 🤔

packages/common/src/ui/components/inputs/CurrencyInput/CurrencyInput.tsx
joshxg: I think I was thinking similarly. Use something like `useMoney()` and also the `useFormatNumber()` (I should have been using it for all decimals actually, oops) - currency could use the `useI18N` hook to get the current language and use that..

packages/host/package.json
joshxg: Ah yes. They were the 'container' packages which were removed previously, I just had forgotten to remove them here. When I `yarn add`'d, I noticed as it bugged out

packages/common/src/ui/components/inputs/CurrencyInput/CurrencyInput.tsx
joshxg: I wasn't sure about it.. generally I think most things can jsut use `BasicTextInput` and extend it - unfortuantely this package didn't have an `InputComponent` prop to be able to use it which would have been good. Thought to leave it here (mainly because I wouldn't know where to put them otherwise!) for now... 🤷

packages/common/src/ui/components/inputs/CurrencyInput/CurrencyInput.tsx
joshxg: Ah. I just clicked that language and currency aren't a one to one relationship so maybe that doesn't work 🤔

Issue #512, opened 2021-11-26
Update graphql schema


joshxg:
The latest development branch has some new types for locations which we are behind on

PR #516, opened 2021-11-26
#512 Updated graphql schema

joshxg:
Just took the schema and added only the dashboard things. All the other additions are gone. Easier to update it 🤷

### DIFF
--- a/packages/common/src/schema.graphql
+++ b/packages/common/src/schema.graphql
@@ -6,6 +6,11 @@ schema {
 # Directs the executor to query only when the field exists.
 directive @ifdef on FIELD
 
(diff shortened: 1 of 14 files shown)


packages/common/src/schema.graphql
```diff
@@ -633,6 +648,13 @@ type Mutations {
   ): BatchOutboundShipmentResponse!
 }
 
+# ISO 8601 calendar date without timezone.
```
mark-prins, line 651: ah thanks!

mark-prins (approved):
👍  Nice! all looks reasonable to me, and is working well

Issue #513, opened 2021-11-26
Update mock server with requisition related data


joshxg:
Want to start working on requisitions soon. So, add requisition based things to the mock server!

Only really need this to work with the mock worker, I think. However, if it works for the apollo server also it will generally make the integration easier, maybe. Not sure what would be the most efficient use of time - spending some time updating the apollo server to *maybe* save time on the integration later, or just spending all that time on the integration in one go 🤔 

I also am not sure about program based requisitions. Right now i'm thinking to keep it simple with just a general order

PR #518, opened 2021-11-27
#513 mock server requisition data 1

joshxg:
#513 part 1!

So far this is only the requisitions themselves and not any lines. A lot of boiler plate! Lines will simply be copy and pasting pretty much all of this code and replacing CustomerRequisition with CustomerRequisitionLine.

What are the chances we end up using Request and Response rather than supplier and customer? 100% I'd say.

### DIFF
--- a/packages/mock-server/src/api/resolvers.ts
+++ b/packages/mock-server/src/api/resolvers.ts
@@ -1,4 +1,9 @@
 import {
+  UpdateRequisitionInput,
+  InsertRequisitionInput,
+  DeleteRequisitionInput,
+} from './../../../common/src/types/schema';
+import {
   ResolvedItem,
   ResolvedInvoice,
   ResolvedStockLine,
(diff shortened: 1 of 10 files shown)


packages/mock-server/src/api/resolvers.ts
```diff
@@ -95,6 +97,59 @@ const createListResponse = <T>(
 });
 
 export const ResolverService = {
+  requisition: {
+    get: {
+      byId: (id: string): ResolvedRequisition => {
```
mark-prins, line 102: I'm not seeing what's wrong - but I'm getting undefined for the `id` if I try to query a specific requisition 🤷

mark-prins (commented):
👍  looks good!

mark-prins (approved):
oops! that last one should have been an approval.

packages/mock-server/src/api/resolvers.ts
joshxg: Ah yes I see the issue, sorry - I'll fix that in a later PR

PR #519, opened 2021-11-27
#513 mock server requisition data 2

### LINKED ISSUE #513: Update mock server with requisition related data

joshxg:
Fixes #513 

All the line stuff for reqs ... !

- Mutations for lines
- Adding lines to requisitions

All the values quantities are just floats because 1) I need to remember what they all do.. 2) would rather just do a new issue updating the values rather than jam it into this one 3) would have to have proper statuses and dates on all the transaction stuff to be able to count stock on hand and things properly - should be fine - will wait till some sort of ui gets started before doing more

### DIFF
--- a/packages/common/src/schema.graphql
+++ b/packages/common/src/schema.graphql
@@ -656,10 +656,22 @@ type Mutations {
   deleteSupplierRequisition(
     input: DeleteSupplierRequisitionInput!
   ): DeleteSupplierRequisitionResponse!
+  updateSupplierRequisitionLine(
+    input: UpdateSupplierRequisitionLineInput!
+  ): UpdateSupplierRequisitionLineResponse!
+  insertSupplierRequisitionLine(
+    input: InsertSupplierRequisitionLineInput!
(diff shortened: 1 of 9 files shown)


mark-prins (approved):
👍  Sweet!
Well, I've read it all.. and the code runs. I figure that I'll test it more carefully once the UI comes together. I've only done a little so far. 
Couple of things:
- I realise that I still don't know what an 'imprest' is
- the error messaging is showing up.. though I still feel fine about deferring that

joshxg:
> I realise that I still don't know what an 'imprest' is

Me either. I just randomly picked fields from the spreadsheet if I'm honest

Issue #514, opened 2021-11-26
Update mock server with stocktake data


joshxg:
Want to start working on requisitions soon. So, add stocktake based things to the mock server!

Only really need this to work with the mock worker, I think. However, if it works for the apollo server also it will generally make the integration easier, maybe. Not sure what would be the most efficient use of time - spending some time updating the apollo server to maybe save time on the integration later, or just spending all that time on the integration in one go 🤔

PR #522, opened 2021-11-28
#514 mock server stocktake line data

joshxg:
#514 

Adds very basic stock take lines and mutations

### DIFF
--- a/packages/common/src/schema.graphql
+++ b/packages/common/src/schema.graphql
@@ -707,7 +707,10 @@ type Mutations {
   batchStocktake(
     insertStocktakes: [InsertStocktakeInput!]
     updateStocktakes: [UpdateStocktakeInput!]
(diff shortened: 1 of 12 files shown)


packages/common/src/schema.graphql
```diff
@@ -1295,6 +1298,60 @@ type StocktakeNode {
...
+  itemName: String!
+  itemCode: String!
+  batch: String
+  expiryDate: NaiveDate
+  snapshotNumPacks: Int
+  snapshotPackSize: Int
+  countedNumPacks: Int
+  sellPricePerPack: Float
+  costPricePerPack: Float
+}
```
mark-prins, line 1315: I think we might need location ( `locationId`? ) and comment and possibly also the `reason` used by Ivory Coast

mark-prins (approved):
🎉  LGTM! thanks! another big load of goodness

packages/common/src/schema.graphql
joshxg: Yup I think there's a fair amount left out .. really just hoping to get all the bulk annoying stuff done so we can make some small more targeted issues

PR #527, opened 2021-11-28
#525 supplier requisition list view

### LINKED ISSUE #525: Add supplier requisition list view
Epic: #99

- Add the list view to supplier requisition

joshxg:
Fixes #525 

Just a basic supplier requisition view - a lot of copy paste suprisngly???

### DIFF
--- a/packages/common/src/intl/locales/en/common.json
+++ b/packages/common/src/intl/locales/en/common.json
@@ -75,6 +75,8 @@
   "label.expiry": "Expiry",
   "label.general": "General",
   "label.goods-receipt": "Goods receipt",
+  "label.requisition-number": "Reqnum",
+  "label.new-requisition": "New Requisition",
   "label.hold": "Hold",
   "label.id": "ID",
   "label.inbound-shipment": "Inbound shipment",
--- a/packages/requisitions/src/SupplierRequisition/ListView/ListView.tsx
+++ b/packages/requisitions/src/SupplierRequisition/ListView/ListView.tsx
@@ -54,7 +54,7 @@ export const SupplierRequisitionListView: FC = () => {
   return (
     <>
       <NameSearchModal
-        type="customer"
+        type="supplier"
         open={open}
         onClose={() => setOpen(false)}
         onChange={async name => {
--- a/packages/requisitions/src/SupplierRequisition/ListView/ListView.tsx
+++ b/packages/requisitions/src/SupplierRequisition/ListView/ListView.tsx
@@ -1,0 +1,108 @@
+import React, { FC, useState } from 'react';
+import { useNavigate } from 'react-router';
+import {
+  DataTable,
+  useColumns,
+  useListData,
+  TableProvider,
+  createTableStore,
+  useNotification,
+  generateUUID,
+  useOmSupplyApi,
+} from '@openmsupply-client/common';
+import { NameSearchModal } from '@openmsupply-client/system/src/Name';
+import { Toolbar } from './Toolbar';
+import { AppBarButtons } from './AppBarButtons';
+import { getSupplierRequisitionListViewApi } from './api';
+import { RequisitionRow } from '../../types';
+
+export const SupplierRequisitionListView: FC = () => {
+  const navigate = useNavigate();
+  const { error } = useNotification();
+  const { api } = useOmSupplyApi();
+
+  const {
+    totalCount,
+    data,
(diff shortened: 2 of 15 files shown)


packages/common/src/intl/locales/en/common.json
```diff
@@ -75,6 +75,8 @@
   "label.expiry": "Expiry",
   "label.general": "General",
   "label.goods-receipt": "Goods receipt",
+  "label.requisition-number": "Reqnum",
```
mark-prins, line 78: I hadn't actually noticed `Invnum` until now 🙄 
Could possibly just have `Number` or `Req #` ...though now is probably not the time to worry about column names. I'm sure there will be opinions available if we want some 😄

packages/requisitions/src/SupplierRequisition/ListView/ListView.tsx
```diff
@@ -0,0 +1,108 @@
...
+    { onChangeSortBy, sortBy },
+    [sortBy]
+  );
+
+  const [open, setOpen] = useState(false);
+
+  return (
+    <>
+      <NameSearchModal
+        type="customer"
```
mark-prins, line 57: supplier?

packages/requisitions/src/SupplierRequisition/ListView/Toolbar.tsx
```diff
@@ -0,0 +1,73 @@
...
+    selectedRows: Object.keys(state.rowState)
+      .filter(id => state.rowState[id]?.isSelected)
+      .map(selectedId => data?.find(({ id }) => selectedId === id))
+      .filter(Boolean) as RequisitionRow[],
+  }));
+
+  const deleteAction = () => {
+    if (selectedRows && selectedRows?.length > 0) {
+      onDelete(selectedRows);
+      success(`Deleted ${selectedRows?.length} invoices`)();
```
mark-prins, line 34: :nit translations for both of these

packages/requisitions/src/SupplierRequisition/ListView/Toolbar.tsx
```diff
@@ -0,0 +1,73 @@
...
+    <AppBarContentPortal
+      sx={{
+        paddingBottom: '16px',
+        flex: 1,
+        justifyContent: 'space-between',
+        display: 'flex',
+      }}
+    >
+      <SearchBar
+        placeholder="Search by comment..."
```
mark-prins, line 59: translation?

mark-prins (approved):
👍  Woo!! Looking good 😁

packages/requisitions/src/SupplierRequisition/ListView/Toolbar.tsx
joshxg: Ah yes. I think i'll make an issue (I need to do the invoices also) - I realised I was also a little unsure of what to do for the new translation file - since we have `distribution`, do you think`replenishment` and then add `deleted X invoices` to both distribution and replenishment? 🤔

mark-prins: yeah, that sounds good. I was in between two approaches: having anything that is used in more than one area put in a shared file - and keeping as much as possible in each separate area. For this one - the line is pretty specific, so I would duplicate it, and have in both distribution and replenishment. That's where I'm leaning now - all the really general stuff in common, and specific things in the file for each area, with duplication as necessary 🤷

joshxg: Ah great that sounds good. One other thought I had was that the columns themselves use a locale key without the namespace specified. So we could either somehow put into the column definition a namespace or use a static namespace for all columns.. something just like a `datatable` namespace which the `column` component uses to translate keys for example

mark-prins: I do like having the locale key on the column, rather than a string. Could have a `datatable` namespace, though I like the idea of having all the translations for a functional area together in the same file. Could pass in the namespace to the column somehow though then the component is tied to a specific implementation of localisation. Could do something hokey with the locale key, say `[namespace]_[localy-key]` but that is bad on even more levels.

I would go with (a) a `datatable` namespace or (b) keep them all in `common` namespace - but really don't mind whatever way. Can't see an ideal option.

PR #530, opened 2021-11-28
Add inventory package

### LINKED ISSUE #529: Add a package for stock takes
I am not sure what this package should be called or what should go in it.

I'm going to take a swing and say `inventory` and stocktakes and inventory adjustments can live there - possibly builds/repacks and things 🤔  

Easy enough to change if it doesn't make sense!

joshxg:
Fixes #529 

- While doing this I noticed the `Stock` package and wondered if I put this in the wrong place. Hm.

### DIFF
--- a/packages/inventory/src/index.ts
+++ b/packages/inventory/src/index.ts
@@ -1,0 +1,10 @@
+export {
+  ListView as SupplierRequisitionListView,
+  DetailView as SupplierRequisitionDetailView,
+} from './Stocktake';
+export {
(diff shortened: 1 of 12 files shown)


packages/inventory/src/index.ts
```diff
@@ -0,0 +1,10 @@
+export {
+  ListView as SupplierRequisitionListView,
```
mark-prins, line 2: 😕  Just reusing the types, right?

mark-prins (approved):
👍  awesome! another one 😁

PR #535, opened 2021-11-28
#533 supplier requisition detail view

### LINKED ISSUE #533: Supplier requisition detail view
Epic #99 

- Add a basic implementation of the detail view for supplier requisitions

joshxg:
Fixes #533 

This adds mainly copy/paste/boiler plate for supplier requisitions. Was going to continue and setup the mutations but thought to stop here.. it's big enough 🙄 

Think we will likely need a way to force a line break in the column headers.

In general this should (I will regret this) not take much more work to complete. Other than, of course, the charts 🤣  - but only have simple lines with a simple input is nice!


packages/requisitions/src/SupplierRequisition/DetailView/DetailView.tsx
```diff
@@ -0,0 +1,110 @@
...
+          accessor: line => {
+            const threeMonthsStock = line.monthlyConsumption * 3;
+            const diff = threeMonthsStock - line.previousStockOnHand;
+            if (diff > 0) {
+              return `${diff.toFixed(2)} (${Math.floor(
+                diff / line.monthlyConsumption
+              )} months)`;
+            } else {
+              return 0;
+            }
```
joshxg, line 71: made an assumption/placeholder that the 'target' is 3 months worth of stock

packages/requisitions/src/SupplierRequisition/DetailView/DetailView.tsx
```diff
@@ -0,0 +1,110 @@
...
+          },
+        },
+      ],
+      ['forecastMethod', { accessor: () => 'AMC' }],
```
joshxg, line 75: Not sure what the forecast method is supposed to be in the ui designs or where it supposed to come from, so just a placeholder for now

packages/requisitions/src/SupplierRequisition/DetailView/DetailView.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  const onChangeSortBy = () => {};
+
+  return { draft, save, dispatch, onChangeSortBy, sortBy: state.sortBy };
+};
+
+export const DetailView: FC = () => {
+  const { draft, save, onChangeSortBy, sortBy } = useDraftSupplierRequisition();
+
+  const columns = useColumns<RequisitionLine>(
```
joshxg, line 43: Not actually sure if I was meant to be following the ui columns or if they were just for the designs - I did it anyway as they seemed nice - though, I don't claim for them to be 100% complete (so I haven't done translations etc)

packages/requisitions/src/SupplierRequisition/DetailView/DetailView.tsx
mark-prins: we were going to remove that column, since we aren't using per-item forecasting methods at the moment.. all values would just be `AMC` most likely 😁

packages/requisitions/src/SupplierRequisition/DetailView/DetailView.tsx
mark-prins: it's the `max MOS` field (now MOS) which isn't shown yet. defaults to `3` - but is a property of the requisition

packages/requisitions/src/SupplierRequisition/DetailView/DetailView.tsx
```diff
@@ -0,0 +1,108 @@
...
+              )} months)`;
+            } else {
+              return 0;
+            }
+          },
+        },
+      ],
+      ['forecastMethod', { accessor: () => 'AMC' }],
+      'requestedQuantity',
+      ['comment', { width: 150 }],
```
mark-prins, line 75: given how much needs to be squeezed on to this table - what about we sneak in the comment 'speech bubble' column?

mark-prins (approved):
🎉  looks awesome!!

Am wondering about reducing the font size in the tables, and doing something with the column headers 🤔

packages/requisitions/src/SupplierRequisition/DetailView/DetailView.tsx
joshxg: Of course, that makes sense!

packages/requisitions/src/SupplierRequisition/DetailView/DetailView.tsx
joshxg: Ah yes, cool!

packages/requisitions/src/SupplierRequisition/DetailView/DetailView.tsx
joshxg: Ah yes. Will need to make a new column for this as the current note is not handling edits.

Re font size. I'm wondering if a `style` prop in the column definition might be useful 🤔

PR #537, opened 2021-11-29
Add updated mutations for supplier requisitions

### LINKED ISSUE #536: Supplier requisition detail view - mutations
Hook up the `onUpdate`/`save` on the supplier requisition detail view

joshxg:
FIxes #536

- Adds the upsert for supplier requisition detail view
- Also adds updating of a few requisition fields

### DIFF
--- a/packages/common/src/operations.graphql
+++ b/packages/common/src/operations.graphql
@@ -218,7 +218,7 @@ query requisitions($params: RequisitionListParameters) {
         id
         comment
         orderDate
-        otherPartyReference
+        theirReference
         requisitionNumber
         status
         otherPartyName
@@ -292,15 +292,13 @@ query requisition($id: String!) {
       id
       orderDate
       comment
-      otherPartyReference
+      theirReference
       type
       requisitionNumber
       thresholdMOS
       maxMOS
       status
       otherPartyId
-      otherPartyName
-      storeId
       lines {
         __typename
         ... on ConnectorError {
@@ -315,6 +313,7 @@ query requisition($id: String!) {
             itemName
             itemCode
             itemUnit
+            itemId
             comment
             monthlyConsumption
             monthsOfSupply
@@ -588,10 +587,10 @@ query itemsWithStockLines(
               itemId
               packSize
               sellPricePerPack
-              storeId
               totalNumberOfPacks
               onHold
               note
+              storeId
             }
             totalCount
           }
--- a/packages/common/src/schema.graphql
+++ b/packages/common/src/schema.graphql
@@ -1274,17 +1274,15 @@ type UpdateOutboundShipmentResponseWithId {
 type RequisitionNode {
   id: String!
 
-  storeId: String!
-
+  color: String
   orderDate: String
   comment: String
-  otherPartyReference: String
+  theirReference: String
   type: RequisitionNodeType
   requisitionNumber: Int!
   thresholdMOS: Int
   maxMOS: Int
   status: SupplierRequisitionNodeStatus!
-
   lines: RequisitionLinesResponse!
 
   otherParty: NameResponse!
@@ -1363,6 +1361,7 @@ enum StocktakeNodeStatus {
 
 type RequisitionLineNode {
   id: String!
+  itemId: String!
   itemName: String
   itemCode: String
   itemUnit: String
@@ -1446,8 +1445,8 @@ input UpdateSupplierRequisitionInput {
   orderDate: String
   nameId: String
   comment: String
-  otherPartyReference: String
-  type: RequisitionNodeType
+  theirReference: String
+  color: String
 }
 
 input UpdateSupplierRequisitionLineInput {
@@ -1455,12 +1454,9 @@ input UpdateSupplierRequisitionLineInput {
   itemName: String
   itemCode: String
   itemUnit: String
-
   comment: String
-
   monthlyConsumption: Float
   monthsOfSupply: Float
-
   supplyQuantity: Float
   openingQuantity: Float
   issuedQuantity: Float
@@ -1515,7 +1511,7 @@ input InsertSupplierRequisitionInput {
   orderDate: String
   otherPartyId: String!
   comment: String
-  otherPartyReference: String
+  theirReference: String
   type: RequisitionNodeType
 }
 
@@ -1690,8 +1686,8 @@ input UpdateCustomerRequisitionInput {
   orderDate: String
   nameId: String
   comment: String
-  otherPartyReference: String
-  type: RequisitionNodeType
+  theirReference: String
+  color: String
 }
 
 input InsertCustomerRequisitionInput {
@@ -1699,7 +1695,7 @@ input InsertCustomerRequisitionInput {
   orderDate: String
   otherPartyId: String!
   comment: String
-  otherPartyReference: String
+  theirReference: String
   type: RequisitionNodeType
 }
 
--- a/packages/common/src/types/schema.ts
+++ b/packages/common/src/types/schema.ts
@@ -404,7 +404,7 @@ export type InsertCustomerRequisitionInput = {
   id: Scalars['String'];
   orderDate?: Maybe<Scalars['String']>;
   otherPartyId: Scalars['String'];
-  otherPartyReference?: Maybe<Scalars['String']>;
+  theirReference?: Maybe<Scalars['String']>;
   type?: Maybe<RequisitionNodeType>;
 };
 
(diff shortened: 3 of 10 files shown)


packages/common/src/operations.graphql
```diff
@@ -218,7 +218,7 @@ query requisitions($params: RequisitionListParameters) {
         id
         comment
         orderDate
-        otherPartyReference
+        theirReference
```
joshxg, line 221: more consistent

packages/common/src/schema.graphql
```diff
@@ -1363,6 +1361,7 @@ enum StocktakeNodeStatus {
 
 type RequisitionLineNode {
   id: String!
+  itemId: String!
```
joshxg, line 1364: seems right

packages/common/src/types/schema.ts
```diff
@@ -1592,12 +1593,12 @@ export type TokenExpired = RefreshTokenErrorInterface & {
...
 
 export type UpdateCustomerRequisitionInput = {
+  color?: Maybe<Scalars['String']>;
   comment?: Maybe<Scalars['String']>;
   id: Scalars['String'];
   nameId?: Maybe<Scalars['String']>;
   orderDate?: Maybe<Scalars['String']>;
-  otherPartyReference?: Maybe<Scalars['String']>;
-  type?: Maybe<RequisitionNodeType>;
+  theirReference?: Maybe<Scalars['String']>;
```
joshxg, line 1601: Jeez. Lots of changed lines for such a simple change

packages/common/src/types/schema.ts
```diff
@@ -2002,7 +2003,7 @@ export type ItemsWithStockLinesQueryVariables = Exact<{
...
...abaseError', description: string, fullError: string } | { __typename: 'PaginationError', description: string, rangeError: { __typename?: 'RangeError', description: string, field: RangeField, max?: number | null | undefined, min?: number | null | undefined } } } | { __typename: 'ItemConnector', totalCount: number, nodes: Array<{ __typename: 'ItemNode', code: string, id: string, isVisible: boolean, name: string, unitName?: string | null | undefined, availableBatches: { __typename: 'ConnectorError', error: { __typename: 'DatabaseError', description: string, fullError: string } | { __typename: 'PaginationError', description: string, rangeError: { __typename?: 'RangeError', description: string, field: RangeField, max?: number | null | undefined, min?: number | null | undefined } } } | { __typename: 'StockLineConnector', totalCount: number, nodes: Array<{ __typename: 'StockLineNode', availableNumberOfPacks: number, batch?: string | null | undefined, costPricePerPack: number, expiryDate?: any | null | undefined, id: string, itemId: string, packSize: number, sellPricePerPack: number, totalNumberOfPacks: number, onHold: boolean, note?: string | null | undefined, storeId: string }> } }> } };
```
joshxg, line 2006: 🤔 excuse me

packages/requisitions/src/SupplierRequisition/DetailView/api.ts
```diff
@@ -39,6 +49,43 @@ const linesGuard = (
   throw new Error('Unknown');
 };
 
+const createUpdateSupplierRequisitionInput = (
```
joshxg, line 52: In general this is just a copy paste of the same thing from the inbound api replacing 'InboundShipment' with 'SupplierRequisition' 😁

packages/requisitions/src/SupplierRequisition/DetailView/reducer.ts
```diff
@@ -1,35 +1,55 @@
...
+      type: RequisitionActionType.Update,
+      payload: { key, value },
+    };
+  },
+  updateOtherParty: (value: Name): RequisitionAction => {
+    return {
+      type: RequisitionActionType.UpdateOtherParty,
+      payload: { value },
+    };
+  },
```
joshxg, line 34: Spent a bit of time trying to iterate on the invoices solution.

I spent a bit too much time trying to come up with some types which could allow for a single `update` function, where the `key` and `value` were typed correctly... 

I created the union on types with:

```
export type UpdateActions<ActionType, EntityType> = {
  [K in keyof EntityType]: {
    type: ActionType;
    payload: { key: K; value: EntityType[K] };
  };
}[keyof EntityType];
```

It gets close, but then in the reducer the key and values are still a union of all types, of course, 

i.e. you still can't do `state.draft[action.payload.key] = action.payload.value` of course 😢 

So I wasn't sure what to do - think just using a unique action per field might just be the way to go. The motivation to try make a nice generic fn was for plugins - if they're passed the draft they can just update it

packages/requisitions/src/utils.ts
```diff
@@ -16,10 +17,21 @@ export const placeholderSupplierRequisition: SupplierRequisition = {
...
-  storeId: '',
+  update: () => {
+    throw new Error(
+      'Placeholder callback update has been triggered. This should never happen!'
+    );
+  },
+  updateOtherParty: () => {
+    throw new Error(
+      'Placeholder callback updateOtherParty has been triggered. This should never happen!'
+    );
```
joshxg, line 28: Also think I like this but not sure.

The idea being that the draft is created 'properly' after merging the server data in with the draft where the callbacks are created - if this doesn't happen well you shouldn't be trying to update the draft!

Don't usually do things like this ("should not happen or throw") so not sure 🤔

packages/common/src/operations.graphql
mark-prins: I prefer it - but isn't `otherPartyReference` more consistent when we have `otherPartyName`?

packages/requisitions/src/SupplierRequisition/DetailView/reducer.ts
```diff
@@ -17,21 +16,37 @@ export interface SupplierRequisitionStateShape {
   sortBy: SortBy<SupplierRequisitionLine>;
 }
 
+const RequisitionActionCreator = {
```
mark-prins, line 19: I had to add some imports to this file for the update to work - is there a missing commit?

packages/requisitions/src/SupplierRequisition/DetailView/reducer.ts
mark-prins: argh. yes, I spent a while trying as well - will have another go later, the goal of having a generic function for plugins is worth it. that would be awesome if possible

packages/requisitions/src/utils.ts
mark-prins: could be useful in development phase, perhaps when moving into production a bugsnag notification or other log entry would be more helpful.

packages/requisitions/src/SupplierRequisition/DetailView/Toolbar.tsx
```diff
@@ -61,7 +61,7 @@ export const Toolbar: FC<ToolbarProps> = ({ draft }) => {
                     disabled={!isRequisitionEditable(draft)}
                     value={draft.otherParty}
                     onChange={name => {
-                      draft.update?.('otherParty', name);
+                      draft.updateOtherParty(name);
```
mark-prins, line 64: the toolbar also needs the MOS / Stock Target field and the view filters. and, now that I look at it.. no option to delete lines

mark-prins (approved):
👍  Fantastic! coming together nicely!

Couple of small things and the general desire to get away from the SHOUTY CASE STATUS values, but 🤷 . Oh and the column headers on my laptop screen 😆  there's a lot of things crammed in!

packages/common/src/operations.graphql
joshxg: 😢 

`theirReference` is on invoices so consistent in relation with that. Maybe it should be `otherPartyReference` for all of them 🤔

packages/requisitions/src/SupplierRequisition/DetailView/reducer.ts
joshxg: Yeah. Does depend on what you allow plugins to do though. If they're only editing some sort of `plugin` object, then it might not be as useful.

Other option could be to write a function to generate `updateFieldName` methods and types which might be easier 🤔

packages/requisitions/src/SupplierRequisition/DetailView/reducer.ts
joshxg: ah nuts sorry about that

packages/requisitions/src/SupplierRequisition/DetailView/Toolbar.tsx
joshxg: Ah yes another question area. I think it might be good if I could meet with you or someone from the ui team and go through supplier requisitions to make sure I understand it all.

I wasn't sure if the option not to delete lines was for program requisitions or general or something, maybe

Mobile also has some other bulk actions (Create an auto order, set requested to forecast) that I thought might be nice to haves?

mark-prins: yeah - I wasn't sure about requisitions, the initial discussions were only between Richard & Manu. Just noticed that there's no delete on the zeplin designs, though I suspect that's an oversight. As I understand it, only program reqs would be restricted in that way.

Would be great to bring the other actions through from mobile!

PR #542, opened 2021-11-30
basic cell text overflow

### LINKED ISSUE #515: Plain/Basic DataTable cell word wrapping is off
I think I broke this while trying to *fix* the data table layout..

Must have forgot to actually test this worked.

I removed word wrap from: https://github.com/openmsupply/openmsupply-client/blob/4358ad024f6e1e4335649b8b2e2d8b1021c287d0/packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx#L63
and added it to: https://github.com/openmsupply/openmsupply-client/blob/4358ad024f6e1e4335649b8b2e2d8b1021c287d0/packages/common/src/ui/layout/tables/components/index.tsx#L17

and its broken!

mark-prins:
Fixes #515 
A small tidy - the span needs to be a block element in order to have a defined width, which is required for overflow. When it takes the width from the `td` that's the full width, and the available is less the padding, so have manually adjusted. Not great, but working ok.

### DIFF
--- a/packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
+++ b/packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
@@ -5,7 +5,7 @@ import { Column } from '../../columns/types';
 import { DomainObject } from '../../../../../types';
 import { useExpanded, useDisabled } from '../../context';
 import { Collapse } from '@mui/material';
-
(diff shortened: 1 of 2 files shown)


joshxg (approved):
thats awesome, thanks.. TIL 😁

PR #544, opened 2021-11-30
#13 bundle size diff

### LINKED ISSUE #13: Bundle size diff github action
Maybe OTT right now, but maybe a gha like: https://github.com/marketplace/actions/bundle-size-diff might be good?

mark-prins:
Fixes #13 
Not strictly necessary, but thought it might be a good learning experience. who knows.. might be useful 🤷 

Will need to get the webpack change into main before it works.. without it, the `stats.json` isn't generated of course and so we can't compare

### DIFF
--- a/.github/workflows/bundle-size-diff.yaml
+++ b/.github/workflows/bundle-size-diff.yaml
@@ -1,7 +1,7 @@
 name: Bundle size diff
 
 on:
-  push:
+  pull_request:
 
 jobs:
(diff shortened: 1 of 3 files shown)


joshxg (approved):
Ohhh cool! this is gonna be awesome!!

I don't really know enough about the yaml to give any useful comments but will come back to it when I have more time to learn 😁

PR #550, opened 2021-12-02
#464 available quantity

### LINKED ISSUE #464: Add item: available quantity not shown next to item name
Previously when adding an item, the available quantity was shown to the right of the name of the item.
With the latest graphql changes this has been dropped - I think we can calculate the value based on stock lines and re-instate this

mark-prins:
Fixes #464 
I haven't taken into account the allocated but unsaved quantities when showing available quantity in the item list - currently the modal doesn't adjust the Available Quantity value when you allocate rows and that seems ok to me. Even so.. I've now changed the form, so that the available updates when you allocate. I like it - but what do you think?

### DIFF
--- a/packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
+++ b/packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
@@ -5,7 +5,6 @@ import { Column } from '../../columns/types';
 import { DomainObject } from '../../../../../types';
 import { useExpanded, useDisabled } from '../../context';
 import { Collapse } from '@mui/material';
-import { cellHorizontalPadding } from '../index';
 interface DataRowProps<T extends DomainObject> {
   columns: Column<T>[];
   onClick?: (rowData: T) => void;
@@ -41,7 +40,6 @@ export const DataRow = <T extends DomainObject>({
           alignItems: 'center',
           height: '40px',
           maxHeight: '45px',
-
           boxShadow: dense
             ? 'none'
             : 'inset 0 0.5px 0 0 rgba(143, 144, 166, 0.5)',
@@ -61,8 +59,8 @@ export const DataRow = <T extends DomainObject>({
                 justifyContent: 'flex-end',
                 overflow: 'hidden',
                 whiteSpace: 'nowrap',
-                paddingLeft: `${cellHorizontalPadding}px`,
-                paddingRight: `${cellHorizontalPadding}px`,
+                paddingLeft: '16px',
+                paddingRight: '16px',
                 ...(hasOnClick && { cursor: 'pointer' }),
                 flex: `${column.width} 0 auto`,
                 minWidth: column.minWidth,
--- a/packages/common/src/ui/layout/tables/components/index.tsx
+++ b/packages/common/src/ui/layout/tables/components/index.tsx
@@ -5,7 +5,6 @@ import { useTranslation, useFormatDate } from '../../../../intl';
 
 export * from './DataRow';
 export * from './Cells';
-export const cellHorizontalPadding = 16;
 
 export const BasicCell = <T extends DomainObject>({
   column,
@@ -15,16 +14,14 @@ export const BasicCell = <T extends DomainObject>({
   const d = useFormatDate();
 
   return (
-    <span
+    <div
       style={{
         overflow: 'hidden',
         textOverflow: 'ellipsis',
-        width: column.width - 2 * cellHorizontalPadding,
-        display: 'inline-block',
       }}
     >
       {column.formatter(column.accessor(rowData), { t, d })}
-    </span>
+    </div>
   );
 };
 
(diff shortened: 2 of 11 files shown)


packages/common/src/ui/layout/tables/components/index.tsx
```diff
@@ -15,16 +14,14 @@ export const BasicCell = <T extends DomainObject>({
   const d = useFormatDate();
 
   return (
-    <span
+    <div
```
mark-prins, line 17: offroad to fix this:

Screen Shot 2021-12-01 at 5 56 46 PM

packages/invoices/src/OutboundShipment/DetailView/modals/ItemDetailsForm.tsx
```diff
@@ -77,7 +78,7 @@ export const ItemDetailsForm: React.FC<ItemDetailsFormProps> = ({
       <ModalRow>
         <ModalLabel label={t('label.available')} />
         <NumericTextInput
-          value={availableQuantity}
+          value={Math.max(availableQuantity - allocatedQuantity, 0)}
```
mark-prins, line 81: it does mean that this can happen:
- available quantity = 100
- assign 500 to a placeholder line
- available : 0 allocated : 500

that makes sense to me, there's no more available. seems better than showing -400 available 🤷

joshxg (approved):
Being able to see the available quantity is super useful.. good to not have to select a bunch of items to finally find one with some stock 🤣  thanks!

I've got a couple already-existing-josh-messed-up-bugs, a suggestion and a comment, woo!

---------

There's a bug in `useBatchRows` where this happens:

```
return {
            ...batch,
            numberOfPacks: matchingInvoiceRow?.numberOfPacks ?? 0,
            availableNumberOfPacks:
              batch.availableNumberOfPacks +
              (matchingInvoiceRow?.numberOfPacks ?? 0),
          };
```

This only works when the line hasn't been saved. When it has been saved, the`batch.availableNumberOfPacks` has already been adjusted. So probably a new issue - but I think we need to store the `numberOfPacks` that the server has, and the `numberOfPacks` the client has on `OutboundShipmentItem` in separate fields so we can adjust if the client value differs from the server value, or something like that..?

for example:
This newly added item has 40 available:

image

When I allocate some it adjusts perfectly

image

When I reopen, the available has had the 2 packs added to it which throws everything off

image

----

Also when we're editing lines, I am not sure that we should be reducing `available` when entering values. I.e.  

This newly added item:
image

When I add some packs to a line 
image

Available reduces, which makes sense, but I think it makes *morer* sense/is more helpful would be if the value was just the *total amount available to allocate*, not including what you have allocated in your invoice. So when `allocated` equals `available`, there is none left to issue, but you can see the total amount able to be allocated in your invoice - it doesn't make sense to me that what you have in your invoice is 'not available' - I suspect this is a "me" thing.


-------

Aaand, the most annoying one which requires another bug fix  - the extra filter fn should not filter out 'deleted' items:

```
extraFilter={item => {
              const itemAlreadyInShipment = draft.items.some(
                ({ id, isDeleted }) => id === item.id && !isDeleted
              );
              return !itemAlreadyInShipment;
            }}
```

Then, If I delete a line which has had some stock assigned and then try to re-add it before saving, then the available doesn't account for the 'deleted' line

packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
```diff
@@ -61,8 +59,8 @@ export const DataRow = <T extends DomainObject>({
                 justifyContent: 'flex-end',
                 overflow: 'hidden',
                 whiteSpace: 'nowrap',
-                paddingLeft: `${cellHorizontalPadding}px`,
-                paddingRight: `${cellHorizontalPadding}px`,
+                paddingLeft: '16px',
+                paddingRight: '16px',
```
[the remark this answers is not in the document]
mark-prins: not required - explicitly setting the width on the span was causing an issue: if there was more space available than expected ( such as the items list, where the name column is 75px and there's only two columns, so the table spreads out ) the span was not growing. 

have reverted the change - this `16px` is what used to be there before I changed it.

mark-prins:
> I'm not sure that having the allocated quantity include the placeholder quantity makes sense to me

I see it as a total quantity, in which case it should include the placeholder quantity. I see what you mean though, if I take 'allocated' to mean the total which has been put aside for this invoice.. then it shouldn't include placeholder

I wonder if the total was labelled as 'requested' rather than 'allocated'? 🤔

mark-prins:
> `useBatchRows`

I'll leave that one for another issue, because I'm confused! As I see it, this is a problem before you save but you say it's only ok before you save 😕 . 

I have 4 packs on line 2, the server says 100 available packs:

Screen Shot 2021-12-02 at 5 39 07 PM

if I change to 5 packs for that line, click OK and edit that row again, without saving:

Screen Shot 2021-12-02 at 5 39 31 PM

I now have 101 packs available. If I save.. well, I get an error. ok, fixed that in 260441b3
The save is inconclusive, it only saved the first change and now won't update??

mark-prins:
Righto, I'll just merge this and leave the other issue. I've done enough mucking around :)

have fixed 
- the placeholder quantity isn't added to allocated
- allocated doesn't reduce available
- deleted lines can be re-added
- buttons are not disabled when you have zero quantity ( would have needed to update with the placeholder change, went with enabling instead )

in 0e6f4ad4

PR #553, opened 2021-12-02
#552 stale code

### LINKED ISSUE #552: Remove useless stories and old mock code
There's a few stories which I don't see value in within the `invoices` package - the `detail` and `list` views for outbound shipment

There's also a few files in the `mock-server` package which aren't used
- populate.ts`
- `full-schema.ts`
- some `.sqlite` files

joshxg:
Fixes #552 

- Just a few chores
- Removed a bunch of dead code
- There was a failing test due to forgetting to put the error code mock responses into the new handler code, so have added those
- Jest sets `NODE_ENV` to `'test'`, so the mocks weren't all running in the test env giving some warnings which were annoying


mark-prins (approved):
👍  thanks! very nice to have it tidied up

PR #598, opened 2021-12-03
#504 date pickers

### LINKED ISSUE #504: Date picker component & Date picker cell
We need a date picker in general (order dates etc for a requisition) - usually attached to a text input? Also a date picker for a cell for expiry dates

joshxg:
Fixes #504 

- Want to add some dates to the detail views so wanted to get this done. Somewhat basic but does the job
- Intention is that the base is used in variants like the expiry or maybe time variants (with cold chain n stuff)
- thought i'd just get the expiry one done while i'm here
- Wasn't sure where to put this. Thought about a pickers folder instead of `inputs`. but 🤷


packages/common/src/ui/components/inputs/TextInput/BasicTextInput.tsx
```diff
@@ -21,6 +21,8 @@ export const BasicTextInput: FC<StandardTextFieldProps> = React.forwardRef(
         disableUnderline: true,
         ...InputProps,
         sx: {
+          border: theme =>
+            error ? `2px solid ${theme.palette.error.main}` : 'none',
```
joshxg, line 25: I quite like this. Needs a nice 'invalid' message also though, I think.

packages/common/src/ui/components/inputs/DatePickers/DatePickerInput/DatePickerInput.tsx
```diff
@@ -0,0 +1,20 @@
...
+  onChange: (value: Date | null) => void;
+}
+
+export const DatePickerInput: FC<DatePickerInputProps> = ({
+  value,
+  onChange,
+}) => {
+  return (
+    <BaseDatePickerInput
+      inputFormat="dd/MM/yyyy"
```
mark-prins, line 15: I wonder is this affected if the browser settings are for a different locale 🤔 
Don't think you should either change or check it. I was just wondering if we'll have problems later on

packages/common/src/ui/components/inputs/TextInput/BasicTextInput.tsx
mark-prins: works really well with the date picker!

mark-prins (approved):
they look great! nice and easy to use as well.. and have keyboard navigation and localisation! 😁

packages/common/src/ui/components/inputs/DatePickers/DatePickerInput/DatePickerInput.tsx
joshxg: yes I think we will 😁

PR #602, opened 2021-12-03
#586 stocktake date

### LINKED ISSUE #586: display/edit Stocktake date
In the stocktake detail view, we need to have an editable stock take date as per https://docs.google.com/spreadsheets/d/1H0Rv1DEOgW0cE07wpfc_KivLf8TcNhYDqlwRp92x-To/edit#gid=0&range=A362

Should need: https://github.com/openmsupply/openmsupply-client/issues/504

joshxg:
Fixes #586 

- Adds stocktake date to stocktake detail view
image

Couple of things came up during this one.
1. Want to be able to have the stocktake date as `null` because that's what mSupply allows. I think it's possible that we might need to make it required. I would want to do this validation on save, I think, as trying to 'force' a value is often very annoying. Alternatively, maybe an `onBlur` of setting todays date might be easier
2. Realised dates coming in from the api are `unknown` which is not ideal. Ideally, we can type them as a `string` and then parse them all as proper `Date` objects as they come in, then parse them back into a string on the way back?

Will make an issue for 1 depending on what you think the behaviour should be. For 2, updating all the date types can be a new issue also


codegen.yml
```diff
@@ -7,4 +7,8 @@ generates:
       - typescript
       - typescript-operations
       - typescript-graphql-request
-      - typescript-msw
\ No newline at end of file
+      - typescript-msw
+    config:
+      scalars:
+        DateTime: string
+        NaiveDate: string
```
joshxg, line 14: little bit of an experiment. Works but it's still not an actual `Date`!

packages/common/package.json
```diff
@@ -13,7 +13,7 @@
     "@bugsnag/js": "^7.11.0",
     "@emotion/react": "^11.4.1",
     "@emotion/styled": "^11.3.0",
-    "@mui/lab": "^5.0.0-alpha.47",
+    "@mui/lab": "^5.0.0-alpha.58",
```
joshxg, line 16: I updated this too see if anything changed with the date picker - it didn't

packages/common/src/ui/components/inputs/TextInput/BasicTextInput.tsx
```diff
@@ -21,8 +21,9 @@ export const BasicTextInput: FC<StandardTextFieldProps> = React.forwardRef(
         disableUnderline: true,
         ...InputProps,
         sx: {
-          border: theme =>
-            error ? `2px solid ${theme.palette.error.main}` : 'none',
+          border: theme => {
```
mark-prins, line 24: perhaps revert this one?

mark-prins:
1. I didn't check to see why mSupply might have a requirement for null stocktake dates. I think that's strange for normal operation, so would prefer to drop that and just default the date to today. Currently when creating a new stocktake the date is set to `06/01/1970` which isn't terribly helpful! If we add validation.. yeah. we should chat about how to handle form validation - I expect that there are various opinions available
2. Oh, are they?! huh. Yes, type as `string` and parse as `Date` coming in, format back as string on the way out. 😄

mark-prins (approved):
👍  Nice!
what you've done with the date picker is looking very neat and fits in perfectly with the overall design

PR #605, opened 2021-12-04
Add status translations and handling of save & next status button

### LINKED ISSUE #584: stocktake statuses
I haven't seen anything about stocktake statuses changing at all so should just be suggested and finalised?

- Correctly translate the statuses
- Correctly hook up the OK & {next status} button

joshxg:
Fixes #584 

- Just using suggested/finalised which is what I think msupply is using. I prefer new/finalised, so I thought if I used suggested/finalised then there would be more chance to lock in draft
- Added translations for the detail view and list view
- Made the next status button work

### DIFF
--- a/packages/inventory/src/Stocktake/DetailView/reducer.ts
+++ b/packages/inventory/src/Stocktake/DetailView/reducer.ts
@@ -1,17 +1,20 @@
-import {
-  StocktakeItem,
-  StocktakeActionType,
-  StocktakeController,
-} from '../../types';
 import { Dispatch } from 'react';
 import {
   produce,
   DocumentActionSet,
   DocumentActionType,
   SortBy,
+  StocktakeNodeStatus,
 } from '@openmsupply-client/common';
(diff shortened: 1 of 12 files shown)


packages/inventory/src/Stocktake/DetailView/reducer.ts
```diff
@@ -103,6 +114,13 @@ export const reducer = (
           state.draft.onHold = !state.draft.onHold;
           break;
         }
+
+        case StocktakeActionType.UpdateStatus: {
+          // TODO: Probably there should be some more checks here for
+          // finalising
```
mark-prins, line 120: as far as I can tell, mSupply only asks. 
If there aren't adjustments to make, it just finalises all the same.

mark-prins (approved):
👍  Looks good!

FWIW, I prefer the status values of `new` & `finalised` - easier for now just to take what mSupply has and tidy up later tho

PR #606, opened 2021-12-04
#585 Add item columns and summary items to stocktake detail view

### LINKED ISSUE #585: Stocktake item columns
Currently the stocktake item columns are just placeholders. Need to add the 'correct' columns!

- Requirements here: https://docs.google.com/spreadsheets/d/1H0Rv1DEOgW0cE07wpfc_KivLf8TcNhYDqlwRp92x-To/edit#gid=320810900&range=A286:S286

joshxg:
Fixes #585 

- At least partly :)
- I am not sure that these columns are right. You could add a bunch of things pack sizes, prices etc. but I though I would just do this for now and get feedback - the pieces are in place to lock in/add columns
- The sorting isn't working well. I started adding it and realised that I would need a new sorting function that needs to account for the `[multiple]` values but wanted to check in to see what would be best for that. Determining the aggregated value of the column (for say, counted number of packs) and sorting by that value (which doesn't work for say, batch) or just sorting all the `[multiple]` as either the 'first' or 'last' values in a sort 🤔

### DIFF
--- a/packages/common/src/intl/locales/en/common.json
+++ b/packages/common/src/intl/locales/en/common.json
@@ -127,5 +127,7 @@
   "link.history": "History",
   "placeholder.enter-a-customers-name": "Enter a customers name",
   "placeholder.enter-a-suppliers-name": "Enter a suppliers name",
-  "placeholder.enter-an-item-code-or-name": "Enter item code or name"
+  "placeholder.enter-an-item-code-or-name": "Enter item code or name",
+  "label.counted-num-of-packs": "Counted # of Packs",
+  "label.snapshot-num-of-packs": "Snapshot # of Packs"
 }
--- a/packages/common/src/ui/layout/tables/hooks/useColumns/useColumns.tsx
+++ b/packages/common/src/ui/layout/tables/hooks/useColumns/useColumns.tsx
@@ -57,7 +57,8 @@ const getDefaultAccessor =
   ): ColumnDataAccessor<T> =>
   (row: T) => {
     const key = column.key as keyof T;
-    return row[key];
+    const value = row[key];
+    return typeof value === 'function' ? value() : value;
   };
 
 const getDefaultFormatter = <T extends DomainObject>(
--- a/packages/common/src/utils/arrays/utils.ts
+++ b/packages/common/src/utils/arrays/utils.ts
@@ -1,15 +1,13 @@
-export const ifTheSameElseDefault = <
-  T,
-  K extends keyof T,
-  J extends T[K] | undefined
->(
+export const ifTheSameElseDefault = <T, K extends keyof T, J>(
   someEntities: T[],
   key: K,
   defaultValue: J
-): J => {
-  const value = someEntities[0]?.[key] as J;
-  const allTheSame = someEntities.every(entity => entity[key] === value);
-  return allTheSame && value ? value : defaultValue;
+): J | T[K] => {
+  const value = someEntities[0]?.[key];
+  const allTheSame = someEntities.every(entity => {
+    return entity[key] === value;
+  });
+  return allTheSame && value != undefined ? value : defaultValue;
 };
 
 export const arrayToRecord = <T extends { id: string }>(
--- a/packages/inventory/src/Stocktake/DetailView/DetailView.tsx
+++ b/packages/inventory/src/Stocktake/DetailView/DetailView.tsx
@@ -10,8 +10,12 @@ import {
   DataTable,
   usePagination,
   useTranslation,
+  getRowExpandColumn,
+  ColumnAlign,
+  Box,
+  Column,
 } from '@openmsupply-client/common';
-import { reducer } from './reducer';
+import { reducer, StocktakeActionCreator } from './reducer';
 import { getStocktakeDetailViewApi } from './api';
 import { Toolbar } from './Toolbar';
 import { Footer } from './Footer';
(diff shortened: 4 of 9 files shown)


packages/common/src/intl/locales/en/common.json
```diff
@@ -127,5 +127,7 @@
   "link.history": "History",
   "placeholder.enter-a-customers-name": "Enter a customers name",
   "placeholder.enter-a-suppliers-name": "Enter a suppliers name",
-  "placeholder.enter-an-item-code-or-name": "Enter item code or name"
-}
+  "placeholder.enter-an-item-code-or-name": "Enter item code or name",
+  "label.counted-num-of-packs": "Counted # of Packs",
+  "label.snapshot-num-of-packs": "Snapshot # of Packs"
+}
```
joshxg, line 133: I've forgotten to do the translation issue to make the data table name space before this so have just doubled up a bit for now

packages/common/src/ui/layout/tables/hooks/useColumns/useColumns.tsx
```diff
@@ -57,7 +57,8 @@ const getDefaultAccessor =
   ): ColumnDataAccessor<T> =>
   (row: T) => {
     const key = column.key as keyof T;
-    return row[key];
+    const value = row[key];
+    return typeof value === 'function' ? value() : value;
```
joshxg, line 61: Will see later on that the values of a summary item are functions rather than concrete values so that they can't get out of sync

packages/common/src/utils/arrays/utils.ts
```diff
@@ -1,15 +1,13 @@
...
-): J => {
-  const value = someEntities[0]?.[key] as J;
-  const allTheSame = someEntities.every(entity => entity[key] === value);
-  return allTheSame && value ? value : defaultValue;
+): J | T[K] => {
+  const value = someEntities[0]?.[key];
+  const allTheSame = someEntities.every(entity => {
+    return entity[key] === value;
+  });
+  return allTheSame && value != undefined ? value : defaultValue;
```
joshxg, line 10: I've changed this to allow the default value to differ from the value of the `entity[key]` (as to use `[multiple]` for number values)

packages/inventory/src/Stocktake/DetailView/DetailView.tsx
```diff
@@ -30,16 +34,55 @@ const useDraftStocktake = () => {
...
+        bgcolor="#c7c9d933"
+      >
+        <span style={{ whiteSpace: 'pre-wrap' }}>
+          {JSON.stringify(rowData, null, 2)}
+        </span>
+      </Box>
+    </Box>
+  );
+};
+
```
joshxg, line 61: just the placeholder

packages/inventory/src/Stocktake/DetailView/DetailView.tsx
```diff
@@ -30,16 +34,55 @@ const useDraftStocktake = () => {
...
+        key: 'countedNumPacks',
+        label: 'label.counted-num-of-packs',
+        width: 150,
+        align: ColumnAlign.Right,
+      },
+      {
+        key: 'snapshotNumPacks',
+        label: 'label.snapshot-num-of-packs',
+        align: ColumnAlign.Right,
+      },
```
joshxg, line 81: The list of columns is already massive. Thinking to just put reused columns there and keep single use ones defined inline 🤷

packages/inventory/src/Stocktake/DetailView/reducer.ts
```diff
@@ -45,13 +49,56 @@ const StocktakeActionCreator = {
...
+    itemName: () => ifTheSameElseDefault(lines, 'itemName', ''),
+    itemCode: () => ifTheSameElseDefault(lines, 'itemCode', ''),
+    batch: () => ifTheSameElseDefault(lines, 'batch', '[multiple]'),
+    expiryDate: () => ifTheSameElseDefault(lines, 'expiryDate', '[multiple]'),
+    countedNumPacks: () =>
+      String(ifTheSameElseDefault(lines, 'countedNumPacks', '[multiple]')),
+    snapshotNumPacks: () =>
+      String(ifTheSameElseDefault(lines, 'snapshotNumPacks', '[multiple]')),
+  };
+};
```
joshxg, line 94: little bit of an experiment/iteration. I don't like trying to keep the summary item values in sync, so calculating each time seems better.. though I suspect that the closure will make this trickier than I would hope

joshxg: Ah, and the untranslated `[multiple]`  - I'm thinking of putting into the cell a check to see if the value is `[multiple]` and using a translated string  🤔

packages/inventory/src/Stocktake/DetailView/DetailView.tsx
mark-prins: that's a good idea

packages/inventory/src/Stocktake/DetailView/reducer.ts
```diff
@@ -121,6 +170,37 @@ export const reducer = (
...
+
+          const newIsDesc = currentSortKey === key ? !currentIsDesc : false;
+          const newDirection: 'asc' | 'desc' = newIsDesc ? 'desc' : 'asc';
+          const newSortBy: SortBy<StocktakeItem> = {
+            key,
+            isDesc: newIsDesc,
+            direction: newDirection,
+          };
+
+          const sortedLines = lines.sort(
```
mark-prins, line 192: the sorting isn't working for me - it looks like it should. not sure if I should be waiting for another part of the PR or not?

mark-prins (approved):
🎉  that's awesome! 
the functional column values are a neat solution. The grouping of batches by item is a bit of a bonus - something not in desktop and pretty neat.

issues: as already noted the `[multiple]` translation, and the column headers ( on a small screen ); and I have a problem with column sorting. nothing worth holding up a merge though, all stuff you've identified for later.

packages/inventory/src/Stocktake/DetailView/reducer.ts
joshxg: mm no, I wasn't sure what to do about the `[multiple]` values and sorting so I've left for now 🤔

PR #607, opened 2021-12-04
#587 stocktake edit item 1

### LINKED ISSUE #587: Stocktake add/edit item functionality
Epic for stocktake add/edit item functionality. Unsure if cells are editable in the main list view, but certainly need a modal for add/editing.

The current add stocktake lines modal is:
image

This shows all the batches:

image

You can "check' each batch if you want to add it as a line in your stock take.

Alternatively, the bottom half is for adding a brand new stock line that you magically found!

Then in the data table, batches are listed and double clicking on a batch opens the edit modal:

image

- you can only edit the pack size of the magically found stock lines.

joshxg:
Fixes #587 

- Bit more substance to this one!
image
- This is adding a pretty basic stocktake edit modal similar to inbounds

TODO: Creating stocktake lines from existing stock lines. 
- I am thinking either having some sort of list which shows existing stock lines (maybe on the left or right of the modal..?) where you can check the stock lines and this adds a stocktake line to the item OR just creating the stock lines and if they're not counted then they get pruned (mobile style) OR having a column "Count this line" (more explicit but doesn't make sense for newly added lines) OR a combo box with tags

🤷  Not sure. Probably need to play around with some options

### DIFF
--- a/packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/StocktakeLineEdit.tsx
+++ b/packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/StocktakeLineEdit.tsx
@@ -1,0 +1,12 @@
+import { FC } from 'react';
+import { StocktakeController, StocktakeItem } from '../../../../types';
+import { ModalMode } from '../../DetailView';
+
+interface StocktakeLineEditProps {
+  item: StocktakeItem | null;
+  onChangeItem: (item: StocktakeItem | null) => void;
+  mode: ModalMode;
+  draft: StocktakeController;
+}
+
+export const StocktakeLineEdit: FC<StocktakeLineEditProps> = () => null;
--- a/packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/StocktakeLineEdit.tsx
+++ b/packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/StocktakeLineEdit.tsx
@@ -2,6 +2,23 @@ import React, { FC } from 'react';
 import { StocktakeController, StocktakeItem } from '../../../../types';
 import { ModalMode } from '../../DetailView';
 import { StocktakeLineEditForm } from './StocktakeLineEditForm';
+import {
+  Divider,
+  TableContainer,
+  TabContext,
+  TabList,
+  Tab,
+  alpha,
+  TabPanel,
+  styled,
+  useTranslation,
+  useIsMediumScreen,
+  ButtonWithIcon,
+  PlusCircleIcon,
+  Box,
+} from '@openmsupply-client/common';
+import { BatchTable, PricingTable } from './StocktakeLineEditTables';
+import { createStocktakeRow, wrapStocktakeItem } from './utils';
 
 interface StocktakeLineEditProps {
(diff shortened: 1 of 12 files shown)


mark-prins:
A small issue: if I create a new stocktake, add a line and save, then I have an issue `Cannot read properties of null (reading 'toISOString')` due to the null `stocktakeDatetime`. a check in `api.ts:44` would do it, depending on what we decide for nullability of stocktakeDate :)

packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/StocktakeLineEdit.tsx
```diff
@@ -0,0 +1,131 @@
...
+const StyledTabPanel = styled(TabPanel)({
+  height: '100%',
+});
+
+enum Tabs {
+  Batch = 'Batch',
+  Pricing = 'Pricing',
+}
+
+export const StocktakeLineEdit: FC<StocktakeLineEditProps> = ({
```
mark-prins, line 39: perhaps when adding an item, automatically add a row on item selection?

packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/StocktakeLineEditTables.tsx
```diff
@@ -0,0 +1,61 @@
...
+      Cell: NumberInputCell,
+    },
+    'expiryDate',
+  ]);
+
+  return (
+    <DataTable
+      columns={columns}
+      data={batches}
+      noDataMessage="Add a new line"
```
mark-prins, line 39: could you add a translation for this please?

packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/StocktakeLineEdit.tsx
```diff
@@ -0,0 +1,131 @@
...
+                <Tab value={Tabs.Batch} label={Tabs.Batch} />
+                <Tab value={Tabs.Pricing} label={Tabs.Pricing} />
+              </TabList>
+            </Box>
+            <Box flex={1} justifyContent="flex-end" display="flex">
+              <ButtonWithIcon
+                color="primary"
+                variant="outlined"
+                onClick={onAddBatch}
+                label={t('label.add-line', { ns: 'inventory' })}
```
mark-prins, line 91: wondered if `Add Batch` would be more specific / helpful

mark-prins (approved):
🎉  badass! looks really good

packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/StocktakeLineEdit.tsx
joshxg: Generally will be counting the already existing stock lines and adding a new arbitrary batch would be a secondary action, I think. A mechanism to add existing stock lines to count would be preferable? Just not sure how exactly, as you also may not want to count some stock lines 🤔

joshxg:
Was simpler to just fix the null issue for now but I don't really mind whether the date is nullable or not

PR #610, opened 2021-12-04
#559 requisition list view columns

### LINKED ISSUE #559: Add order date+ color and fix requisition number column in requisition list views
Currently the requisition number column is `Reqnum` which isn't ideal. Also there is no `order date` column

- Add an `order date` column
- Change `reqnum` to .. `number` or `requisition number`

joshxg:
Fixes #559 

- Added an order date column
- made the header reqnum -> Number
- added the color column and made the update work correctly

### DIFF
--- a/packages/common/src/schema.graphql
+++ b/packages/common/src/schema.graphql
@@ -1273,8 +1273,6 @@ type UpdateOutboundShipmentResponseWithId {
 
 type RequisitionNode {
   id: String!
-
-  color: String
   orderDate: String
   comment: String
   theirReference: String
--- a/packages/common/src/schema.graphql
+++ b/packages/common/src/schema.graphql
@@ -1284,7 +1284,7 @@ type RequisitionNode {
   maxMOS: Int
   status: SupplierRequisitionNodeStatus!
   lines: RequisitionLinesResponse!
-
+  color: String
   otherParty: NameResponse!
   otherPartyId: String!
   otherPartyName: String!
@@ -1446,7 +1446,7 @@ type RequisitionConnector {
 input UpdateSupplierRequisitionInput {
   id: String!
   orderDate: String
-  nameId: String
+  otherPartyId: String
   comment: String
   theirReference: String
   color: String
@@ -1688,7 +1688,7 @@ type DeleteCustomerRequisitionLineResponseWithId {
 input UpdateCustomerRequisitionInput {
   id: String!
   orderDate: String
-  nameId: String
+  otherPartyId: String
   comment: String
   theirReference: String
   color: String
--- a/packages/mock-server/src/worker/handlers/requisition/index.ts
+++ b/packages/mock-server/src/worker/handlers/requisition/index.ts
@@ -11,6 +11,7 @@ import {
   BatchSupplierRequisitionInput,
   BatchCustomerRequisitionInput,
   DeleteCustomerRequisitionInput,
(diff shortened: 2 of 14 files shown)


packages/common/src/schema.graphql
```diff
@@ -1446,7 +1444,7 @@ type RequisitionConnector {
 input UpdateSupplierRequisitionInput {
   id: String!
   orderDate: String
-  nameId: String
+  otherPartyId: String
```
joshxg, line 1447: more consistent I think

packages/mock-server/src/worker/handlers/requisition/index.ts
```diff
@@ -32,6 +33,30 @@ const invoicesQuery = mockRequisitionsQuery((req, res, ctx) => {
...
+  (req, res, ctx) => {
+    return res(
+      ctx.data({
+        updateCustomerRequisition: MutationService.requisition.customer.update(
+          req.variables.input
+        ),
+      })
+    );
+  }
+);
```
joshxg, line 58: just adding the mocks for a single update mutation

packages/requisitions/src/CustomerRequisition/ListView/ListView.tsx
```diff
@@ -44,7 +47,24 @@ export const CustomerRequisitionListView: FC = () => {
...
+        key: 'requisitionNumber',
+        label: 'label.number',
+      },
+      'status',
+      {
+        key: 'orderDate',
+        label: 'label.requisition-date',
+        width: 100,
+        accessor: rowData => (rowData.orderDate ? d(rowData.orderDate) : ''),
+      },
```
joshxg, line 64: I suppose since these are used in two places you could argue that defining them in the set builder file is better?  At least defined here then you know that `rowData` is a `RequisitionRow` and it's a bit type saferer

packages/requisitions/src/CustomerRequisition/ListView/api.ts
```diff
@@ -21,7 +21,16 @@ const onRead =
...
     const { nodes } = requisitions;
 
-    return { nodes, totalCount: requisitions.totalCount };
+    return {
+      nodes: nodes.map(requisition => ({
+        ...requisition,
+        color: requisition?.color ?? '#8f90a6',
+        orderDate: requisition?.orderDate
+          ? new Date(requisition.orderDate)
+          : null,
```
joshxg, line 30: parsing orderDate to a date and assigning a default `color` .. maybe it's better to just have a default color in the component 🤔

packages/requisitions/src/CustomerRequisition/ListView/ListView.tsx
mark-prins: more type safe and keeps the list of columns down a little

packages/requisitions/src/CustomerRequisition/ListView/api.ts
mark-prins: yes, I like that idea. default colour in the component, makes things like this much easier.

and a date mapper? some other nifty way to map the dates?

mark-prins (approved):
👍  very nice!!

joshxg:
Hmm yes, I was thinking that once the backend has adjusted how errors are returned, mapping returns will be a lot easier so thought to wait until then to make a nice mapper for dates etc

PR #611, opened 2021-12-05
#589 requisition order and requisition dates

### LINKED ISSUE #589: editable order and requisition date on a customer requisition
msupply has editable requisition and order dates on an adhoc customer requisition:

image

joshxg:
Fixes #589

- Adds requisition + order date inputs to customer requisition
- Adds requisition date input to supplier requisition

I'm not sure where to put these was just trying to tick things off the list. Maybe should have waited for the meeting with dicky. I tried the side panel but thought that at least for customer requisitons that these dates were in a 'primary' position in mSupply so maybe they should go here - then wanted the consistency for supplier requisitions... even though i'm sure programs will throw everything into an inconsistent state

### DIFF
--- a/packages/requisitions/src/CustomerRequisition/DetailView/SidePanel.tsx
+++ b/packages/requisitions/src/CustomerRequisition/DetailView/SidePanel.tsx
@@ -13,8 +13,6 @@ import {
   PanelRow,
   PanelField,
   ColorSelectButton,
-  useFormatDate,
-  DatePickerInput,
 } from '@openmsupply-client/common';
 import { isRequisitionEditable } from '../../utils';
 import { CustomerRequisition } from '../../types';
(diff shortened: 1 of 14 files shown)


packages/requisitions/src/CustomerRequisition/DetailView/SidePanel.tsx
```diff
@@ -59,6 +61,24 @@ const RelatedDocumentsSection: FC<SidePanelProps> = () => {
   );
 };
 
+const DatesSection: FC<SidePanelProps> = ({ draft }) => {
```
mark-prins, line 64: should this be here, now that the date is in the top bit?
If so - the heading should change 😄

mark-prins (approved):
👍  changes look great!

I have an issue if I try to update the date for the supplier requisition - would you mind checking that one for me? I get the `Placeholder callback updateRequisitionDate has been triggered. This should never happen!` when I select a date, for an existing req

packages/requisitions/src/CustomerRequisition/DetailView/SidePanel.tsx
joshxg: whoops 😁

joshxg:
fixed the date - oops!

Issue #556, opened 2021-12-02
Update GraphQL schema and types


joshxg:
Backend has merged some PRs - time to update the graphql schema and types

PR #614, opened 2021-12-06
Add shift all additions out to separate file

joshxg:
Starting #556 !

- I wanted to update the gql schema. An easier way for this is to just use the URL of the demo server
- So, taking all the conflicts out of the file into the additions file - should make it easier to share, also.
- Thought to do this in stages as the diffs are so big and annoying to read. Next up I will change over to using the URL to generate the schema


mark-prins (approved):
👍  awesome! thanks for doing that - a bit tedious.. but great to have done

mark-prins (approved):
huh. I thought that I had approved already 🙄  soz.

PR #616, opened 2021-12-07
#556 update gql schema 2

### LINKED ISSUE #556: Update GraphQL schema and types

joshxg:
Fixes #556 

- Ok this is generating types from the demo server. Good idea? Maybe.
- Means we might have some annoyances when types are changed and added, but I suppose best we deal with that asap..
- It's pretty outrageous the amount of changes I needed to hack in just because of `item` being added to `InvoiceLineNode`. Was quite infuriating, actually! The way we're parsing the response using the generated types means that you need to query for all the required fields otherwise typescript barfs. Ugh! For now I've just added the overfetching. I *think* a solution might be to use fragments to generate smaller types, but I've yet to try it yet. Thinking of waiting till we have better queries, but also don't want to wait too long. Sigh!
- This has really made me want to decouple the frontend types from the backend types as much as possible - the backend adding a required field to a type (where the frontend didn't even start using it yet!) shouldn't stop the frontend compiling, let a long take so much time to fix.. sorry rant over!

### DIFF
--- a/codegen.yml
+++ b/codegen.yml
@@ -1,5 +1,5 @@
 overwrite: true
-schema: ["./packages/common/src/schema.graphql", "./packages/common/src/additions.schema.graphql"]
+schema: ["https://demo-open.msupply.org:8000/graphql", "./packages/common/src/additions.schema.graphql"]
 documents: './packages/common/src/operations.graphql'
 generates:
   ./packages/common/src/types/schema.ts:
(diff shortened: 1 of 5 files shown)


joshxg:
It is possible to get generated types from fragments also, so we could use this to simplify our usage of generated types, I think. I'm very tempted to do this but I think it would be best to just wait until the api is in a state where we can less tediously map the results of a query

joshxg:
😢  have just noticed the mock server isn't started cause of a ts barf because of more issues with adding the item field. On that now...

joshxg:
Sorry I'm a muppet and completely forgot about the mock server! 🙃

PR #617, opened 2021-12-07
#556 update gql schema 2

### LINKED ISSUE #556: Update GraphQL schema and types

joshxg:
Fixes #556 

- Took me a bit longer to get working than I had hoped..
- Ok this is generating types from the demo server. Good idea? Maybe.
- It's also taking the type defs from the server and merging with the additions when starting our apollo server
- Means we might have some annoyances when types are changed and added, but I suppose best we deal with that asap..
- It's pretty outrageous the amount of changes I needed to hack in just because of `item` being added to `InvoiceLineNode`. Was quite infuriating, actually! The way we're parsing the response using the generated types means that you need to query for all the required fields otherwise typescript barfs. Ugh! For now I've just added the overfetching. I *think* a solution might be to use fragments to generate smaller types, but I've yet to try it yet. Thinking of waiting till we have better queries, but also don't want to wait too long. Sigh!
- This has really made me want to decouple the frontend types from the backend types as much as possible - the backend adding a required field to a type (where the frontend didn't even start using it yet!) shouldn't stop the frontend compiling, let a long take so much time to fix.. sorry rant over!

### DIFF
--- a/codegen.yml
+++ b/codegen.yml
@@ -1,5 +1,5 @@
 overwrite: true
-schema: ["./packages/common/src/schema.graphql", "./packages/common/src/additions.schema.graphql"]
+schema: ["https://demo-open.msupply.org:8000/graphql", "./packages/common/src/additions.schema.graphql"]
 documents: './packages/common/src/operations.graphql'
 generates:
   ./packages/common/src/types/schema.ts:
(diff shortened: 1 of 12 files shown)


mark-prins (approved):
🥇  Impressive! it all looks so clear and simple when you've done it - what a mission to get to this point though.

Really like the idea of taking the current schema and merging the additions. important to keep up to date with the api schema.. though totally agree about the separation! can't have a minor change breaking the compilation

PR #619, opened 2021-12-07
#479 ts path aliases

### LINKED ISSUE #479: Setup ts path aliases
OK I've gotten annoyed enough to make an issue for this now.

- ts path aliases for at least just common, for now. 
- `@common/components`, `@common/types`, `@common/intl` or something?

mark-prins:
Fixes #479 

Have added aliases for 
- `@common/components`
- `@common/hooks`
- `@common/icons`
- `@common/intl`
- `@common/styles`
- `@common/types`

and updated the imports for a bunch of usages as examples.
Have left `mock-server` as that probably isn't a permanent feature and since that isn't using webpack it isn't as easy to do!

Found that the tsconfig wasn't valid for some of the packages and therefore when I fixed that, we had a bunch of compiler complaints 🙄  bit of an offroad to fix those - possibly not correctly, but it at least passes now.

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,4 +1,5 @@
 const path = require('path');
+const TsconfigPathsPlugin = require('tsconfig-paths-webpack-plugin');
 const toPath = filePath => path.join(process.cwd(), filePath);
 
 module.exports = {
@@ -21,6 +22,7 @@ module.exports = {
           '@emotion/core': toPath('node_modules/@emotion/react'),
           'emotion-theming': toPath('node_modules/@emotion/react'),
         },
+        plugins: [new TsconfigPathsPlugin()],
       },
     };
   },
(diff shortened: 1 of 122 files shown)


.storybook/main.js
```diff
@@ -1,4 +1,5 @@
 const path = require('path');
+const TsconfigPathsPlugin = require('tsconfig-paths-webpack-plugin');
```
mark-prins, line 2: using the plugin to avoid copying alias values in three different places :) need to have them in `tsconfig.json` and webpack for host and storybook.

jest.config.js
```diff
@@ -6,5 +9,13 @@ module.exports = {
   setupFilesAfterEnv: ['<rootDir>/jest-setup.ts'],
   moduleNameMapper: {
     'react-i18next': '<rootDir>/__mocks__/react-i18next.ts',
+    ...pathsToModuleNameMapper(compilerOptions.paths, {
```
mark-prins, line 12: could just type the alias values out, but this way it will read them from `tsconfig`

joshxg (approved):
This is so great. Thanks so much for getting it done it's going to be so helpful! And sorry you had to fix all the compilation mistakes!

Issue #618, opened 2021-12-07
Stocktake: Add existing stock lines


joshxg:
Currently we can only add new arbitrary batches. We need a way to add existing stock lines to count!

![image](https://user-images.githubusercontent.com/35858975/145098509-af8b7556-d91f-4597-bcff-dd8e56c8e733.png)

I've not really any idea how to do the ui for this 🤔

PR #623, opened 2021-12-08
#618 existing lines take 2

joshxg:
Another one!

### DIFF
--- a/packages/common/src/types/schema.ts
+++ b/packages/common/src/types/schema.ts
@@ -1796,6 +1796,7 @@ export type StocktakeLineNode = {
   sellPricePerPack?: Maybe<Scalars['Float']>;
   snapshotNumPacks?: Maybe<Scalars['Int']>;
   snapshotPackSize?: Maybe<Scalars['Int']>;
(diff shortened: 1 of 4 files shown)


mark-prins (approved):
👍  that's nice to use and, I think, is clear and easy to understand.

A couple of general comments:
- I found the re-ordering when a line is checked confusing. would be happier with the order staying consistent and the item I checked staying where it was.. though I get what you mean with the 'uncounted' lines sinking to the bottom and being out of the way.  Maybe animation and/or a faded font/ background colour to the row.. something showing that this row is dead to me now?
- would be nice if the row selected itself if you enter a value in the counted field
- similarly, when adding a row, would be good to pre-select it

joshxg:
Hmm ok I was planning on this just being a draft but I'll just merge it and work on it from here heh.

For the reordering... yeah. Maybe an initial sort when you open the modal but no reordering while you work on it 🤔

mark-prins:
sounds good - I think that it's functional enough to merge and iterate! 😀

PR #624, opened 2021-12-08
limit server setup in tests

### LINKED ISSUE #554: Mock data is generated for each test
Whenever each test is run, the mock data is generated.

Just makes some tests take a bit longer than what would be ideal. Think it basically needs to all be chucked into a function rather than at the modules top level and called when either the mock server runs or the service worker starts

mark-prins:
Fixes #554 

Spent a while exploring various options and landed here. It was getting very involved trying to encapsulate and isolate the mock data generation.. and even then the `beforeAll` in `jest-setup.ts`wasn't keeping global state across the test suite, only for each test file. so I went back to the original problem, which is that the server was being instantiated for every test file and then the data generated for every test within every file. Simple solution was to lift that out and only setup the server for tests that use it (3 of them). Tests now run faster ( originally running ~80s now around 60s for the suite )

### DIFF
--- a/jest-setup.ts
+++ b/jest-setup.ts
@@ -1,6 +1,5 @@
 import '@testing-library/jest-dom';
 import { setScreenSize_ONLY_FOR_TESTING } from '@openmsupply-client/common';
-import { setupMockServer } from '@openmsupply-client/mock-server/src/worker/server';
 
 let mockStorage: Record<string, string> = {};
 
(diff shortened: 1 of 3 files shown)


joshxg (approved):
Oh! Of course. Ugh that must have been annoying finding that, I was a bit off in my guess of what was wrong! Thanks

PR #625, opened 2021-12-09
#160 table parameters in url

### LINKED ISSUE #160: Outbound shipment: ListView table parameters should be in the URL
More of a question :)

Should the `sortBy`, `limit`, `offset` all be in the URL so they can be linked to?

mark-prins:
Fixes #160 

Adds url parameters to keep `page` `sort` & `desc`
Thought about filtering and decided not ( the filter types were more complex than just having a search term in the URL parameters ). Now reconsidering.. maybe later?

### DIFF
--- a/packages/common/src/hooks/usePagination/usePagination.ts
+++ b/packages/common/src/hooks/usePagination/usePagination.ts
@@ -1,6 +1,5 @@
 import { useRef, useState, useEffect } from 'react';
-import { useSearchParams } from 'react-router-dom';
-
+import { useSearchParameters } from '../useSearchParameters';
 export interface Pagination {
   page: number;
   offset: number;
@@ -18,11 +17,9 @@ export interface PaginationState extends PaginationController {
 }
 
 export const usePagination = (initialFirst = 20): PaginationState => {
-  const [searchParams, setSearchParams] = useSearchParams();
-  const getPageFromSearchParams = () => {
-    const paramPage = Number(searchParams.get('page'));
-    return Number.isNaN(paramPage) ? 0 : Math.max(0, paramPage - 1);
-  };
+  const searchParams = useSearchParameters();
+  const getPageFromSearchParams = () =>
+    Math.max(0, searchParams.getNumber('page') - 1);
   const [first, onChangeFirst] = useState(initialFirst);
   const [offset, onChangeOffset] = useState(0);
   const [page, onChangePage] = useState(getPageFromSearchParams());
@@ -32,7 +29,7 @@ export const usePagination = (initialFirst = 20): PaginationState => {
   useEffect(() => {
     onChangeOffset(page * first);
     pageRef.current = page;
-    setSearchParams({ page: String(page + 1) });
+    searchParams.set({ page: String(page + 1) });
   }, [first, page]);
 
   useEffect(() => {
--- a/packages/common/src/hooks/usePagination/usePagination.ts
+++ b/packages/common/src/hooks/usePagination/usePagination.ts
@@ -1,4 +1,5 @@
 import { useRef, useState, useEffect } from 'react';
+import { useSearchParams } from 'react-router-dom';
 
 export interface Pagination {
   page: number;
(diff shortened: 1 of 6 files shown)


packages/common/src/hooks/useSearchParameters/useSearchParameters.test.tsx
```diff
@@ -0,0 +1,80 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import { act } from 'react-dom/test-utils';
+import { MemoryRouter, useSearchParams } from 'react-router-dom';
+import { Route, Routes } from 'react-router';
+
+import { useSearchParameters } from './useSearchParameters';
+
+const ShowParams = () => {
```
mark-prins, line 9: yeah.. had a bit of a problem getting the history out in order to test and so bodged this in. Not great, but good enough? maybe.. 🤷

packages/common/src/hooks/useSearchParameters/useSearchParameters.ts
```diff
@@ -0,0 +1,26 @@
+import { useSearchParams } from 'react-router-dom';
+
+interface SearchParams {
+  get: (key: string) => string | null;
+  set: (params: Record<string, string>) => void;
+  getNumber: (key: string) => number;
+}
+
+export const useSearchParameters = (): SearchParams => {
```
mark-prins, line 9: popped this in because:
- the `set` method on `searchParams` didn't do anything
- the `setSearchParams` replaces search params with the object passed to it, which isn't helpful if you want to update one parameter only and retain others

packages/common/src/hooks/usePagination/usePagination.ts
```diff
@@ -17,17 +17,27 @@ export interface PaginationState extends PaginationController {
...
     pageRef.current = page;
+    searchParams.set({ page: String(page + 1) });
   }, [first, page]);
 
+  useEffect(() => {
+    const newPage = getPageFromSearchParams();
+    onChangeOffset(newPage * first);
+    pageRef.current = newPage;
+  }, [location]);
+
```
joshxg, line 40: If for example I click on the navigation menu for the same page i'm on, the page param disappears. Not a big deal but seems inconsistent..

Thought something like

```
useEffect(() => {
    if (!searchParams.getNumber('page')) {
      searchParams.set({ page: String(page + 1) });
    }
  });
```

similar with the initial sort state, though I see you have a test that the initial state is empty, so maybe this is not right

packages/common/src/hooks/useSearchParameters/useSearchParameters.test.tsx
joshxg: Ah yes I can see how annoying that would be. Seems fine. I hope to write some tests of the actual pages and maybe we can put some tests there with more 'real' use cases.. though would still need to figure out how to get the history out, so maybe not!

packages/common/src/hooks/useSearchParameters/useSearchParameters.ts
joshxg: cool, I like!

packages/common/src/ui/layout/tables/DataTable.tsx
```diff
@@ -36,6 +36,13 @@ export const DataTable = <T extends DomainObject>({
     if (data.length) setActiveRows(data.map(({ id }) => id as string));
   }, [data]);
 
+  // guard against a page number being set which is greater than the data allows
+  useEffect(() => {
+    if (!pagination || !onChangePage || !pagination.total) return;
+    const { page, first, total } = pagination;
+    if (page * first > total) onChangePage(0);
+  }, [pagination]);
```
joshxg, line 44: annoying that `total` isn't in the `usePagination` or `useQueryParams` hooks :(

joshxg (approved):
Nice 😁

packages/common/src/ui/layout/tables/DataTable.tsx
mark-prins: yup - that would have made it neater!

packages/common/src/hooks/usePagination/usePagination.ts
mark-prins: good catch! thanks

mark-prins:
in setting the defaults, picked up a couple of other errors and had to update the tests in doing so
55ed4687

PR #629, opened 2021-12-09
#618 existing lines

### LINKED ISSUE #618: Stocktake: Add existing stock lines

joshxg:
Fixes #618  .. in a way!

- I've shifted the 'static' part of the panel out as I added the checkbox and line number here also
- Check is checked for counted lines and when adding a batch/item, unchecked otherwise. Though I've not completely done the mutations for this


mark-prins (approved):
👍  that looks great! 
it's all clear enough to me when using it 😃

PR #633, opened 2021-12-12
Add updated invoiceCounts

### LINKED ISSUE #632: Update invoiceCounts query to match the new api
The `invoiceCounts` query has changed on the backend API causing the code gen to break

joshxg:
Fixes #632 

- the `created` isn't supported yet so just randomising a number. Everything else seems straight forward


mark-prins (approved):
👍  very nice!

Issue #634, opened 2021-12-12
Example inbound shipments page without a save button


joshxg:
Want to do an example of the inbound shipment page without a save button, automatically saving things on the fly.

- Update comment/color/supplier ref on the fly - on losing focus? debounced? (race condition with updating a field and then clicking 'change status' - i.e. if. you change to finalised before losing focus/debouncing and the 'update' comes after for fields you're not able to update?)
- Send mutation when clicking OK on the add/edit modal

PR #635, opened 2021-12-13
#634 inbound no save button

joshxg:
Starts #634 

- This ended up being relatively easy, though I'm not sure I can say that as I haven't fully implemented everything yet!
- The saving of items is quite easy right now. Though, it could likely be implemented better. Currently just sending the full batch mutation as before. Inserting could easily be done better (just check `ModalMode` and use an insert..) but updating would still require a check for inserting or updating lines. It's easier to just leave it as a single mutation 😁 
- The inputs on the main document are slow and laggy. There's a few ways to handle it - thought I would leave that for now


packages/common/src/hooks/useDialog/useDialog.tsx
```diff
@@ -90,9 +90,10 @@ export const useDialog = (dialogProps?: DialogProps): DialogState => {
       // TODO: If you want to change the slide direction or other animation details, add a prop
       // slideAnimationConfig and add a parameter to `useSlideAnimation` to pass in the config.
       WrappedNextButton = React.cloneElement(nextButton, {
-        onClick: () => {
-          onTriggerSlide();
-          onClick();
+        onClick: async () => {
+          const result = await onClick();
+          if (!!result) onTriggerSlide();
+          return result;
```
joshxg, line 96: I've just done some adjusting here so that the animation after clicking `onNext` can be stopped and also waits for the `onClick`.

Not the best, I think, as I'd prefer that the animation happen by default and you can return something so that it doesn't happen 🤔

packages/common/src/intl/locales/en/common.json
```diff
@@ -33,7 +33,7 @@
   "button.ok-and-next": "OK & Next",
   "button.ok": "OK",
   "button.print": "Print",
-  "button.save-and-confirm-status": "Save & Confirm {{status}}",
+  "button.save-and-confirm-status": "Confirm {{status}}",
```
joshxg, line 36: I wasn't sure what to put here

packages/common/src/utils/arrays/utils.ts
```diff
@@ -18,3 +18,20 @@ export const arrayToRecord = <T extends { id: string }>(
...
+  const lookup: Record<string, T[]> = {} as Record<string, T[]>;
+  things.forEach(thing => {
+    const value = String(thing[key]);
+    if (!lookup[value]) {
+      lookup[value] = [];
+    }
+    (lookup[value] ?? []).push(thing);
+  });
+  return lookup;
+};
```
joshxg, line 37: shifted here from elsewhere which sparked the `lodash` question - will make an issue for adding that package

packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
```diff
@@ -1,60 +1,108 @@
...
+} from '../../utils';
+import {
+  InboundShipmentItem,
+  InboundShipment,
+  Invoice,
+  OutboundShipmentSummaryItem,
+} from '../../types';
 
 const useDraftInbound = () => {
+  const queryClient = useQueryClient();
```
joshxg, line 40: is a bit rushed for now but gives a good example.

Not using the reducer at all here. Which means that sorting isn't working. But it's a lot of code and complexities removed which is good, I think.

packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
```diff
@@ -1,60 +1,108 @@
...
-  );
+  const onChangeSortBy = () => {};
+
+  const { data } = useQuery(['invoice', id], () => {
+    return queries.onRead(id);
+  });
+
+  const draft = data
+    ? { ...data, items: inboundLinesToSummaryItems(data?.lines ?? []) }
+    : placeholderInbound;
```
joshxg, line 53: it's possible that to enabling sorting, we could store this `draft` in state and then sort that copy. Also possible to use the react-query `select` parameter.

packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
```diff
@@ -1,60 +1,108 @@
...
+        ...previousInbound,
+        ...patch,
+      });
+
+      return { previousInbound, patch };
+    },
+    onSettled: () => queryClient.invalidateQueries(['invoice', id]),
+    onError: (_, __, context) => {
+      queryClient.setQueryData(['invoice', id], context.previousInbound);
+    },
```
joshxg, line 74: Just a simple optimistic mutation implementation for now. This is pretty generic so should be able to make a custom hook for this easily.

All that trouble trying to think of a way to get a function with `(key, value)` parameters to work fully type safe when I couldv'e just been using `Partial<T>` the whole time!

packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
```diff
@@ -1,60 +1,108 @@
...
+  });
+
+  const { isLoading: isAddingItem, mutateAsync: noOptimisticMutate } =
+    useMutation(queries.onUpdate, {
+      onSettled: () => queryClient.invalidateQueries(['invoice', id]),
+    });
```
joshxg, line 80: this one is just standard non-optimistic for upserting items. Possibly better to have this not send the invoice

packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
joshxg: Also prooobably better to just update the invoice here

packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
```diff
@@ -1,60 +1,108 @@
...
   };
 
-  return { draft, save, dispatch, onChangeSortBy, sortBy: state.sortBy };
-};
+  const upsertItem = async (item: OutboundShipmentSummaryItem) => {
+    const itemIdx = draft.items.findIndex(i => i.id === item.id);
+    if (itemIdx >= 0) draft.items[itemIdx] = item;
+    else draft.items.push(item);
+
+    // throw new Error('testing!');
```
joshxg, line 91: was just using this for to test the behaviour when throwing an error when adding/updating an item

packages/invoices/src/InboundShipment/DetailView/SidePanel.tsx
```diff
@@ -19,9 +19,10 @@ import { InboundShipment } from '../../types';
 
 interface SidePanelProps {
   draft: InboundShipment;
+  update: (patch: Partial<InboundShipment>) => Promise<InboundShipment>;
```
joshxg, line 22: Instead of passing the update function (and even the draft) - at this point we could just `useQuery` and `useMutation` in these components and use the cached invoice value in all of these components. Just use a `useInbound(id)` and `useUpdateInvoice(id)` hooks and it would be quite easy

joshxg (commented):
I was concerned about the lag of adding items and I still am a bit. On a heavily throttled connection it seems to be OK, though I think in the real world it might get a bit frustrating.

Overall the code would be significantly simpler and vastly simpler, if we can take the hit on adding/editing item lag on ok/ok&next

packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
```diff
@@ -1,60 +1,108 @@
...
+      ]);
+
+      queryClient.setQueryData(['invoice', id], {
+        ...previousInbound,
+        ...patch,
+      });
+
+      return { previousInbound, patch };
+    },
+    onSettled: () => queryClient.invalidateQueries(['invoice', id]),
```
mark-prins, line 71: wondering if this is required, since `setQueryData` is used on mutate and `onError`. The refetch _should_ just be setting the cached value to be the same as that returned by the query

packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
mark-prins: is that a `pessimisticMutate` 🤔

packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
```diff
@@ -1,60 +1,108 @@
...
+    onError: (_, __, context) => {
+      queryClient.setQueryData(['invoice', id], context.previousInbound);
+    },
+  });
+
+  const { isLoading: isAddingItem, mutateAsync: noOptimisticMutate } =
```
mark-prins, line 77: should that be:
```suggestion
  const { isLoading: isAddingItem, mutateAsync: nonOptimisticMutate } =
```
wasn't sure.

packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
```diff
@@ -1,60 +1,108 @@
...
 
-  const onChangeSortBy = (column: Column<InboundShipmentItem>) => {
-    dispatch(InboundAction.onSortBy(column));
+  const updateInvoice = async (patch: Partial<InboundShipment>) => {
+    return mutateAsync({ ...data, ...patch, items: [] });
   };
 
-  return { draft, save, dispatch, onChangeSortBy, sortBy: state.sortBy };
-};
+  const upsertItem = async (item: OutboundShipmentSummaryItem) => {
```
mark-prins, line 86: Is it time to create a type of `InboundShipmentSummaryItem`?

packages/invoices/src/InboundShipment/DetailView/SidePanel.tsx
mark-prins: that's a very neat idea!

packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEdit.tsx
```diff
@@ -67,7 +70,9 @@ export const InboundLineEdit: FC<InboundLineEditProps> = ({
     }
   });
 
-  return (
+  return loading ? (
+    <BasicSpinner />
```
mark-prins, line 74: would be good to add a text prop to the spinner so that this instance could say 'saving' rather than 'loading' - which could still be the default if no prop given

packages/invoices/src/InboundShipment/DetailView/reducer/reducer.ts
```diff
@@ -1,3 +1,4 @@
+import { groupBy } from './../../../../../common/src/utils/arrays/utils';
```
mark-prins, line 1: time to introduce `@common/utils` - if we weren't planning to replace with lodash 😃

mark-prins (approved):
👍  that's very nice - like how it is to use, and some parts of the code are much simpler 😁 

One extra, which I think you've made an issue for (or something like this) already, would be to debounce the mutations. I typed a sentence into the comment input, and it created a lot of mutation POSTs!

packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
joshxg: Oops sorry I didn't address this. We could just assume that the cache update was correct. I'm not sure I see the harm in refetching, though, just incase? For the case of adding items, this should do the side effects such as getting new pricing totals etc which maybe we should be doing in the optimistic update, also? I suppose it's also a way to update values all throughout the invoice if others are updating it also which could be useful? (also, harmful?)

mark-prins: Fine to refetch, as long as we debounce - currently there are a lot of mutate & query combos getting fired off! in most cases a quick cache update would be exactly the same as the response from the query, but you're right, it might not be always, and is safer to refetch.

PR #636, opened 2021-12-13
Group by item for outbound shipments

### LINKED ISSUE #311: List View: group by item

mark-prins:
Fixes #311 

The issue has been around for a while, so putting this out there to close it off. Doesn't add a huge amount of great functionality, though it does tidy up the columns for the outbound shipment view.

Was concerned that the default view for outbounds was the grouped view now, so have changed that around and used the row expanding version only if requested. Not retaining that state either.. possibly should.. decided it was ok like this for now.


packages/common/src/ui/components/inputs/Switch/Switch.tsx
```diff
@@ -0,0 +1,99 @@
+import React from 'react';
```
mark-prins, line 1: styled the switch to look like the version in zeplin. Not that keen on the square-ish border, that's personal preference I guess.

packages/common/src/ui/layout/tables/DataTable.tsx
```diff
@@ -69,9 +70,14 @@ export const DataTable = <T extends DomainObject>({
 
   return (
     <TableContainer
-      sx={{ display: 'flex', flexDirection: 'column', height: '100%' }}
+      sx={{
+        display: 'flex',
+        flexDirection: 'column',
+        height: '100%',
+        overflowX: 'unset',
```
mark-prins, line 77: was getting a vertical scroll showing up - and the horizontal scroll for the table was showing below the pagination. with this it shows at the bottom of the page which was neater looking - not fussed if you want to change back tho

packages/invoices/src/OutboundShipment/DetailView/tabs/GeneralTab.tsx
```diff
@@ -38,22 +80,80 @@ export const GeneralTabComponent: FC<
...
+
+      if (isOn) {
+        newGrouped.push({
+          ...row,
+          lineTotal,
+          sellPricePerUnit: lineTotal / row.unitQuantity,
+          batch: ifTheSameElseDefault(batches, 'batch', '[multiple]'),
+          expiryDate: ifTheSameElseDefault(batches, 'expiryDate', '[multiple]'),
+          canExpand: Object.keys(row.batches).length > 1,
+        });
```
joshxg, line 107: Since this is a summary item, the `batch` and `expiryDate` should already be set like this. The `sellPricePerUnit` and `lineTotal` and `canExpand` field could also just be added to `createSummaryItem` in the reducer file (and then would need to update `recalculateSummary` 🙄 ) but then if `isOn`, you don't do anything?

packages/invoices/src/OutboundShipment/DetailView/tabs/GeneralTab.tsx
```diff
@@ -38,22 +80,80 @@ export const GeneralTabComponent: FC<
...
+      } else {
+        batches.forEach(batch => {
+          newGrouped.push({
+            ...row,
+            batch: batch.batch,
+            expiryDate: batch.expiryDate,
+            packSize: batch.packSize,
+            lineTotal,
+            sellPricePerUnit: lineTotal / row.unitQuantity,
```
joshxg, line 116: This `lineTotal` is for all the lines of the item, rather than this specific line - is that intentional? 

Actually I think maybe the `lineTotal` column should just use an `accessor` function rather than trying to keep this field up to date? 

Should the `sellPricePerUnit` be coming from the API? 

I would have thought the `sellPricePerPack` would be more useful than the `sellPricePerUnit` which we aren't displaying?

joshxg (approved):
I did think we were just sticking with a single grouped view with the expand option - but this is looking great. Agree that retaining the state would be nice in local storage, but can wait and be a nice cherry on top later.

I am getting some rows as disabled after expanding/collapsing. There's several bugs in `TableContext.ts` where the `isDisabled` state is being set to the `isExpanded` state - i.e. https://github.com/openmsupply/openmsupply-client/blob/eca1959535d98dfb90ab280579ba70bfea5df310/packages/common/src/ui/layout/tables/context/TableContext.ts#L50

I also had a little bit of weird behaviour if I were to add a line to an item, then I wouldn't see the newly added item in the expando view but haven't had time to look deeper into that.

Maybe need to disable/remove the expando column when not grouped by item, also? Could put the 'grouped' state in the `TableContext` for an easy way to disable

packages/common/src/ui/components/inputs/Switch/Switch.tsx
mark-prins: The 'off' state I took from [the requisition filter design](https://app.zeplin.io/project/60e44fff6fa942168092f1ff/screen/619a9523bb78cca911bba916) in zeplin which is:

Screen Shot 2021-12-14 at 3 05 01 PM

I see now, that there's a more rounded version hiding in the background!

Screen Shot 2021-12-14 at 3 06 15 PM
 
Just no 'on' state.

packages/invoices/src/OutboundShipment/DetailView/tabs/GeneralTab.tsx
mark-prins: I realised that this morning! will do.

packages/invoices/src/OutboundShipment/DetailView/tabs/GeneralTab.tsx
joshxg: Hm that's weird .. wouldn't really expect anything other than the line total in the summary row!

packages/common/src/ui/components/inputs/Switch/Switch.tsx
joshxg: 🤣  I thought this was the 'on' state 😁

PR #637, opened 2021-12-13
Add sorting and remove reducer

### LINKED ISSUE #634: Example inbound shipments page without a save button

joshxg:
Fixes #634 

- This adds the sorting and a bit of reshuffling of things. 
- There's not much of a pattern yet to follow which is a bit disappointing. The whole `Api` interface with an `onRead` and `onUpdate` methods makes less sense now if we change to need more specific `onUpdateInvoice` `onUpsertLines` and the many other potential ones.. `onAllocatePlaceholderLines` etc


packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
```diff
@@ -1,110 +1,20 @@
...
-    upsertItem,
-    draft:
-      { ...data, items: inboundLinesToSummaryItems(data?.lines ?? []) } ??
-      placeholderInbound,
-
-    onChangeSortBy,
-  };
-};
+import { getNextInboundStatus, isInboundEditable } from '../../utils';
+import { InboundShipmentItem } from '../../types';
```
joshxg, line 17: Here I've taking the `useDraftInbound` hook into the `api` file. Though, the name draft inbound is not great anymore.

I've also moved the columns to `GeneralTab` and made a hook specifically for the lines

packages/invoices/src/InboundShipment/DetailView/GeneralTab.tsx
```diff
@@ -34,14 +36,34 @@ const Expand: FC<{ rowData: InboundShipmentItem }> = ({ rowData }) => {
...
 export const GeneralTabComponent: FC<GeneralTabProps<InboundShipmentItem>> = ({
-  data,
-  columns,
   onRowClick,
 }) => {
   const { pagination } = usePagination();
-  const activeRows = data.filter(({ isDeleted }) => !isDeleted);
   const t = useTranslation('common');
 
+  const { data, sortBy, onSort } = useInboundLines();
```
joshxg, line 44: this hook just returns the lines of the inbound lines which I thought was quite nice. Maybe should pass the `id` rather than `useParams`.

packages/invoices/src/InboundShipment/DetailView/api.ts
```diff
@@ -204,3 +218,121 @@ export const getInboundShipmentDetailViewApi = (
     throw new Error(':shrug');
   },
 });
+
+export const useInboundShipment = (): UseQueryResult<Invoice, unknown> => {
```
joshxg, line 222: at this point it's just a random assortment of hooks. They're nice and useful, but not a specific interface to follow which might not be such a bad thing, but it makes me sad

packages/invoices/src/InboundShipment/DetailView/api.ts
```diff
@@ -204,3 +218,121 @@ export const getInboundShipmentDetailViewApi = (
...
+
+  const upsertItem = async (item: OutboundShipmentSummaryItem) => {
+    const itemIdx = draft.items.findIndex(i => i.id === item.id);
+    if (itemIdx >= 0) draft.items[itemIdx] = item;
+    else draft.items.push(item);
+
+    const result = await mutateAsync(draft);
+
+    return result;
+  };
```
joshxg, line 330: I think what would be best would be to separate out the mutations into individual hooks and use them closer to where they are used.

For example, in the `SidePanel` component, using `useInboundSelector` to select the specific parts of the invoice that are being displayed, and `useUpdateInbound` to get the mutate function to update rather than passing down props to do all this.

But wanted to just wait and get some feedback first!

packages/invoices/src/InboundShipment/DetailView/api.ts
mark-prins: the hook is a great idea can see this being useful

packages/invoices/src/InboundShipment/DetailView/api.ts
```diff
@@ -204,3 +218,121 @@ export const getInboundShipmentDetailViewApi = (
...
+      return { previousInbound, patch };
+    },
+    onSettled: () => queryClient.invalidateQueries(['invoice', id]),
+    onError: (_, __, context) => {
+      queryClient.setQueryData(['invoice', id], context.previousInbound);
+    },
+  });
+};
+
+export const useInboundLines = () => {
```
mark-prins, line 274: also very nice! 😁

packages/invoices/src/InboundShipment/DetailView/api.ts
```diff
@@ -204,3 +218,121 @@ export const getInboundShipmentDetailViewApi = (
...
+    },
+    [sortBy]
+  );
+
+  const { data } = useInboundShipmentSelector(selectItems);
+
+  return { data, sortBy, onSort };
+};
+
+export const useDraftInbound = () => {
```
mark-prins, line 301: I'm fine with continuing to use `draft` even though the status is `NEW` now. the term draft still makes sense, don't think there's a need to go changing it all..

packages/invoices/src/InboundShipment/DetailView/api.ts
mark-prins: yes! agree on both counts - would be great to separate out the mutations, and yeah, good to wait for feedback

packages/invoices/src/InboundShipment/DetailView/GeneralTab.tsx
mark-prins: maybe should.. but the `useParams` is so clever, shame not to use that approach. and it is neater to just call `useInboundLines()` with no params 😁

mark-prins (approved):
👍  looking good!
while its a shame to say goodbye to the reducer, this is all clear and easy to read. simpler overall, I think

packages/invoices/src/InboundShipment/DetailView/GeneralTab.tsx
joshxg: I'm actually realising that it might also be better to *only* keep the `sort`, `filter` and `pagination` states in the URL and `useParams` for that also.. 🤔

PR #638, opened 2021-12-13
Add auto heigh/width for paper popover when not specified

### LINKED ISSUE #628: Footer status popup
Couple of minor things arising from the status changes.

1. The popover is flashy - I have the mouse in the same spot, but the hover cycles between focus and blur
poppy status

2. The popup could do with being bigger now that there is more in it
Screen Shot 2021-12-10 at 11 04 51 AM

joshxg:
Fixes #628 

- Just added auto height/width when not specified


mark-prins (approved):
😁  that does the business!

PR #646, opened 2021-12-15
#645 debounced mutations

### LINKED ISSUE #645: Inbound shipment debounce optimistic mutations
Need to debounce the optimistic mutations on the inbound shipment detail view

joshxg:
Fixes #645 

- Went a bit further and was investigating some different ways to structure everything. Haven't got very far with that, but have added a hook `useInboundFields`. The idea being to get away from drilling the invoice down to every nook and cranny and instead, essentially use a selector hook closer to where data is used. This makes for a pretty clean and easy way to use the 'document'. I think the API is nice being able to destructure the values you ask for, though it could be going too far.
- I've added a 'buffer' state hook and debounced the api call. This means that inputs correctly 'update' their state, but the mutation is debounced. The problems I was trying to solve with this was that we could potentially have 'updates' come from the server (i.e. someone else editing the invoice), which means we can't use a 'default value'/'internal state' for the text inputs and also if we trigger the mutation on `blur` there is a race condition on pushing 'next status'. The second problem is still there and might mean that we should reduce the debounce timer. Alternatives include disabling the status change button by tracking focus or the debounce/mutation 'state' which I'd rather avoid but it might be inevitable as the debounce timer is a likely unsafe/inaccurate - but.. might do for now.

### DIFF
--- a/packages/common/src/hooks/useBufferState/useBufferState.ts
+++ b/packages/common/src/hooks/useBufferState/useBufferState.ts
@@ -1,0 +1,21 @@
+import { useEffect, useState, Dispatch, SetStateAction } from 'react';
+
+// It is generally discouraged to sync props and state. Generally, you should just use
+// props. Other times, you may want to seed some state with props (useState(props.value))
+// it is rare that you should try to sync state and props - essentially, this is the same
+// as seeding your state with props, however, new prop values can update the state. This occurs
+// when there is some other side effect that is able to update the content of some component -
+// i.e. if a network request results in different values for some prop than what a component
+// was initially mounted with. If you can, try to avoid using this hook and find a different way
+// but if there isn't one, here you go!
+export const useBufferState = <T>(
+  value: T
+): [T, Dispatch<SetStateAction<T>>] => {
+  const [buffer, setBuffer] = useState(value);
+
+  useEffect(() => {
+    setBuffer(value);
+  }, [value]);
+
+  return [buffer, setBuffer];
+};
--- a/packages/common/src/ui/components/inputs/TextArea/TextArea.tsx
+++ b/packages/common/src/ui/components/inputs/TextArea/TextArea.tsx
@@ -9,33 +9,21 @@ export const TextArea: FC<StandardTextFieldProps> = ({
   minRows = 4,
   InputProps,
   ...props
-}) => {
-  const [buffer, setBuffer] = React.useState(value);
-
-  React.useEffect(() => {
-    setBuffer(value);
-  }, [value]);
-
-  return (
-    <BasicTextInput
-      sx={{ width: '100%' }}
-      InputProps={{
-        ...InputProps,
-        sx: {
-          backgroundColor: 'white',
-          ...InputProps?.sx,
-        },
-      }}
-      multiline
-      value={buffer}
-      onChange={e => {
-        const newVal = e.target.value;
-        onChange?.(e);
-        setBuffer(newVal);
-      }}
-      maxRows={maxRows}
-      minRows={minRows}
-      {...props}
-    />
-  );
-};
+}) => (
+  <BasicTextInput
+    sx={{ width: '100%' }}
+    InputProps={{
+      ...InputProps,
+      sx: {
+        backgroundColor: 'white',
+        ...InputProps?.sx,
+      },
+    }}
+    multiline
+    value={value}
+    onChange={onChange}
+    maxRows={maxRows}
+    minRows={minRows}
+    {...props}
+  />
+);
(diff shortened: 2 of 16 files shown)


packages/invoices/src/InboundShipment/DetailView/api.ts
```diff
@@ -271,6 +283,72 @@ const useOptimisticInboundUpdate = () => {
...
+      if (Array.isArray(keyOrKeys)) {
+        const mapped = keyOrKeys.reduce((acc, val) => {
+          acc[val] = invoice[val];
+          return acc;
+        }, {} as { [k in KeyOfInvoice]: Invoice[k] });
+
+        return mapped;
+      } else {
+        return { [keyOrKeys]: invoice[keyOrKeys] } as {
+          [k in KeyOfInvoice]: Invoice[k];
```
joshxg, line 308: couldn't get this to work without the assertions, unfortuantely!

packages/common/src/hooks/useBufferState/useBufferState.ts
```diff
@@ -0,0 +1,21 @@
+import { useEffect, useState, Dispatch, SetStateAction } from 'react';
+
+// It is generally discouraged to sync props and state. Generally, you should just use
+// props. Other times, you may want to seed some state with props (useState(props.value))
+// it is rare that you should try to sync state and props - essentially, this is the same
+// as seeding your state with props, however, new prop values can update the state. This occurs
+// when there is some other side effect that is able to update the content of some component -
+// i.e. if a network request results in different values for some prop than what a component
+// was initially mounted with. If you can, try to avoid using this hook and find a different way
```
mark-prins, line 9: excellent description and the warning is spot on

packages/common/src/ui/components/inputs/TextInput/BufferedTextInput.tsx
```diff
@@ -0,0 +1,31 @@
...
+}) => {
+  const [buffer, setBuffer] = useBufferState(value);
+
+  return (
+    <BasicTextInput
+      {...rest}
+      value={buffer}
+      onChange={e => {
+        setBuffer(e.target.value);
+        onChange && onChange(e);
```
mark-prins, line 27: or `onChange?.(e)` to match the above?

packages/common/src/ui/components/inputs/TextArea/TextArea.tsx
```diff
@@ -9,21 +9,33 @@ export const TextArea: FC<StandardTextFieldProps> = ({
...
-    multiline
-    value={value}
-    onChange={onChange}
-    maxRows={maxRows}
-    minRows={minRows}
-    {...props}
-  />
-);
+}) => {
+  const [buffer, setBuffer] = React.useState(value);
```
mark-prins, line 13: Is this now `BufferedTextArea`? I feel like it deserves to be renamed with these changes

packages/invoices/src/InboundShipment/DetailView/SidePanel.tsx
```diff
@@ -14,15 +14,16 @@ import {
...
+import { useInboundFields, useIsInboundEditable } from './api';
 
 interface SidePanelProps {
   draft: InboundShipment;
-  update: (patch: Partial<InboundShipment>) => Promise<InboundShipment>;
 }
 
-const AdditionalInfoSection: FC<SidePanelProps> = ({ draft, update }) => {
+const AdditionalInfoSection: FC = () => {
+  const { comment, update } = useInboundFields('comment');
```
mark-prins, line 25: ...and there's the payoff! that's very neat

mark-prins (approved):
👍  comprehensively done! 
that's looking good - and the hooks to get and update draft properties is very neat. We may end up with a lot of props getting destructured and have to revisit.. I think that mostly there's not many that we need, so that should work well.

packages/common/src/ui/components/inputs/TextArea/TextArea.tsx
joshxg: I've added a whole new one, though that's probably overkill

packages/common/src/ui/components/inputs/TextInput/BufferedTextInput.tsx
joshxg: Ah yes - done!

joshxg:
> that's looking good - and the hooks to get and update draft properties is very neat. We may end up with a lot of props getting destructured and have to revisit.. I think that mostly there's not many that we need, so that should work well.

There is still the `useInboundShipment` hook which is just the normal `useQuery` with the inbound shipment to use when needed. Using the selector, the components using the selector should only be re-rendering when the fields change that are passed which is also a nice bonus

PR #648, opened 2021-12-15
#643 outbound detail grouping

### LINKED ISSUE #643: Fix outbound detail grouping
Issue to address the issues in the [outbound grouping PR](https://github.com/openmsupply/openmsupply-client/pull/636)

- [x] make the Switch component look better ( see [comment](https://github.com/openmsupply/openmsupply-client/pull/636#discussion_r768262018)
  - [x] tidy the border radius
  - [x] try a light gray version
- [x] retain state in local storage?
- [x] disabled state ( see [comment](https://github.com/openmsupply/openmsupply-client/pull/636#pullrequestreview-829842658) )
- [x] remove expando column when not grouped ( put 'grouped' state into the table context )
- [x] in `GeneralTab` change `palette['gray']['light']` to `palette.gray.light`
- [x] review the `noLines` prop
- [ ] Styled component for the Switch?
- [x] `GeneralTab` the `DataTable` should be using `activeRows` (x2)
- [x] memoise `activeRows`?
- [x] update summary item in reducer [see comment](https://github.com/openmsupply/openmsupply-client/pull/636#discussion_r767528213)
- [x] the lineTotal column should just use an accessor function

mark-prins:
Fixes #643 

I tried using `styled` for the Switch - couldn't get the transition working, when using styled it was always re-rendering instead. So I did what any normal person would do.. and gave up 😄


packages/common/src/styles/theme.ts
```diff
@@ -50,6 +50,12 @@ declare module '@mui/material/Checkbox' {
   }
 }
 
+declare module '@mui/material/Switch' {
```
mark-prins, line 53: not strictly necessary now, but neat. and someone may want to use it!

packages/common/src/ui/components/inputs/Switch/Switch.tsx
```diff
@@ -50,25 +68,28 @@ export const Switch: React.FC<SwitchProps> = ({
...
-      borderColor: 'border',
+      borderColor: 'gray.main',
       backgroundColor: '#fff',
+      borderRadius: isSmall ? '11px' : '13px',
       transition: (theme: Theme) =>
         theme.transitions.create(['background-color'], {
           duration: 500,
         }),
     },
+    '& .Mui-checked ~ .MuiSwitch-track': {
```
mark-prins, line 90: sibling selector - was using `checked` but that only works for controlled components!

packages/common/src/ui/layout/tables/DataTable.tsx
```diff
@@ -100,7 +99,7 @@ export const DataTable = <T extends DomainObject>({
                 onClick={onRowClick}
                 rowData={row}
                 rowKey={String(idx)}
-                noLines={noLines}
+                dense={dense}
```
mark-prins, line 102: removed the lines - but also reduced font and padding, to make the dense version denser. I think it was a bit too much with all the lines

packages/invoices/src/InboundShipment/DetailView/SidePanel.tsx
```diff
@@ -106,7 +106,7 @@ const RelatedDocumentsSection: FC<SidePanelProps> = () => {
 
 export const SidePanel: FC<SidePanelProps> = ({ draft }) => {
   const { success } = useNotification();
-  const t = useTranslation(['outbound-shipment', 'common']);
+  const t = useTranslation(['distribution', 'common']);
```
mark-prins, line 109: offroad fix

packages/invoices/src/OutboundShipment/DetailView/reducer.ts
```diff
@@ -51,16 +51,29 @@ const getExistingLine = (
...
   summaryItem: OutboundShipmentSummaryItem
-): { unitQuantity: number; numberOfPacks: number } => {
-  const unitQuantity = Object.values<OutboundShipmentRow>(
-    summaryItem.batches
-  ).reduce(getUnitQuantity, 0);
-
-  const numberOfPacks = Object.values<OutboundShipmentRow>(
-    summaryItem.batches
-  ).reduce(getSumOfKeyReducer('numberOfPacks'), 0);
+): Partial<OutboundShipmentSummaryItem> => {
```
mark-prins, line 54: this was why the `[multiple]` wasn't working - the summary item was being created before all the batches were added and then not recalculated after they were. Moved that into the reducer and out of GeneralTab; likely will need to repeat this for inbound & reqs & stocktakes.

packages/invoices/src/OutboundShipment/DetailView/tabs/GeneralTab.tsx
```diff
@@ -60,14 +58,14 @@ const Expand: FC<{
   if (!rowData?.canExpand) return <></>;
 
   return (
-    <Box p={1} style={{ padding: '0 50px 0 200px' }}>
+    <Box p={1} style={{ padding: '0 100px 0 100px' }}>
```
mark-prins, line 61: I didn't like it without some padding, the full width version looks odd to me. I also didn't like having the left hand edge more left than the first column in the summary, but agree that it isn't great. thought it was slightly better when centred?

packages/mock-server/src/data/data.ts
```diff
@@ -585,7 +585,10 @@ const createRequisition = (
     maxMOS: 3,
     thresholdMOS: 3,
     color: '#8f90a6',
-    status: SupplierRequisitionNodeStatus.Draft,
+    status: takeRandomElementFrom([
```
mark-prins, line 588: offroad, just to make testing reqs a bit easier.

packages/invoices/src/OutboundShipment/DetailView/tabs/GeneralTab.tsx
```diff
@@ -79,36 +77,33 @@ const Expand: FC<{
...
-  const { pagination } = usePagination();
-  const { isOn, toggle } = useToggle();
-
-  const [grouped, setGrouped] = React.useState<OutboundShipmentSummaryItem[]>(
-    []
+  const [storedIsGrouped, setStoredIsGrouped] = useLocalStorage(
+    '/outboundshipment/groupbyitem'
   );
+  const { pagination } = usePagination();
+  const [isGroupedByItem, setIsGroupedByItem] = useState(storedIsGrouped);
```
joshxg, line 84: Is this needed? We can't just use `storedIsGrouped`?

packages/invoices/src/OutboundShipment/DetailView/tabs/GeneralTab.tsx
```diff
@@ -117,13 +112,23 @@ export const GeneralTabComponent: FC<
...
+    }
+  }, [isGroupedByItem, data]);
 
   const t = useTranslation('distribution');
-  const activeRows = grouped.filter(({ isDeleted }) => !isDeleted);
+  const activeRows = useMemo(
+    () => grouped.filter(({ isDeleted }) => !isDeleted),
+    [grouped]
+  );
+  const { setIsGrouped } = useTableStore();
```
joshxg, line 125: The column header isn't showing when navigating with the persisted `isGrouped` as true

A little effect should handle it

```
useEffect(() => {
    setIsGrouped(storedIsGrouped);
  }, [storedIsGrouped, setIsGrouped]);
```

Could then also remove this from the toggle fn also, I think

joshxg (approved):
It's really great. It's a tough one to make, that's for sure. I'm a little bit concerned about the annoyance that is the summary item but what you've done is awesome

Have another nit about the columns being 'sortable' but not being 'hooked' up. I think either we can pass columns like `['batch', {sortable: false}]`, we could add into `options` a `sortableOverride` field and use that in `createColumnWithDefaults`.. or maybe sortable default should be set as `onChangeSortBy && sortBy`.

Aaand for the future: 
- Wondered if the 'row click' should also apply to the expansion. 
- A bit annoying that the `Header` and `Cell` components of a column aren't defining their own widths? If it was, we could change the width based on being grouped (to essentially make the column invisible 🤔 )

packages/invoices/src/OutboundShipment/DetailView/tabs/GeneralTab.tsx
mark-prins: thanks. yep - have fixed by removing the `100%` width on the grid, which was constraining it; that was in there to get the switch to sit on the RH side, so I've moved that over to the left, which I don't like as much, though it does fix the problem of the switch sitting in the middle when the table stretches and you scroll over a bit.

packages/common/src/localStorage/keys.ts
```diff
@@ -3,6 +3,7 @@ import { SupportedLocales } from '@common/intl';
 export type LocalStorageRecord = {
   '/appdrawer/open': boolean;
   '/localisation/locale': SupportedLocales;
+  '/outboundshipment/groupbyitem': boolean;
```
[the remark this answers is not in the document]
mark-prins: like the serialised though. went with it

PR #651, opened 2021-12-15
remove counts from mocks

### LINKED ISSUE #649: Update schema for stock counts
The stock counts changes were merged to in the API, time to update. Currently getting
`Error: Unable to merge GraphQL type "Queries": Field "stockCounts" already defined with a different type. Declared as "StockCounts", but you tried to override with "StockCountsResponse` from the mock server

mark-prins:
Fixes #649

### DIFF
--- a/packages/common/src/types/schema.ts
+++ b/packages/common/src/types/schema.ts
@@ -195,11 +195,6 @@ export type ConnectorErrorInterface = {
   description: Scalars['String'];
 };
 
-export type CountError = {
-  __typename?: 'CountError';
-  description: Scalars['String'];
-};
-
 export enum CustomerRequisitionNodeStatus {
   Finalised = 'FINALISED',
   InProgress = 'IN_PROGRESS',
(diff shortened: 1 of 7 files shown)


packages/common/src/types/schema.ts
```diff
@@ -419,6 +414,12 @@ export type DeleteSupplierRequisitionResponseWithId = {
   response?: Maybe<DeleteSupplierRequisitionResponse>;
 };
 
+export type EqualFilterBigNumberInput = {
+  equalAny?: InputMaybe<Array<Scalars['Int']>>;
+  equalTo?: InputMaybe<Scalars['Int']>;
+  notEqualTo?: InputMaybe<Scalars['Int']>;
+};
```
joshxg, line 421: Are we not allowed to use little numbers? So many limitations of this api 😱

joshxg (approved):
Nice, thanks

packages/common/src/types/schema.ts
mark-prins: 😄  I missed that! Is `Int` a 'BigNumber' then?

PR #653, opened 2021-12-15
ignore status in shipment updates

### LINKED ISSUE #650: OS/IS: Update mutation shouldn't set status
Inbound & Outbound shipments aren't updating because they are trying to set the status, and only prescribed values are allowed for the status field, so the update fails.

The update mutation now allows partials updates, we can send only modified fields

mark-prins:
Fixes #650 

I now have a problem with focus - when you start typing into a field, pause, and the page rerenders the input, you lose focus and have to click in the text area/field again which is super annoying. have left that little issue for now.

### DIFF
--- a/packages/common/src/utils/arrays/utils.ts
+++ b/packages/common/src/utils/arrays/utils.ts
@@ -26,7 +26,7 @@ export const groupBy = <T, K extends keyof T & string>(
   key: K
 ): Record<string, T[]> => {
   const lookup: Record<string, T[]> = {} as Record<string, T[]>;
-  things.forEach(thing => {
+  things?.forEach(thing => {
     const value = String(thing[key]);
     if (!lookup[value]) {
       lookup[value] = [];
--- a/packages/invoices/src/InboundShipment/DetailView/api.ts
+++ b/packages/invoices/src/InboundShipment/DetailView/api.ts
@@ -1,5 +1,6 @@
 import { useCallback } from 'react';
 import {
+  InvoiceNodeStatus,
   UpdateInboundShipmentMutation,
   useParams,
   useOmSupplyApi,
@@ -89,6 +90,17 @@ const stockLineGuard = (stockLine: StockLineResponse): StockLineNode => {
   throw new Error('Unknown');
 };
 
+const getPatchStatus = (patch: Partial<Invoice>) => {
+  switch (patch.status) {
+    case InvoiceNodeStatus.Verified:
+      return UpdateInboundShipmentStatusInput.Verified;
+    case InvoiceNodeStatus.Delivered:
+      return UpdateInboundShipmentStatusInput.Delivered;
+    default:
+      return undefined;
+  }
+};
+
 const invoiceToInput = (
   patch: Partial<Invoice> & { id: string }
 ): UpdateInboundShipmentInput => {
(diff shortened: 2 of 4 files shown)


packages/common/src/utils/arrays/utils.ts
```diff
@@ -26,7 +26,7 @@ export const groupBy = <T, K extends keyof T & string>(
   key: K
 ): Record<string, T[]> => {
   const lookup: Record<string, T[]> = {} as Record<string, T[]>;
-  things.forEach(thing => {
+  things?.forEach(thing => {
```
mark-prins, line 29: offroad fix - was throwing an error if there were no lines

packages/invoices/src/OutboundShipment/DetailView/api.ts
```diff
@@ -97,19 +98,30 @@ export const onRead =
     };
   };
 
+const getPatchStatus = (patch: Partial<OutboundShipment>) => {
```
mark-prins, line 101: yeah. but it's temporary, right?

packages/invoices/src/OutboundShipment/DetailView/tabs/GeneralTab.tsx
```diff
@@ -96,12 +96,12 @@ export const GeneralTabComponent: FC<
   const toggleGrouped = () => {
     setIsGroupedByItem({
       ...isGroupedByItem,
-      outboundShipment: !isGroupedByItem.outboundShipment,
+      outboundShipment: !isGroupedByItem?.outboundShipment,
```
mark-prins, line 99: and the other offroad. oops!

packages/invoices/src/InboundShipment/DetailView/api.ts
```diff
@@ -325,10 +325,7 @@ export const useInboundFields = <KeyOfInvoice extends keyof Invoice>(
           id,
         ]);
 
-        queryClient.setQueryData(['invoice', id], {
-          ...previousInbound,
-          ...patch,
-        });
+        queryClient.setQueryData(['invoice', id], patch);
```
joshxg, line 328: This is causing the loss of focus - The query data becomes just a partial inbound like `{ id, comment}`, which the whole page reacts to - there's no lines, so the table goes away, there's no other party so the other party input goes away then the status isn't an editable status, so the whole form is disabled causing the loss of focus. I've put it back to what is was previously and it's working fine and updating - but adding items isn't as that uses the `upsert` mutation which includes updating the shipment including the status..!

packages/invoices/src/OutboundShipment/DetailView/api.ts
joshxg: Yup! 😁

joshxg (approved):
I added this commit 535c53c

which is working well for me against the demo server api. Though also temporary as we can't change status with it!

packages/common/src/utils/arrays/utils.ts
mark-prins: lucky me 🤷  Am using the demo server API and one of the invoices ( inbound, I think ) which has no lines was giving an error. just a console error, so no harm done.

packages/invoices/src/InboundShipment/DetailView/api.ts
mark-prins: oh! 🤦  thanks

PR #654, opened 2021-12-15
Remove info column

### LINKED ISSUE #647: Remove outbound info column on the item details modal
The `info` column with the popover can be removed

joshxg:
Fixes #647 

🤷 

Pretty confident this will make a comeback when the API has more available data, but for now..


mark-prins (approved):
👍  that does it! Yep, likely to return in some form..

PR #656, opened 2021-12-16
#655 dont update inbound when upserting items

### LINKED ISSUE #655: Inbound upsert items mutation shouldn't also update the shipment itself
So the batch mutation we are using right now for items is also updating the shipment itself. It doesn't need to.

I think with the new react-query hooks we might be able to simplify the modal component a little as well which I might do a little offroading to investigate..

joshxg:
Fixes #655 

- Bit big sorry, though a lot of the diff is that damn schema.
- Went a bit far changing things up, but it seems to make things easier..
- Basically changed things up to use a lot more react-query based hooks rather than passing props around and putting a lot of the callbacks etc into the `InboundLineEdit` component rather than the parent `DetailView`

### DIFF
--- a/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEdit.tsx
+++ b/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEdit.tsx
@@ -128,7 +128,21 @@ const useDraftInboundLines = (itemId: string) => {
         key: K,
         value: DraftInboundLine[K]
       ) => {
-        updateDraftLine({ id: line.id, [key]: Number(value) });
+        const parseValue = (value: DraftInboundLine[K]) => {
+          switch (key) {
+            case 'numberOfPacks':
+              return Number(value);
+            case 'packSize':
+              return Number(value);
+            case 'sellPricePerkPack':
+              return Number(value);
+            case 'costPricePerPack':
+              return Number(value);
+            default:
+              return value;
+          }
+        };
+        updateDraftLine({ id: line.id, [key]: parseValue(value) });
       };
       return line;
     }),
--- a/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEdit.tsx
+++ b/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEdit.tsx
@@ -1,4 +1,4 @@
-import React, { FC, useEffect } from 'react';
+import React, { FC, useState, useEffect } from 'react';
 import {
   Divider,
   TableContainer,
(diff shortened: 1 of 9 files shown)


packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEdit.tsx
```diff
@@ -14,130 +14,252 @@ import {
...
+      }
+      okButton={<DialogButton variant="ok" onClick={saveLines} />}
+      height={600}
+      width={1024}
+    >
+      {isLoading ? (
+        <BasicSpinner />
       ) : (
-        <Box sx={{ height: isMediumScreen ? 400 : 500 }} />
+        <>
```
joshxg, line 205: Not much actually changed here, mainly the diff is from adding the `Modal`

packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEdit.tsx
```diff
@@ -14,130 +14,252 @@ import {
...
+  | 'location'
+  | 'numberOfPacks'
+  | 'packSize'
+  | 'itemId'
+  | 'invoiceId'
+> & {
+  isCreated?: boolean;
+  isUpdated?: boolean;
+  update?: (key: keyof DraftInboundLine, value: any) => void;
+};
```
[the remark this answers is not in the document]
mark-prins: not confusing yet. At some point we should write this up though as a pattern to help devs. I might add another readme or two

packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEdit.tsx
```diff
@@ -14,130 +14,266 @@ import {
...
+      );
+      if (drafts.length === 0) drafts.push(createDraftInvoiceLine(itemId, id));
+      setDraftLines(drafts);
+    } else {
+      setDraftLines([]);
+    }
+  }, [lines, itemId]);
+
+  const addDraftLine = () => {
+    const newBatch = createDraftInvoiceLine(itemId, id);
```
mark-prins, line 105: `newLine` would be more consistent with the 'lines' naming

packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEdit.tsx
```diff
@@ -14,130 +14,266 @@ import {
...
+            }
+          }}
+        />
+      }
+      okButton={<DialogButton variant="ok" onClick={saveLines} />}
+      height={600}
+      width={1024}
+    >
+      {isLoading ? (
+        <BasicSpinner />
```
mark-prins, line 203: still bothers me that the spinner says 'loading' when it is saving. what about a 'mode' prop on the spinner to indicate the action you are waiting on? In this modal we are only waiting for save, not load, so that should be easy enough to use

packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEdit.tsx
```diff
@@ -14,130 +14,266 @@ import {
...
+              await saveLines();
+              setCurrentItem(mode === ModalMode.Update ? nextItem : null);
+              return true;
+            } catch (e) {
+              return false;
+            }
+          }}
+        />
+      }
+      okButton={<DialogButton variant="ok" onClick={saveLines} />}
```
mark-prins, line 198: The ok behaviour isn't what I expect - is this just me?
Currently, OK saves the lines. I would expect it to save and close the modal. If the button only saves, I expect it to say `Save`.
Also - there is a cancel button which is used to close, which means if I want to save and then close, I have to click `Ok` and then `Cancel` and in my mind, the cancel is used to invalidate something, so that process flow feels wrong.

I'd prefer the `Ok` to save and close. and the `Cancel` to abandon changes and close.
Otherwise, the `Ok` should say `Save` and `Cancel` should say `Close` - but that might be confusing with the detail view which saves without clicking save and so `Close` has a different meaning there.. or would, if we had one!

mark-prins (approved):
🎉  that's very neat! - almost time to update the `Outbound` now as well.. maybe after the partial PR which I also saw come through 😄

packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEdit.tsx
joshxg: Ah yes, sorry I missed this behaviour. Should be good in the upcoming PRs, I think!

packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEdit.tsx
joshxg: Yes, sorry not forgotten (maybe a little!!). Will do!

PR #658, opened 2021-12-17
#callbacks should use patches

### LINKED ISSUE #657: Cell setters should use a patch rather than key, value pairs
Right now we have `Cell`s which use callbacks for updating values that are key/value pairs.

It's much easier and typesafe to use a patch `Partial<T>` variable.

I think as much as I like having the setter attached to the data, it's to much overhead to be worthwhile. So, I think that an alternative is needed.

Adding a `setter` proper to a `Column` definition: `type ColumnSetter<T> = (patch: Partial<T>) => void` then we can just use it like 

`const columns = useColumns(["numberOfPacks",  { setter: (patch) => updateLine(patch)])`

joshxg:
Fixes #657 

- Added a prop to `Column<T>` type, `setter` where the intention is that columns which update things/are editable should be passed a `setter` override when rendering
- Took the updater functions off the shipment types as not needed now
- Updated cells etc to use the setter function of the column


mark-prins (approved):
👍  much better thanks!

PR #661, opened 2021-12-19
#249 url is deadend

### LINKED ISSUE #249: Bug: customers url is a deadend
At the moment theres nothing in the `/customers` url path

I got here by clicking the breadcrumb

image

image

Should we remove the `/customers` path? We could just use `outbound-shipment` (or `customer-invoice`)?

mark-prins:
Fixes #249 

Tidied the behaviour of the hover for the left nav, as that was annoying me. This might be more annoying 🤷  see what you think..
- hover over is the same
- hover out no longer collapses the menu; this avoids the jumping around
- when a menu item is clicked the state is reset and so any menu sections which were expanded due to the hover, are collapsed
- if the left nav is closed (ie the small screen view) then the hover still opens it, and when you aren't hovering over any menu items, it collapses again

### DIFF
--- a/packages/common/src/hooks/useDrawer.ts
+++ b/packages/common/src/hooks/useDrawer.ts
@@ -2,15 +2,16 @@ import create from 'zustand';
 import LocalStorage from '../localStorage/LocalStorage';
 
 type DrawerController = {
-  hoverActive: Record<string, boolean>;
+  hoverActive: Record<string, string>;
   hoverOpen: boolean;
   isOpen: boolean;
   hasUserSet: boolean;
   open: () => void;
   close: () => void;
   toggle: () => void;
-  setHoverActive: (key: string, active: boolean) => void;
+  setHoverActive: (key: string, active: string) => void;
   setHoverOpen: (open: boolean) => void;
+  clearHoverActive: () => void;
 };
 
 export const useDrawer = create<DrawerController>(set => {
@@ -30,6 +31,7 @@ export const useDrawer = create<DrawerController>(set => {
     close: () => set(state => ({ ...state, isOpen: false, hoverOpen: false })),
     toggle: () =>
       set(state => ({ ...state, isOpen: !state.isOpen, hasUserSet: true })),
+    clearHoverActive: () => set(state => ({ ...state, hoverActive: {} })),
   };
 });
 
--- a/packages/common/src/hooks/useDrawer.ts
+++ b/packages/common/src/hooks/useDrawer.ts
@@ -2,14 +2,14 @@ import create from 'zustand';
 import LocalStorage from '../localStorage/LocalStorage';
 
 type DrawerController = {
-  hoverActive: Record<string, string>;
+  hoverActive: Record<string, boolean>;
   hoverOpen: boolean;
   isOpen: boolean;
   hasUserSet: boolean;
   open: () => void;
   close: () => void;
   toggle: () => void;
-  setHoverActive: (key: string, active: string) => void;
+  setHoverActive: (key: string, active: boolean) => void;
   setHoverOpen: (open: boolean) => void;
   clearHoverActive: () => void;
 };
--- a/packages/common/src/intl/locales/en/inventory.json
+++ b/packages/common/src/intl/locales/en/inventory.json
@@ -1,12 +1,13 @@
 {
-    "label.stocktake-date": "Stocktake Date",
-    "label.suggested": "Suggested",
-    "label.finalised": "Finalised",
-    "label.counted-num-of-packs": "Counted # of Packs",
-    "label.add-batch": "Add batch",
-    "label.add-new-line": "Add a new line",
-    "button.delete-lines": "Delete lines",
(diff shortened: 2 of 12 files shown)


packages/common/src/hooks/useDrawer.ts
```diff
@@ -2,15 +2,16 @@ import create from 'zustand';
 import LocalStorage from '../localStorage/LocalStorage';
 
 type DrawerController = {
-  hoverActive: Record<string, boolean>;
+  hoverActive: Record<string, string>;
```
mark-prins, line 5: now storing the three states of 'hover over', 'hover out' and 🤷 
this allows the sub menu open state to be less directly connected to the hover state

packages/common/src/intl/locales/en/inventory.json
```diff
@@ -1,12 +1,13 @@
...
-    "label.counted-num-of-packs": "Counted # of Packs",
-    "label.add-batch": "Add batch",
-    "label.add-new-line": "Add a new line",
-    "button.delete-lines": "Delete lines",
-    "message.cant-delete-stocktakes": "Can only delete stocktakes with a status of 'Suggested'",
-    "message.deleted-stocktakes": "Deleted {{number}} stocktakes",
-    "message.select-rows-to-delete": "Select rows to delete"
-}
\ No newline at end of file
+  "dev.log-draft": "Log draft",
```
mark-prins, line 2: noticed that `log-draft` wasn't being translated for all pages

packages/common/src/ui/components/navigation/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -26,14 +26,16 @@ export const Breadcrumbs: React.FC = () => {
...
+    parts.reduce((fullPath, part, index) => {
       if (part === '') return '';
       const path = `${fullPath}/${part}`;
-      urlParts.push({
-        path,
-        key: `${part}` as unknown as LocaleKey,
-        value: part,
-      });
+
+      if (index > 1)
```
mark-prins, line 33: skip first level, which is what the issue was _actually_ about

packages/common/src/ui/components/navigation/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -43,7 +45,7 @@ export const Breadcrumbs: React.FC = () => {
     if (index === urlParts.length - 1) {
       const title = /^\d+$/.test(part.value)
         ? t('breadcrumb.item', { id: part.value })
-        : t(part.key, '');
+        : t(part.key);
```
mark-prins, line 48: now displaying the uuids (e.g. requisition detail view)

joshxg (approved):
Nice thanks for finding the time to address this one - I think you know and they're just super annoying to fix, but i'll jot them down below. I think the behaviour is way better and anything else could wait.

- I'm not able to select a lot of pages if I'm on a tablet (i.e. if I'm on the dashboard I can't navigate anywhere) as I can't hover the nav menu items, so they don't open. 

For this, I think we'll have to turn the top level menu items into buttons which expand/collapse the sub menus? I think that fixes a few things, primarily that the top level menu items look like buttons and are pressable but don't do anything (I think if we do this then making them look different would also be good - maybe just a chevron heh)

- The nav items with no sub menus don't exapand but collapse the drawer when it is expanded which is a bit unexpected 😁  Also when going toward the bottom menu items if I am hovering the drawer but not the nav items I often can't get to the buttons in time.

I'm not sure what to do about this other than maybe having these menu items follow the same behaviour of causing the expand etc. 

For the hover, I experimented a little with just using css on the `StyledDrawer` component:

```
'&:hover': {
    ...openedMixin(theme),
  },
  '&:not(:hover)': {
    ...closedMixin(theme),
  },
```

It's a lot smoother and I think this behaviour is a bit nicer and less jerky but of course introduces other problems... mainly that the menu items need to change also. 

The only way I can think of for that would be to use explicit `ids` and to essentially hide/show components with css based on the css hover state of the drawer.. like

```
&:hover: {
    .someNavItemSmall: display: none
    .someNavItemBig: display: block
}
```

or something like that.. I think it might be worthwhile to just hard code ids and use pure css.. I also have been getting quite frustrated 😁 

The other issue would be that the hover state changing the width shouldn't happen when the drawer is 'expanded'..

mark-prins:
thanks for the feedback - and prompt to sort out the nav for mobile! have had that on my list for ages and kinda forgot about it.

f4e401ed

Have added an onClick for the nav items, so you can use them on mobile. have also added the hover and instituted a class-based hover-ness to get the drawer hover smoother. It is still possible to make it janky if you try - overall though it's much easier to use.

Didn't change the look of the parent nav items.. because I don't have any ideas of what to do there. I think they're ok with just the icon differentiating 🤷 

One last weird thing - on mobile, you click and expand, then navigate.. but to get the nav to un-expand again, you have to click on the page. I give up on that one.. it's better than it was..

joshxg:
ooo its feeling sooo much better!

PR #664, opened 2021-12-20
stick header row

### LINKED ISSUE #361: Stick headers
At the moment the `DataTable` headers aren't sticky and it would be better if they were.

The problem, from what I understand, is that you usually should be able to use `position: sticky` on the `<td/>` within a `<th>` and this should automatically make them sticky.

However, we use `display: flex` on the `th`, rather than the default, which means this doesn't work.

Not really sure how to make this work without spending a good chunk of time hacking away at some css 🙃

mark-prins:
Fixes #361 

turned out, wasn't too bad 😁 
had to add a background colour ( was transparent, and the body cells showed through! ) and a z-index ( the colour picker and checkbox cells were showing 🙄  )

### DIFF
--- a/packages/common/src/ui/layout/tables/DataTable.tsx
+++ b/packages/common/src/ui/layout/tables/DataTable.tsx
@@ -77,7 +77,17 @@ export const DataTable = <T extends DomainObject>({
       }}
     >
       <MuiTable sx={{ display: 'block', flex: 1 }}>
-        <TableHead sx={{ flex: 1, display: 'flex', flexDirection: 'column' }}>
+        <TableHead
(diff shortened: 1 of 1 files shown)


joshxg (approved):
wow!! You are the CSS king! The wasted time noooo! 😢 

I tried adding a box shadow to try and make it stick out more.. I don't think it looks good but the idea is there, I suppose 🙄  🤣 

image

mark-prins:
was playing around with a border or background.. it needed something, I think. have gone with a box shadow 😉

joshxg:
Nice its lookin great!

PR #665, opened 2021-12-20
Font fun

### LINKED ISSUE #492: Fix Fout
Sometimes, especially if I have put a throttle on speed, there is a flash of unstyled text which is ugly as!

did similar here, though maybe we can do better 🤷  https://github.com/openmsupply/patient-hub/pull/61/files

mark-prins:
Fixes #492 - kinda. Possibly makes it worse

Spent a while looking at options, since I've had the fout show for me too! Found it hard to replicate actually, so decided after much playing around, that it isn't a huge problem.. and also that it might be better to have the text show than to delay display of the site. We already have react suspense, and if you are using chrome or firefox, there is a 3s FOIT (flash of invisible text) time anyway before the default font is used and then updated to the correct font.

I did add a fallback of `Arial` rather than the serif default, this is much closer to `Inter` so the fout is less obvious, maybe?

And then made things a little worse, possibly, by adding in the font weights that we are using - I did wonder why the fonts weren't looking exactly the same as in zeplin. turns out we weren't loading all the weights.

Then went offroad a little and fixed the table headers, which were looking like this:

### Stocktake
Screen Shot 2021-12-21 at 11 08 10 AM

### Outbound shipment
Screen Shot 2021-12-21 at 11 07 25 AM

and now look better 😀 
### Stocktake
Screen Shot 2021-12-21 at 11 09 01 AM

### Outbound shipment
Screen Shot 2021-12-21 at 11 09 35 AM

### DIFF
--- a/packages/host/public/index.html
+++ b/packages/host/public/index.html
@@ -1,8 +1,10 @@
 <html>
   <head>
+    <link rel="preconnect" href="https://fonts.googleapis.com" />
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
     <link
+      href="https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600;700&display=swap"
       rel="stylesheet"
-      href="https://fonts.googleapis.com/css?family=Inter"
     />
     <script type="text/javascript" src="/config.js"></script>
   </head>
--- a/packages/host/public/index.html
+++ b/packages/host/public/index.html
@@ -1,11 +1,5 @@
 <html>
   <head>
-    <link rel="preconnect" href="https://fonts.googleapis.com" />
-    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
-    <link
-      href="https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600;700&display=swap"
-      rel="stylesheet"
-    />
     <script type="text/javascript" src="/config.js"></script>
   </head>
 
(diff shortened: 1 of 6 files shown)


packages/host/public/index.html
```diff
@@ -1,8 +1,10 @@
 <html>
   <head>
+    <link rel="preconnect" href="https://fonts.googleapis.com" />
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
     <link
+      href="https://fonts.googleapis.com/css2?family=Inter:wght@400;500;600;700&display=swap"
```
joshxg, line 6: Wouldn't these not work if you're on a lan?

We're using the `fontsource` package for `Inter` for self-hosting the font and we have all the different weights in node modules.. maybe they're just not getting loaded properly?

image

joshxg (approved):
It's looking much better - I am a little confused about fonts if I'm honest.. I thought just adding the fontsource and importing worked like https://fontsource.org/docs/getting-started

But I suppose not 🤷

packages/host/public/index.html
mark-prins: huh. I didn't realise we were using `fontsource` oops. removing the `link` from the html does screw up the fonts though, so if there is fontsource it isn't used for the web version..??

mark-prins:
ah! fountsource was only used for storybook, and the google fonts cdn was still used for the site.
I've changed the css references to use `fontsource` and also updated to use variable fonts ( soz IE 11 ). I'm told that variable fonts reduce the bundle size, so have swapped to those.
the change has now upped the bundle by 184kb .. but this is due to the hosting of the woff files, and it includes the various character sets which we don't use. and the files are browser-cacheable

PR #674, opened 2021-12-21
Add fix mock server types

### LINKED ISSUE #670: Mock - server tsconfig is wrong and has a bunch of errors

joshxg:
Fixes #670 

- Was mainly "Need all files under ./src"
- Would prefer to just import like `@openmsupply-client/common` but the tsconfig path and also the window object are givin me issues 🤔

### DIFF
--- a/packages/inventory/src/types.ts
+++ b/packages/inventory/src/types.ts
@@ -1,5 +1,5 @@
-import { Column } from './../../common/src/ui/layout/tables/columns/types';
 import {
+  Column,
   StocktakeLineNode,
   StocktakeNode,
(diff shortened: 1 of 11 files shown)


mark-prins (approved):
👍 

To get this to compile, I also had to rename

```/packages/mock-server/src/schema/stockLine.ts```
to 
```/packages/mock-server/src/schema/StockLine.ts```

joshxg:
Argh that's annoying. I also renamed it

image

But it seems the change isn't coming through..  🤔 .. will merge and 🤞

PR #678, opened 2021-12-21
#677 inbound detail view crash

### LINKED ISSUE #677: Inbound detail view broken!
I did an oopsie and the detail view is constantly crashing when you navigate to it!

joshxg:
Fixes #677 

- Early return caused the number of hooks rendering to vary. doh, no idea how I did this!
- Then I thought to check tests and figured I'd do a quick fix for the breadcrumbs
- Then I noticed that the stocktake page was also having an error .. so I offroaded that also.
- Then I noticed the inbound shipment footer wasn't showing when it wasn't on hold... so I offroaded that also.


mark-prins (approved):
👍

PR #681, opened 2021-12-21
Add updated requisition types

### LINKED ISSUE #672: Requisition tsconfig is wrong and has a bunch of errors
😁

joshxg:
Fixes #672 

- Maybe the nullable dates shouldn't be a thing. Looking to change this up and remove the reducer anyway, though, so.. 🤷

### DIFF
--- a/packages/common/src/additions.schema.graphql
+++ b/packages/common/src/additions.schema.graphql
@@ -314,7 +314,7 @@ union RequisitionsResponse = ConnectorError | RequisitionConnector
 
 type RequisitionConnector {
   totalCount: Int!
-  nodes: [RequisitionNode]!
+  nodes: [RequisitionNode!]!
 }
 
 input UpdateSupplierRequisitionInput {
--- a/packages/common/src/ui/components/buttons/ColorSelectButton.tsx
+++ b/packages/common/src/ui/components/buttons/ColorSelectButton.tsx
@@ -5,7 +5,7 @@ import { CircleIcon } from '@common/icons';
 import { useTranslation } from '@common/intl';
 
 interface ColorSelectButtonProps {
-  color: string;
+  color?: string | null;
   onChange: (color: Color) => void;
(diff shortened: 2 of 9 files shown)


packages/common/src/additions.schema.graphql
```diff
@@ -314,7 +314,7 @@ union RequisitionsResponse = ConnectorError | RequisitionConnector
 
 type RequisitionConnector {
   totalCount: Int!
-  nodes: [RequisitionNode]!
+  nodes: [RequisitionNode!]!
```
mark-prins, line 317: huh. so the `RequisitionNode` is required and the `nodes` prop is also required? is that what this means?

packages/common/src/ui/components/buttons/ColorSelectButton.tsx
```diff
@@ -44,7 +44,7 @@ export const ColorSelectButton: FC<ColorSelectButtonProps> = ({
         disabled={disabled}
         icon={
           <CircleIcon
-            htmlColor={color}
+            htmlColor={color ?? '##8f90a6'}
```
mark-prins, line 47: are two hashes required?

mark-prins (approved):
👍  that all looks good to me!

packages/common/src/additions.schema.graphql
joshxg: Yeah.. that's how the backend does it in the schema also, I did it incorrectly for this. The type being generated was an array of `RequisitionNode | null` (and removing the final ! meant nodes: Array<RequisitionNode | null> | null).. so to get Array<RequisitionNode> needs `!` in both places

packages/common/src/ui/components/buttons/ColorSelectButton.tsx
joshxg: 😁

PR #687, opened 2021-12-22
#668 outbound sort

### LINKED ISSUE #668: Outbound sort isn't working when not grouped
when the detail view is not grouped by item, some sort columns aren't correct:

Screen Shot 2021-12-21 at 12 33 36 PM

mark-prins:
Fixes #668 #469 

Added the sort function for columns and fixed up sorting for outbound detail.
Had to move the grouping up into the reducer, so that I could use the column values when sorting


packages/host/webpack.config.js
```diff
@@ -71,7 +71,7 @@ module.exports = env => {
         },
         {
           test: /\.(woff|woff2|ttf|eot)$/,
-          use: 'file-loader?name=fonts/[name].[ext]!static',
+          use: 'file-loader?name=fonts/[name].[ext]',
```
mark-prins, line 74: oops! fonts were not working on the demo server, turns out that windows won't serve files with `!` in the name 🙄

packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
```diff
@@ -112,9 +129,16 @@ export const DetailView: FC = () => {
         align: ColumnAlign.Right,
         format: ColumnFormat.Currency,
         accessor: ({ rowData }) =>
-          ((rowData.sellPricePerPack ?? 0) * (rowData.numberOfPacks ?? 0)) /
-          rowData.unitQuantity,
-        sortable: false,
+          Object.values(rowData.batches).reduce(
```
mark-prins, line 132: fixed some dodgy math

packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
```diff
@@ -123,8 +147,15 @@ export const DetailView: FC = () => {
         align: ColumnAlign.Right,
         format: ColumnFormat.Currency,
         accessor: ({ rowData }) =>
-          (rowData.sellPricePerPack ?? 0) * (rowData.numberOfPacks ?? 0),
-        sortable: false,
+          Object.values(rowData.batches).reduce(
```
mark-prins, line 150: I do wonder if the sort function should just return the accessor value 🤔

packages/invoices/src/OutboundShipment/DetailView/reducer.ts
```diff
@@ -112,11 +113,15 @@ export const OutboundAction = {
...
   }),
+
+  flattenRows: (): OutboundShipmentAction => ({ type: ActionType.Flatten }),
+  groupRows: (): OutboundShipmentAction => ({ type: ActionType.Group }),
 };
 
 export interface OutboundShipmentStateShape {
   draft: OutboundShipment;
   sortBy: SortBy<OutboundShipmentSummaryItem>;
+  isGrouped: boolean;
```
mark-prins, line 124: I did have an action to set this explicitly, but then only used it on load, so thought it wasn't worth it 🤷

joshxg (approved):
Nice! Thanks for that.. and sorry about the conflicts 😁

packages/common/src/hooks/useSortBy/useSortBy.ts
```diff
@@ -8,6 +8,7 @@ export interface SortRule<T> {
 
 export interface SortBy<T> extends SortRule<T> {
   direction: 'asc' | 'desc';
+  getSortValue?: (row: T) => string | number;
```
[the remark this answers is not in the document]
mark-prins: it's not ideal.. and is here so that I can easily re-sort the data when flipping between grouped/ungrouped. I suppose that the grouping shouldn't be modifying the order of the items and there shouldn't be a need to re-sort.. but the grouping was using the original un-sorted dataset. likewise.. didn't see any good alternatives 😬

mark-prins:
thanks! yeah - still thinking that using the accessor for sorting might be the best option. if the accessor is returning `[multiple]` and that's what is shown in the table, it makes sense to sort on that rather than any underlying data. One possible use would be where the display is tidied up, like a date formatter, where we'd want to sort on the underlying Date object or numerical value. but that's what the formatter is for, I guess! So.. 🤷

PR #690, opened 2021-12-22
Add style updates

### LINKED ISSUE #689: Inbound shipment column widths
The column widths for inbound shipment are a little short. Should also double check if they're the right columns to be using

joshxg:
Fixes #689 

- Ended up being a few different style updates
- Turned out the columns just weren't flexing because of the Grid added around the DataTable in the `GeneralTab`.. I think
- Then I made the scroll bar on the right 'auto' .. which when it wasn't showing when there were just a few items made the header row overlap the side panel .. so changed some zIndicies..

### DIFF
--- a/packages/common/src/ui/layout/tables/DataTable.tsx
+++ b/packages/common/src/ui/layout/tables/DataTable.tsx
@@ -78,7 +78,7 @@ export const DataTable = <T extends DomainObject>({
         overflowX: 'unset',
       }}
     >
-      <MuiTable sx={{ display: 'block', flex: 1 }}>
+      <MuiTable>
         <TableHead
           sx={{
             backgroundColor: dense ? 'transparent' : 'background.white',
--- a/packages/common/src/ui/layout/tables/DataTable.tsx
+++ b/packages/common/src/ui/layout/tables/DataTable.tsx
@@ -6,6 +6,8 @@ import {
   CircularProgress,
   TableBody,
   TableHead,
+  TableFooter,
+  TableRow,
   TableContainer,
   Table as MuiTable,
   Typography,
@@ -85,7 +87,7 @@ export const DataTable = <T extends DomainObject>({
             flexDirection: 'column',
             position: 'sticky',
             top: 0,
-            zIndex: 10,
+            zIndex: 'tableHeader',
             boxShadow: dense ? null : theme => theme.shadows[2],
           }}
         >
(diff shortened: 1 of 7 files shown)


packages/common/src/ui/layout/tables/DataTable.tsx
```diff
@@ -116,16 +118,29 @@ export const DataTable = <T extends DomainObject>({
...
+              <PaginationRow
+                page={pagination.page}
+                offset={pagination.offset}
+                first={pagination.first}
+                total={pagination.total ?? 0}
+                onChange={onChangePage}
+              />
+            )}
+          </TableRow>
+        </TableFooter>
```
joshxg, line 142: Ah, also the pagination row was dropping to the bottom so this is trying to keep it sticky at the bottom

packages/host/src/Host.tsx
```diff
@@ -223,7 +223,7 @@ const Host: FC = () => (
                           overflow="hidden"
                         >
                           <AppBar />
-                          <Box display="flex" flex={1} overflow="scroll">
+                          <Box display="flex" flex={1} overflow="auto">
```
joshxg, line 226: this is trying to make the scroll bar on the RHS go away when there's no scrolling to do

packages/invoices/src/InboundShipment/DetailView/GeneralTab.tsx
```diff
@@ -47,23 +47,24 @@ export const GeneralTab: FC<
...
+          <Switch
+            label={t('label.group-by-item', { ns: 'replenishment' })}
+            onChange={toggleIsGrouped}
+            checked={isGrouped}
+            size="small"
+            disabled={rows?.length === 0}
+            color="secondary"
+          />
+        </Grid>
+      )}
```
joshxg, line 67: this is trying to make the headers flex again. Of course, I will find out what it breaks and revert it 😢

joshxg (commented):
There is one more annoyance - when there is quite a lot of horizontal scroll, the header row background gets cut off.. I had a quick look and it seems that the actual table width is not as wide and the rows are overflowing.. or something. Not sure, but enough css for tonight..

mark-prins:
found another thing with the header.. on a small screen, if you have the side panel closed, then open it, then scroll the table horizontally, you can get this effect with the header:
Screen Shot 2021-12-23 at 9 34 49 AM

I've added e4f23211 which I think fixes without breaking other stuff?

mark-prins (approved):
👍  looks great!

joshxg:
Oh awesome thanks for that commit!

PR #695, opened 2021-12-22
#692 inbound optimistic onhold

### LINKED ISSUE #692: Inbound "on hold" should use an optimistic mutation
Currently using a plain mutation, the inbound on hold button should use an optimistic mutation

joshxg:
Fixes #692 

- Turns out the problem was actually because the mutation was being debounced.. which gets into a bit more of a tricky problem: The mutation is debounced, so the optimistic update doesn't trigger. You could maybe do some other tricky dicky things like debounce the actual query rather than the 'update' function, which might be better? But I think you might get into more complicated situations where the cache has been updated, then a timer for a debounce query starts, and then something else updates the cache but doesn't update and doesn't cancel the query being sent and then other overrides happen.. or something? I'm not sure if it's really a problem, but I thought this would work for now.. though, also need to use that buffer state for the status button.. lucky they're right next to each other, I guess!
- Could also do a manual cache update for this special case using `useQueryClient().setQueryData` 🤔  - though need to manage a rollback and things after that also..

Also getting closer to finding a pattern with the api stuff, I think ... 🙄 

something like...
- `useInbound` - a query for the inbound
- `useInboundUpdate` - a query for updating the inbound
- `useInboundFields` (uses `useInbound` and `useInboundUpdate`
- `useInboundLines`/`useInboundItems`/ `useInboundLineUpsert`/`Delete`/

It's a lot of hooks and doesn't organise the queries for them very nicely..

### DIFF
--- a/packages/invoices/src/InboundShipment/DetailView/Footer.tsx
+++ b/packages/invoices/src/InboundShipment/DetailView/Footer.tsx
@@ -23,13 +23,8 @@ import {
   useIsInboundEditable,
 } from './api';
 
-interface InboundDetailFooterProps {
-  draft: Invoice;
-  save: () => Promise<void>;
-}
-
-const createStatusLog = (draft: Invoice) => {
-  const statusIdx = inboundStatuses.findIndex(s => draft.status === s);
+const createStatusLog = (invoice: Invoice) => {
+  const statusIdx = inboundStatuses.findIndex(s => invoice.status === s);
   const statusLog: Record<InvoiceNodeStatus, null | string | undefined> = {
     [InvoiceNodeStatus.New]: null,
     [InvoiceNodeStatus.Picked]: null,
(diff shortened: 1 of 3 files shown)


packages/invoices/src/InboundShipment/DetailView/Footer.tsx
```diff
@@ -103,8 +107,15 @@ export const Footer: FC<InboundDetailFooterProps> = ({ draft, save }) => {
                     variant="contained"
                     color="secondary"
                     onClick={async () => {
-                      success('Saved invoice! 🥳 ')();
-                      save();
+                      update(
+                        {
+                          onHold: onHoldBuffer,
+                          status: InvoiceNodeStatus.Verified,
```
joshxg, line 113: Not ideal but thought to address this after talking about statuses a bit

packages/invoices/src/InboundShipment/DetailView/Footer.tsx
```diff
@@ -103,8 +107,15 @@ export const Footer: FC<InboundDetailFooterProps> = ({ draft, save }) => {
...
+                        },
+                        {
+                          onSuccess: success('Saved invoice! 🥳'),
```
joshxg, line 116: Not sure if a toast here or something else, so haven't translated. Was thinking to have a blocking spinner when saving?

packages/invoices/src/InboundShipment/DetailView/api.ts
```diff
@@ -327,7 +338,7 @@ export const useInboundFields = <KeyOfInvoice extends keyof Invoice>(
     }
   );
 
-  const update = useDebounceCallback(mutateAsync, [], timeout);
+  const update = useDebounceCallback(mutate, [], timeout);
```
joshxg, line 341: The `mutate` fn allows for passing extra options from the caller, which is nice to be able to add toasts etc

packages/invoices/src/InboundShipment/DetailView/Footer.tsx
mark-prins: a blocking spinner is probably best, otherwise the update might overwrite local changes if it is a slow update.
Wondered about just disabling the save button while updating and/or having a spinner icon on the button while updating, but think we need to block

mark-prins (approved):
👍  LGTM!

joshxg:
Yup. Possible to copy the response from the server over and then merge any data from the server but the easiest for now is to just block 🤷

PR #696, opened 2021-12-23
#691 add a bit of color to invoices

### LINKED ISSUE #691: Add color back to invoices
Add the color columns and inputs back to the invoices view

joshxg:
Fixes #691 

- Adds color back to the listviews and detail views of invoices.
- Offroad fixed a dom validation problem that was coming up with the pagination row

### DIFF
--- a/packages/common/src/ui/layout/tables/DataTable.tsx
+++ b/packages/common/src/ui/layout/tables/DataTable.tsx
@@ -6,8 +6,6 @@ import {
   CircularProgress,
   TableBody,
   TableHead,
-  TableFooter,
-  TableRow,
   TableContainer,
   Table as MuiTable,
   Typography,
@@ -118,29 +116,28 @@ export const DataTable = <T extends DomainObject>({
             );
           })}
         </TableBody>
-        <TableFooter
-          sx={{
-            flex: 1,
-            display: 'flex',
-            flexDirection: 'column',
-            position: 'sticky',
-            insetBlockEnd: 0,
-            backgroundColor: 'white',
-          }}
-        >
-          <TableRow>
-            {pagination && onChangePage && (
-              <PaginationRow
-                page={pagination.page}
-                offset={pagination.offset}
-                first={pagination.first}
-                total={pagination.total ?? 0}
-                onChange={onChangePage}
-              />
-            )}
-          </TableRow>
-        </TableFooter>
       </MuiTable>
+      <Box
+        sx={{
+          flex: 1,
+          display: 'flex',
+          flexDirection: 'column',
+          position: 'sticky',
+          insetBlockEnd: 0,
+          backgroundColor: 'white',
+          justifyContent: 'flex-end',
+        }}
+      >
+        {pagination && onChangePage && (
+          <PaginationRow
+            page={pagination.page}
+            offset={pagination.offset}
+            first={pagination.first}
+            total={pagination.total ?? 0}
+            onChange={onChangePage}
+          />
+        )}
+      </Box>
     </TableContainer>
   );
 };
--- a/packages/common/src/ui/layout/tables/DataTable.tsx
+++ b/packages/common/src/ui/layout/tables/DataTable.tsx
@@ -119,7 +119,7 @@ export const DataTable = <T extends DomainObject>({
       </MuiTable>
       <Box
         sx={{
-          flex: 1,
+          flex: 0,
           display: 'flex',
           flexDirection: 'column',
           position: 'sticky',
--- a/packages/invoices/src/InboundShipment/ListView/ListView.tsx
+++ b/packages/invoices/src/InboundShipment/ListView/ListView.tsx
@@ -4,10 +4,9 @@ import {
   DataTable,
   useColumns,
   useListData,
-  // getNameAndColorColumn,
+  getNameAndColorColumn,
   TableProvider,
   createTableStore,
-  // Color,
   InvoiceNodeStatus,
   useOmSupplyApi,
   useNotification,
@@ -31,7 +30,7 @@ export const InboundListView: FC = () => {
     data,
     isLoading,
     onDelete,
-    // onUpdate,
+    onUpdate,
     sortBy,
     onChangeSortBy,
     onCreate,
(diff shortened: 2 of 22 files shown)


packages/invoices/src/InboundShipment/ListView/ListView.tsx
```diff
@@ -51,10 +50,7 @@ export const InboundListView: FC = () => {
 
   const columns = useColumns<InvoiceRow>(
     [
-      // getNameAndColorColumn((row: InvoiceRow, color: Color) => {
-      //   onUpdate({ ...row, color: color.hex });
-      // }),
-      'otherPartyName',
+      [getNameAndColorColumn(), { setter: onUpdate }],
```
mark-prins, line 53: the colour update in the `ListView` is throwing an error, as its updating the status field too. The version in the detail view / SitePanel is fine; that one is using `useInboundFields` which isn't sending status.

packages/requisitions/src/CustomerRequisition/ListView/ListView.tsx
```diff
@@ -47,7 +47,7 @@ export const CustomerRequisitionListView: FC = () => {
 
   const columns = useColumns<RequisitionRow>(
     [
-      { ...getNameAndColorColumn(), setter: onUpdate },
+      [getNameAndColorColumn(), { setter: onUpdate }],
```
mark-prins, line 50: currently the requisitions ( customer & supplier ) aren't saving the change when updated using the sidepanel, though the comment is.

packages/requisitions/src/SupplierRequisition/ListView/ListView.tsx
```diff
@@ -48,7 +48,7 @@ export const SupplierRequisitionListView: FC = () => {
 
   const columns = useColumns<RequisitionRow>(
     [
-      { ...getNameAndColorColumn(), setter: onUpdate },
+      [getNameAndColorColumn(), { setter: onUpdate }],
```
mark-prins, line 51: just an observation.. it is possible to update the colour of finalised invoices from the ListView - though you cannot make changes in the DetailView. I think for colour it's probably fine to update finalised invoices, but then the DetailView functionality is inconsistent

mark-prins (approved):
🥳  nice to have the colour back!

a couple of things, as noted. main issue is the inbound shipment mutation on the ListView

packages/common/src/ui/layout/tables/DataTable.tsx
```diff
@@ -118,29 +116,28 @@ export const DataTable = <T extends DomainObject>({
...
-                total={pagination.total ?? 0}
-                onChange={onChangePage}
-              />
-            )}
-          </TableRow>
-        </TableFooter>
       </MuiTable>
+      <Box
+        sx={{
+          flex: 1,
```
mark-prins, line 122:
```suggestion
          flex: 0,
```

the `flex-grow` here is causing this:

table oddity

The footer is taking up half the space and is hiding some rows in the table. I would change to `flex: 0` or explicitly set `flex-grow: 0`. this does mean that when you have a small table, the pagination is at the bottom of the table, which I think you were trying to avoid? I actually like the pagination to be close to the table, so it suits me 😄  the pagination is still sticky, just not at the bottom of the page 🤷‍♂️

packages/invoices/src/InboundShipment/ListView/ListView.tsx
joshxg: Oops! I've changed this mutation to only update colors

packages/requisitions/src/CustomerRequisition/ListView/ListView.tsx
joshxg: ugh! Lucky you're here

packages/requisitions/src/SupplierRequisition/ListView/ListView.tsx
joshxg: Yup - similar on the detail views with being able to select rows when finalised..

I'd like your opinion before doing anything about it. 

- If we just disable the row, what do we do with the 'color button' - hard to make it look disabled?
- Make the 'enabled' button look different so that it's easier to make the 'disabled' button look disabled?
- Snack/toast saying 'can't change colour of this row'?

🤔

packages/common/src/ui/layout/tables/DataTable.tsx
joshxg: I've changed to flex: 0.

Now another problem which i'm not sure about.. when on a smaller screen that needs horizontal scroll, the pagination row stretches the full width 😢

packages/requisitions/src/SupplierRequisition/ListView/ListView.tsx
mark-prins: I wasn't sure how to make the colour button look disabled, or how to make it look more enabled! at least, not without making it look clunky. Agree with the underlying concern though, that we need to make the distinction obvious somehow. it would be annoying to have the colour button look the same, but work sometimes and not work for other rows.

Slightly dodgy, but It would be easier to enable the colour change on the detail view. It's worth asking if we should just enable colour change in both. there might be a reason you want to mark specific invoices.. even finalised ones? maybe..?

If that's not ok, I like the toast option - allow people to click the colour button and show a message. I'm usually against that approach; I think if the system knows that I can't perform an action, why does it allow me to do it, and then pop up a dialog telling me? Just disable the action 🤷  The case for using that approach is to give some information, saying why I shouldn't have clicked that button. for the colour button.. the case for doing it, is that it's hard to style the colour button 😁  and maybe also informing the user

packages/common/src/ui/layout/tables/DataTable.tsx
mark-prins: oh! balls.

that's the sort of thing I had the grid in place for, to get the 'group by' switch to be visible even when there's horizontal scroll. there are just so many combinations of things to test with this thing!!

joshxg: Yeah! Ugh!

Side note - I think the group by item toggle 'dissapearing' when you scroll is actually quite neat and wondered if all the filters could go in a row like that 😁

PR #697, opened 2021-12-23
#688 Handle multiple notes

### LINKED ISSUE #688: Note column for grouped rows
The `note` popover column for the grouped rows should show all of the notes for each grouped line, rather than just a single note.

joshxg:
Fixes #688 

- Adds handling of multiple notes
image
- Done it quite differently, but I think it works OK and allows for a reasonably diverse range of uses.
- Also the hovering was doing that weird dissapear and reappear thing so have tried to adjust the virtual ref slightly

### DIFF
--- a/packages/common/src/ui/components/popover/BasePopover/usePopover.tsx
+++ b/packages/common/src/ui/components/popover/BasePopover/usePopover.tsx
@@ -50,8 +50,8 @@ export const usePopover = (): UsePopoverControl => {
         left: e.clientX,
         bottom: e.clientY,
         right: e.clientX,
-        width: 0,
-        height: 0,
+        width: 25,
+        height: 25,
       } as DOMRect);
 
     setAnchorElCallback.current?.({ getBoundingClientRect });
--- a/packages/common/src/ui/layout/tables/columns/NotePopoverColumn/NotePopoverColumn.tsx
+++ b/packages/common/src/ui/layout/tables/columns/NotePopoverColumn/NotePopoverColumn.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { FC } from 'react';
 import { DomainObject } from '@common/types';
 import { ColumnAlign, ColumnDefinition } from '../types';
 import { MessageSquareIcon } from '@common/icons';
@@ -6,14 +6,28 @@ import { PaperPopover, PaperPopoverSection } from '@common/components';
 import { useTranslation } from '@common/intl';
 import { isProduction } from '@common/utils';
 
-interface ObjectWithNote {
-  note: string;
+interface NoteObject {
+  header: string;
+  body: string;
 }
 
-const isObjectWithANote = (
+const NoteSection: FC<{ header: string; body: string }> = ({
+  header,
+  body,
+}) => {
+  return (
+    <>
+      <b>{header}</b>
+      <span>{body}</span>
+    </>
+  );
+};
+
+const hasRequiredFields = (
   variableToCheck: unknown
-): variableToCheck is ObjectWithNote =>
-  (variableToCheck as ObjectWithNote).note !== undefined;
+): variableToCheck is NoteObject =>
+  (variableToCheck as NoteObject).header !== undefined &&
+  (variableToCheck as NoteObject).body !== undefined;
 
 export const getNotePopoverColumn = <T extends DomainObject>(
   label?: string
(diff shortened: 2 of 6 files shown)


packages/common/src/ui/components/popover/BasePopover/usePopover.tsx
```diff
@@ -50,8 +50,8 @@ export const usePopover = (): UsePopoverControl => {
         left: e.clientX,
         bottom: e.clientY,
         right: e.clientX,
-        width: 0,
-        height: 0,
+        width: 25,
+        height: 25,
```
joshxg, line 54: this is making the virtual ref has some width/height which seems to help. Alternative I tried playing with was to try offset the left/bottom/right/top a little based on the position of the popover.. this worked a little better

packages/common/src/ui/layout/tables/columns/NotePopoverColumn/NotePopoverColumn.tsx
```diff
@@ -1,19 +1,33 @@
...
-const isObjectWithANote = (
+const NoteSection: FC<{ header: string; body: string }> = ({
+  header,
+  body,
+}) => {
+  return (
+    <>
+      <b>{header}</b>
+      <span>{body}</span>
+    </>
```
joshxg, line 22: uhh.. not the best styling.. but seemed to look OK like this... 🤷

packages/common/src/ui/layout/tables/columns/NotePopoverColumn/NotePopoverColumn.tsx
```diff
@@ -23,19 +37,23 @@ export const getNotePopoverColumn = <T extends DomainObject>(
...
+
+        This column requires the fields 'header' and 'body' to be present in the row data to render
+        correctly.
+
+        Have you forgotten to provide a custom accessor to return the right value? i.e.
+        [ getNotePopoverColumn(), { accessor: ({rowData}) => ({header: rowData.batch, body: rowData.note}) }]
         `);
       } else {
-        return '';
+        return { header: '', body: '' };
```
joshxg, line 56: changed this to require the row data to have header and body fields rather than a `note` field.. easy enough to do a quick map in a custom accessor 🤷

packages/mock-server/src/data/data.ts
```diff
@@ -453,10 +453,11 @@ const createStockLines = (items: Item[]) => {
...
             onHold: faker.datatype.number(10) < 2,
-            note:
-              faker.datatype.number(10) < 4
-                ? faker.commerce.productDescription()
-                : null,
+            note: faker.commerce.productDescription(),
+            // note:
+            //   faker.datatype.number(10) < 4
+            //     ? faker.commerce.productDescription()
+            //     : null,
```
joshxg, line 460: just changed this to make it easier to check the notes.. will likely get sick of it and change it back?

packages/common/src/ui/layout/tables/columns/NotePopoverColumn/NotePopoverColumn.tsx
```diff
@@ -47,13 +65,19 @@ export const getNotePopoverColumn = <T extends DomainObject>(
...
+    }
+
     return value ? (
       <PaperPopover
         width={400}
-        height={180}
         Content={
           <PaperPopoverSection label={label ?? t('label.notes')}>
-            {String(value)}
+            {content.map(({ header, body }) => (
```
mark-prins, line 78: could you filter out blank bodies please? otherwise you can get something like this:

Screen Shot 2021-12-23 at 4 03 31 PM

packages/mock-server/src/data/data.ts
mark-prins: I've also added a few notes to the demo server database to help with testing

mark-prins (approved):
👍  very nice - that looks good!

packages/common/src/ui/layout/tables/columns/NotePopoverColumn/NotePopoverColumn.tsx
joshxg: Ah, thanks for noticing this.

I've filtered out the sections with no note, but left the ones which don't have a batch. Not ideal, but not kind of need to differentiate and if theres no batch code, not sure what to use 🙄

PR #699, opened 2021-12-23
animate group transition

### LINKED ISSUE #693: Animate the switch of grouping by item
Thought it would be good to add an animation when switching to group by items.

Either just wrapping the whole table in a `Slide` or maybe using a transition group https://mui.com/components/transitions/#transitiongroup ?

mark-prins:
Fixes #693 

Finding the transition a little strange, but it's something! better than what it was 🤷‍♂️ 
See what you think anyway. I did try some of the other options - `Collapse` was the best option. Didn't try sliding in the whole table though, so that's still an option. I just liked the rows animating and using the TransitionGroup is easy

### DIFF
--- a/packages/common/src/ui/layout/tables/DataTable.tsx
+++ b/packages/common/src/ui/layout/tables/DataTable.tsx
@@ -4,6 +4,7 @@ import React, { useEffect } from 'react';
 import {
   Box,
   CircularProgress,
+  Collapse,
   TableBody,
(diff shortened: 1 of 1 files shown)


joshxg (approved):
on a roll, this is awesome! I tried slide out a little bit.. annoying that it needs a ref and I think collapse is better 😁 

Minor change but has a good amount of impact, I reckon!

PR #700, opened 2021-12-23
move docs link

### LINKED ISSUE #680: All about that Docs button!
Been wondering about the docs button.

- Does it link to the internet hosted docs? Or a version hosted by the server you're on? 🤔 
- Should probably like to the relevant page?
- Does it have to be so big and dominant?

Prompted to make this issue also because I had an idea:

image

- Replace the 'back arrow' which 1) isn't useful 2) has really weird back behaviour 3) maybe wasn't intended as a button but it looks (to me!) to be one with an `info` button for the current page. There's still issues with it, but when isn't there? :)

mark-prins:
Fixes #680 

moved docs to the bottom section of the app drawer, touching as many files as possible for such a simple change 🤷

### DIFF
--- a/packages/common/src/ui/components/navigation/ExternalNavLink.tsx
+++ b/packages/common/src/ui/components/navigation/ExternalNavLink.tsx
@@ -1,0 +1,83 @@
+import React, { FC } from 'react';
+import {
+  ListItem,
+  ListItemIcon,
+  ListItemText,
+  Tooltip,
+  ListItemButton,
+  Box,
+  ListItemProps,
+} from '@mui/material';
+import { styled } from '@mui/material/styles';
+
+const getListItemCommonStyles = () => ({
+  height: 40,
+  borderRadius: 20,
+  justifyContent: 'center',
+  alignItems: 'center',
+});
+
+const StyledListItem = styled<FC<ListItemProps & { to: string }>>(ListItem)(
+  ({ theme }) => ({
+    ...getListItemCommonStyles(),
+    backgroundColor: 'transparent',
+    boxShadow: 'none',
+    marginTop: 5,
+    '&:hover': {
+      boxShadow: theme.shadows[8],
+    },
+  })
+);
+
+export interface ExternalNavLinkProps {
+  icon?: JSX.Element;
+  text?: string;
+  to: string;
+}
+
+export const ExternalNavLink: FC<ExternalNavLinkProps> = props => {
+  const { icon = <span style={{ width: 2 }} />, text, to } = props;
+
+  const CustomLink = React.useMemo(
+    () =>
+      React.forwardRef<HTMLAnchorElement>((linkProps, ref) => (
+        <a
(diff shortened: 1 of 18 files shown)


packages/common/src/ui/components/navigation/ExternalNavLink.tsx
```diff
@@ -0,0 +1,83 @@
+import React, { FC } from 'react';
```
mark-prins, line 1: all this to have a nav link that looks the same as the others, but doesn't link to a relative path, or both with selected state

packages/host/src/components/AppBar/SectionIcon.tsx
```diff
@@ -0,0 +1,67 @@
...
+  return undefined;
+};
+
+export const SectionIcon: React.FC = () => {
+  const t = useTranslation('app');
+  const section = getSection();
+  const icon = getIcon(section);
+
+  return icon ? (
+    <Tooltip title={t(section as LocaleKey)}>
```
mark-prins, line 46: iffy, and probably not required? not sure that we need a tooltip on the icon.. generally think that it is helpful tho

packages/host/src/components/AppBar/SectionIcon.tsx
```diff
@@ -0,0 +1,75 @@
...
+    const match = matchPath(
+      RouteBuilder.create(route ?? '')
+        .addWildCard()
+        .build(),
+      location.pathname
+    );
+    if (!!match)
+      return {
+        icon: getIcon(route),
+        titleKey: route as LocaleKey,
```
mark-prins, line 60: iffy, and possibly not needed - having a tooltip on the icon. I think it's helpful to explain them though, so put the tooltip on.

packages/host/src/components/AppDrawer/AppDrawer.tsx
```diff
@@ -207,6 +209,11 @@ export const AppDrawer: React.FC = () => {
             icon={<RadioIcon fontSize="small" color="primary" />}
             text={t('sync')}
           />
+          <ExternalNavLink
+            to={ExternalURL.PublicDocs}
+            icon={<BookIcon fontSize="small" color="primary" />}
```
mark-prins, line 214: thinking that if/when we do have specific docs links for specific pages then we can manage that in a number of different ways. That was part of the reason for having the docs link in the AppBarButtons - but I like the button better tucked away in the left nav

joshxg: Yup!

joshxg (approved):
This is awesome! Thanks. Think it looks good and nice to have everything consolidated in a single place.

packages/common/src/ui/components/navigation/ExternalNavLink.tsx
mark-prins: good point! thanks

PR #701, opened 2021-12-23
Add simple locations hook

### LINKED ISSUE #510: Location queries
- The current development branch on the backend now has support for querying for locations
- Note: No support for mutations yet
- Would like to setup some queries for locations
- I would really like to setup something similar to `system/item` and `system/names`, where we can export a `useLocations` hook as well as a `LocationSearchInput` or something component. But - locations aren't system data. So, where can we put it?

I wonder if we should put off playing with locations until there are mutations. They won't sync as they're store data, so we will need to manually create the locations in the database which is a bit annoying but not the end of the world(since there also isn't any migration when the database schema changes we need to redo it 🤷 )  . We forgot to explicitly ask to be able to create or update locations, only to read them .. sigh!

joshxg:
Fixes #510 

- I want to start correcting/completing the inbound cells in the edit modal, so wanted locations.
- Just the first query to get it on the board with all the boilerplate.. the hook I used in just the list view of the inbound shipment page and it's returning from the mock server and against the demo server

### DIFF
--- a/packages/common/src/operations.graphql
+++ b/packages/common/src/operations.graphql
@@ -1175,3 +1175,41 @@ mutation insertInboundShipment($id: String!, $otherPartyId: String!) {
     }
   }
 }
+
+query locations {
+  locations {
+    __typename
+    ... on LocationConnector {
+      __typename
+      nodes {
+        __typename
+        id
+        name
+        onHold
+        code
+      }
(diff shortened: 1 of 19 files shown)


mark-prins (approved):
LGTM! 😄

PR #702, opened 2021-12-28
# add simple column resize implementation

joshxg:
Random PR - I just had a thought and wanted to give it a go. Certainly not a full implementation, but wanted to get your thoughts on whether it would be worthwhile to do so or wait until a later phase.. I'm guessing a later phase.

### DIFF
--- a/packages/common/src/ui/layout/tables/components/Header/Header.tsx
+++ b/packages/common/src/ui/layout/tables/components/Header/Header.tsx
@@ -1,7 +1,7 @@
 import React, { FC } from 'react';
-import { TableCell, TableRow, TableSortLabel } from '@mui/material';
+import { Box, TableCell, TableRow, TableSortLabel } from '@mui/material';
 import { Column } from '../../columns/types';
-import { SortDescIcon } from '@common/icons';
+import { MenuDotsIcon, SortDescIcon } from '@common/icons';
 import { DomainObject } from '@common/types';
 import { useDebounceCallback } from '@common/hooks';
 
@@ -37,6 +37,8 @@ export const HeaderCell = <T extends DomainObject>({
     Header,
   } = column;
 
+  const [x, setX] = React.useState(0);
+
   const { direction, key: currentSortKey } = sortBy ?? {};
 
   const isSorted = key === currentSortKey;
@@ -58,8 +60,8 @@ export const HeaderCell = <T extends DomainObject>({
         borderBottom: '0px',
         paddingLeft: '16px',
         paddingRight: '16px',
-        width,
-        minWidth,
+        width: width + 10,
+        minWidth: minWidth + 10,
         flex: `${width} 0 auto`,
         fontWeight: 'bold',
         fontSize: dense ? '12px' : '14px',
@@ -68,17 +70,49 @@ export const HeaderCell = <T extends DomainObject>({
       sortDirection={isSorted ? direction : false}
     >
       {sortable ? (
-        <TableSortLabel
-          hideSortIcon={false}
-          active={isSorted}
-          direction={direction}
-          IconComponent={SortDescIcon}
+        <Box
+          flexDirection="row"
+          display="flex"
+          justifyContent="space-between"
+          alignItems="center"
         >
-          <Header column={column} />
-        </TableSortLabel>
+          <TableSortLabel
+            hideSortIcon={false}
+            active={isSorted}
+            direction={direction}
+            IconComponent={SortDescIcon}
+          >
+            <Header column={column} />
+          </TableSortLabel>
+
+          {column?.onChangeWidth && (
+            <Box
+              sx={{
+                cursor: 'col-resize',
+                backgroundColor: 'transparent',
+              }}
+              draggable
+              onDragStart={e => {
+                setX(e.clientX);
+              }}
+              onDragEnd={e => {
+                column?.onChangeWidth &&
+                  column.onChangeWidth(
+                    Math.max(column.width + (e.clientX - x), column.minWidth)
+                  );
+
+                setX(e.clientX);
+              }}
+            >
+              <MenuDotsIcon />
+            </Box>
+          )}
+        </Box>
       ) : (
         <Header column={column} />
       )}
     </TableCell>
   );
 };
+
+<div />;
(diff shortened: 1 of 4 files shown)


packages/common/src/ui/layout/tables/components/Header/Header.tsx
```diff
@@ -58,8 +60,8 @@ export const HeaderCell = <T extends DomainObject>({
         borderBottom: '0px',
         paddingLeft: '16px',
         paddingRight: '16px',
-        width,
-        minWidth,
+        width: width + 10,
+        minWidth: minWidth + 10,
```
joshxg, line 64: Just put a placeholder icon for the button to resize right now and it needed some more room 🤷

packages/common/src/ui/layout/tables/components/Header/Header.tsx
```diff
@@ -68,17 +70,49 @@ export const HeaderCell = <T extends DomainObject>({
...
+              onDragStart={e => {
+                setX(e.clientX);
+              }}
+              onDragEnd={e => {
+                column?.onChangeWidth &&
+                  column.onChangeWidth(
+                    Math.max(column.width + (e.clientX - x), column.minWidth)
+                  );
+
+                setX(e.clientX);
```
joshxg, line 104: Just went with a drag-then-resize rather then resizing in real time - easier.. though having it realtime is probably nicer

packages/invoices/src/InboundShipment/DetailView/columns.ts
```diff
@@ -12,12 +13,25 @@ import { InvoiceLine, InboundShipmentItem } from './../../types';
 
 export const useInboundShipmentColumns = (): Column<
   InvoiceLine | InboundShipmentItem
->[] =>
-  useColumns<InvoiceLine | InboundShipmentItem>(
+>[] => {
+  const [widths, setWidths] = useState<number[]>([
+    100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100,
+  ]);
```
joshxg, line 19: would see these widths coming from local storage and persisting when changed

packages/common/src/ui/layout/tables/components/Header/Header.tsx
mark-prins: the current icon is, well, a little too bold for my tastes 😄

packages/common/src/ui/layout/tables/components/Header/Header.tsx
```diff
@@ -68,17 +70,49 @@ export const HeaderCell = <T extends DomainObject>({
...
+            direction={direction}
+            IconComponent={SortDescIcon}
+          >
+            <Header column={column} />
+          </TableSortLabel>
+
+          {column?.onChangeWidth && (
+            <Box
+              sx={{
+                cursor: 'col-resize',
```
mark-prins, line 91: very nice!

packages/common/src/ui/layout/tables/components/Header/Header.tsx
mark-prins: the icon jumping back into place, and then the column resizing is a bit weird at first. ideally, on press a vertical line would show over the top of the table which gets dragged with the icon. I don't know if this is possible though

mark-prins (approved):
😄  that is fantastic!!
I'm usually using on my laptop screen and the small column widths have been bugging me. I've just spent an inappropriate amount of time resizing columns like a happy little pixie.

All totally OTT and not asked for and we should definitely include it now. If the column widths persisted in local storage then that's all I could ask for. oh, and a nicer icon!

packages/common/src/ui/layout/tables/components/Header/Header.tsx
joshxg: Neither. Just a place holder to gauge reactions of spending a bit of time to do this!

packages/common/src/ui/layout/tables/components/Header/Header.tsx
joshxg: Yeah it is.

- I think real-time column movements are just going to be too laggy. Though, maybe not and I'm overreacting - *especially* since we don't use editable columns in the larger list 🤔  .. basically my initial thought was that the types of cells we're going to have - different inputs, drop downs, dates etc will make it a little render heavy. But where those cells are used there aren't many rows 🤔 
- I went to google sheets and this 'style' of drag then resize is used - but you're right in that it uses an indicator.
- I *think* this should be significantly snappier on a built version so the lag between drag and resize should be a lot smaller
- It's probably worth giving the real time resize a proper goon a built version. Could also likely do it with refs and/or pure css though I suspect that might take a little longer
- I've done a quick attempt at the indicator using a react-table code sandbox as an example: https://codesandbox.io/s/empty-silence-rkukq?file=/src/App.js which seems to be OK. Only thing left to do would be getting the height of the actual table 🤔  -  using:
```
        &:active {
          &:before {
            content: "";
            position: absolute;
            background-color: red;
            z-index: 1000;
            width: 1px;
            min-height: 100vh;
          }
        }
```
- I've mainly been basing my assumptions of the column resize based on the react-table example - already with 20 rows and very basic data the table seems to be a bit janky in the code sandbox - though, a production version would likely feel fine. But increasing to 100 lines and it feels really bad and since this table seems to be relatively simple in term of what it's rendering I assumed real-time resize would be out of the question. But again, I think it's probably worth a shot?

joshxg:
> All totally OTT and not asked for and we should definitely include it now

Is this meant to be "not include it now"? 🤣 

Will keep this on the back burner for now, I think - might think about how a realtime version could be done and also a way to determine the height of the resize-indicator.

Oh and a better icon 😁

joshxg:
But yes, the motivation is primarily so we don't have to make sure columns are the 100% pixel perfect correct width on all dimensions - so annoying!

mark-prins:
ok - after playing with google sheets I think that real-time updating is overrated. The sheets version is nice to use - just having the vertical indicator is enough! probably helps that it is very snappy and you don't see the dragged icon snapping back into place before the resize - that's not a big issue for me though. would rather have the resize functionality in, than wait for perfection.

one other thought was adding a tooltip or title to the drag icon, 'resize columns' or something

mark-prins:
no, wait, what? I liked the resize.. is it gone now?

joshxg:
Oh no, I just got annoyed at the PR being open.. 😁  i've not gotten round to fixing it up, but hope to!

PR #704, opened 2021-12-28
#703 locations to inbound

### LINKED ISSUE #703: Add location selection to inbound shipments
Need to be able to assign a batch/line to a location in inbound shipments

joshxg:
Doesn't fix #703 🙃 

Some misc. fixes and setup for the real PR....! Also just getting my feet wet again after taking basically a year off!

- Fixed a bug which was making now batches show up for an edit
- Moved the 'static' columns into their own 'panel' component similar to how the stocktake modal is done - but I simplified it further by just using a second `DataTable`
- Fixed some big re-rendering issues where the currency input was always triggering 2 re-renders when first rendered. The table always still rerenders when switched to (i.e. the pricing table) but is only doing it once now 🎉

### DIFF
--- a/packages/common/src/ui/components/inputs/CurrencyInput/CurrencyInput.tsx
+++ b/packages/common/src/ui/components/inputs/CurrencyInput/CurrencyInput.tsx
@@ -58,13 +58,16 @@ export const CurrencyInput: FC<CurrencyInputProps> = ({
   // loses focus when the buffer is in this invalid state, the currency input will
   // append two zeros so will always result in a valid number.
   useEffect(() => {
+    // circuit break when the value is equal to the buffer - we don't need to do anything.
+    if (Number(buffer) === value) return;
+
     if (buffer == null) {
       return onChangeNumber(0);
     }
     if (!Number.isNaN(buffer)) {
       return onChangeNumber(Number(buffer));
     }
-  }, [buffer]);
+  }, [buffer, value]);
 
   return (
     <StyledCurrencyInput
--- a/packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
+++ b/packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
@@ -20,7 +20,7 @@ export enum ModalMode {
 }
 
 export const toItem = (line: InboundShipmentItem | InvoiceLine): Item => ({
-  id: 'lines' in line ? line.lines[0].id : line.itemId,
+  id: 'lines' in line ? line.lines[0].itemId : line.itemId,
   name: 'lines' in line ? line.lines[0].itemName : line.itemName,
   code: 'lines' in line ? line.lines[0].itemCode : line.itemCode,
   isVisible: true,
--- a/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEdit.tsx
+++ b/packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEdit.tsx
@@ -100,17 +100,18 @@ const useDraftInboundLines = (itemId: string) => {
     setDraftLines([...draftLines, newLine]);
   };
 
-  const updateDraftLine = (
-    patch: Partial<DraftInboundLine> & { id: string }
-  ) => {
-    const batch = draftLines.find(line => line.id === patch.id);
-    if (batch) {
-      const newBatch = { ...batch, ...patch, isUpdated: true };
-      const index = draftLines.indexOf(batch);
-      draftLines[index] = newBatch;
-      setDraftLines([...draftLines]);
-    }
-  };
+  const updateDraftLine = React.useCallback(
+    (patch: Partial<DraftInboundLine> & { id: string }) => {
+      const batch = draftLines.find(line => line.id === patch.id);
+      if (batch) {
+        const newBatch = { ...batch, ...patch, isUpdated: true };
+        const index = draftLines.indexOf(batch);
+        draftLines[index] = newBatch;
+        setDraftLines([...draftLines]);
+      }
+    },
+    [draftLines, setDraftLines]
+  );
 
   const saveLines = async () => {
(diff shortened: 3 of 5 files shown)


packages/common/src/ui/components/inputs/CurrencyInput/CurrencyInput.tsx
```diff
@@ -58,13 +58,16 @@ export const CurrencyInput: FC<CurrencyInputProps> = ({
   // loses focus when the buffer is in this invalid state, the currency input will
   // append two zeros so will always result in a valid number.
   useEffect(() => {
+    // circuit break when the value is equal to the buffer - we don't need to do anything.
+    if (Number(buffer) === value) return;
```
joshxg, line 62: was triggering the callback on mounting

packages/invoices/src/InboundShipment/DetailView/DetailView.tsx
```diff
@@ -20,7 +20,7 @@ export enum ModalMode {
 }
 
 export const toItem = (line: InboundShipmentItem | InvoiceLine): Item => ({
-  id: 'lines' in line ? line.lines[0].id : line.itemId,
+  id: 'lines' in line ? line.lines[0].itemId : line.itemId,
```
joshxg, line 23: oops, was using the invoice line ID as the item ID to lookup all of the invoice lines with the same item ID

packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEdit.tsx
```diff
@@ -106,17 +100,18 @@ const useDraftInboundLines = (itemId: string) => {
...
+      const batch = draftLines.find(line => line.id === patch.id);
+      if (batch) {
+        const newBatch = { ...batch, ...patch, isUpdated: true };
+        const index = draftLines.indexOf(batch);
+        draftLines[index] = newBatch;
+        setDraftLines([...draftLines]);
+      }
+    },
+    [draftLines, setDraftLines]
+  );
```
joshxg, line 114: TBH I'm flip flopping on `useCallback`/`useMemo`. When they first came out I thought 'why not just memoize all the things" and then I caused a bunch of bugs for myself and it made the code harder to read.

Now i'm flopping back to 'why not memoize all the things!' ... I have no middle ground?

packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEditPanel.tsx
```diff
@@ -0,0 +1,61 @@
...
+    <StyledTabPanel value={value}>
+      <StyledTabContainer>
+        <StyledStaticArea>
+          <DataTable dense columns={columns} data={lines} />
+        </StyledStaticArea>
+        {children}
+      </StyledTabContainer>
+    </StyledTabPanel>
+  );
+};
```
joshxg, line 61: Technically the table is a 'different table' for each tab, but since the content would be the same, there's no visible change. Possible performance hit by not having a single table and the 'tab' content change.. but this seems .. easier 🤔

packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEdit.tsx
mark-prins: heh. I have noticed a few `useCallback`s appearing 😃 
Would be good to know what the gotchas are if we are heading back into memoland.

packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEdit.tsx
```diff
@@ -235,29 +230,27 @@ export const InboundLineEdit: FC<InboundLineEditProps> = ({
...
-                    height: isMediumScreen ? 300 : 400,
-                    backgroundColor: theme =>
-                      alpha(theme.palette['background']['menu'], 0.4),
-                    position: 'absolute',
-                    borderRadius: '20px',
-                  }}
-                />
-                <StyledTabPanel value={Tabs.Batch}>
-                  <BatchTable
+                <InboundLineEditPanel
```
mark-prins, line 233: I wonder if a Slide animation would work for the tab switching - to reinforce the idea that the first couple of columns are static and the other columns are swapped in and out?

mark-prins (approved):
👍  that's great - like the tidy up

packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEdit.tsx
joshxg: Yes I wondered the same and added it - but then removed it 😁 

I kind of just felt like it was a little too distracting and busy and felt like when I was switching tab the time for the animation was annoying. Maybe I'll add it back and throw it up as a draft to see what you think.

One thing that is super cool is the animation on add batch, though!!

joshxg:
TBH the recent `useCallback` / `useMemo`'s I've been adding recently have had an explicit render target they're preventing.. mainly the flip flop is "memoize when you "need to"" v "just memoize all the things then you never need to think about it" 🤷 

The gotcha with them both is just that you should always have all dependencies in the arrays. The bugs come if you capture a variable in the memoised closure .. basically either
 1. "it'll never change"   - just add it - something will change later and it *will* change, and then you will get bugs as no one thought to also update the dependency arrays
 2. "I don't want 'this' to change if this other 'thing' changes, so I'll leave it out" - just add it - and then also make the reference for that 'thing' stable also. Otherwise, 'this' changes and you *do* want 'thing' to change

From my experience those really are the two things (other than the extra code you have to look through) which make `useMemo`/`useCallback` prone to bugs (also reminds me that we should really add https://www.npmjs.com/package/eslint-plugin-react-hooks )

PR #706, opened 2021-12-29
translation tidy

### LINKED ISSUE #548: Add new translation namespaces
Want to add some new translation name spaces:

- Inventory
- Replenishments
- data-table

Then shift any translations into the appropriate name spaces and also update the translation calls in the appropriate packages

mark-prins:
Fixes #548 

Moves a few things around. Maybe without breaking stuff.
Not sure if we still need the `data-table` namespace mentioned in the issue? Wasn't sure what to put in it if so.


joshxg (approved):
Nice thanks.

The data table translations aren't needed if you just duplicate all the labels used for the data table  into the json for the package which uses them - i.e. for the `BasicHeader` - then you have the 'shared columns/cells' like the `RowExpand` column which uses a `useTranslation` call internally for the tooltip and currently has the translations in `common`  which works also

PR #707, opened 2021-12-29
#703 locations to inbound

### LINKED ISSUE #703: Add location selection to inbound shipments

joshxg:
Fixes #703 

- Very annoying graphql changes needing to be added for such a simple change ..
- I've added a new tab with a single input which I agree doesn't look great - not sure where to go with it though. Maybe there will be more fields added, maybe should be showing the `code` also? 
image


packages/common/src/operations.graphql
```diff
@@ -93,6 +93,58 @@ query invoice($id: String!) {
             numberOfPacks
             packSize
             note
+            location {
```
joshxg, line 96: Going to type out my problem here .. I *think* this should be a non-issue if and when(?) the talk about changing the API to not return errors on nested fields happens.. but in the meantime

The annoyance here is:

- I want to parse the query result with the generated type "InvoiceQuery" into an `Invoice`. An `Invoice` has fields like `lines: InvoiceLine[]` rather than `lines: InvoiceLineRespinse` (which is `ConnectorError | InvoiceLineConnector`) or something.. so we have to parse the lines as we want `location: Location` rather than `location: Location | NodeError` etc
- So, we start by parsing and asserting that there is an `InvoiceNode`
- Then we parse the lines. Which involves mapping and asserting the stock line. Then we do the location.
- The location, also has a field .. `stock`. Which is a `StockLineResponse` ..
- So, if we don't query for `stock`, then the response is technically not a `LocationResponse` and so generated types don't work well. And more than that.. if we don't query for the full `StockLine`, then the types still don't work

Hm.

I think this can be fixed with using fragments. If we used a fragment of the location without the `stock` field, we should be able to generate a type ... `LocationRowFragment` or something like that and use that type rather than `LocationResponse`.

Though I'm not sure if I should go down that route and instead just wait for the API change, or start writing my own types..

packages/system/src/Location/Components/LocationSearchInput.tsx
```diff
@@ -0,0 +1,45 @@
+import React, { FC } from 'react';
```
joshxg, line 1: Not sure if this will be used elsewhere.. but maybe!

packages/system/src/Location/Components/LocationSearchInput.tsx
```diff
@@ -0,0 +1,45 @@
...
+          ...value,
+          label: value.name,
+        }
+      }
+      loading={isLoading}
+      onChange={(_, location) => {
+        onChange(location);
+      }}
+      options={defaultOptionMapper(data?.nodes ?? [], 'name')}
+      renderOption={getDefaultOptionRenderer('name')}
```
joshxg, line 41: Possibly should change this up to be something like `${code} -- ${name}`

packages/host/webpack.config.js
```diff
@@ -16,6 +16,7 @@ module.exports = env => {
   return {
     entry: './src/index',
     mode: isProduction ? 'production' : 'development',
+    devtool: 'source-map',
```
joshxg, line 19: Can revert this - it's making the initial build a little longer but gives better source maps for debugging in chrome. Might want to change it to hidden source maps for production builds

mark-prins:
when I mutate against the remote API I get this: 
```
{
    "data": null,
    "errors": [
        {
            "message": "Failed to parse \"NaiveDate\": trailing input (occurred while parsing \"UpdateInboundShipmentLineInput\") (occurred while parsing \"[UpdateInboundShipmentLineInput!]\")",
            "locations": [
                {
                    "line": 5,
                    "column": 33
                }
            ],
            "path": [
                "batchInboundShipment"
            ]
        }
    ]
}
```

I had to reformat the expiry date to get it to work. Was `30-4-2023` and I had to change to `2023-4-30` 🙄

packages/common/src/operations.graphql
```diff
@@ -93,6 +93,58 @@ query invoice($id: String!) {
...
+              ... on LocationNode {
+                id
```
mark-prins, line 98: the `__typename` isn't requested for `location` but is required by the `locationGuard` 
not sure how this is working?

packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/TabTables.tsx
```diff
@@ -69,11 +75,27 @@ export const PricingTableComponent: FC<{
...
   );
 };
 
 export const PricingTable = React.memo(PricingTableComponent);
+
+export const LocationTableComponent: FC<TableProps> = ({
+  lines,
+  updateDraftLine,
+}) => {
+  console.log('lines', lines);
```
mark-prins, line 91: can be removed?

packages/system/src/Location/Components/LocationSearchInput.tsx
mark-prins: I expect it will 😃

packages/system/src/Location/Components/LocationSearchInput.tsx
mark-prins: this looks ok to me - probably because the test locations we have a pretty simple ( A1 with code of A1! )

mark-prins (approved):
👍  great to see this come in!

I have an issue with the 'real' API - it isn't saving.. it's also not reporting an error, so hopefully just an API thing? 

Really wasn't sure how to make the location bit look better than you have it. originally thought that location could go on the first tab.. but now that it says 'quantities' that wouldn't fit and the location is better on its own tab 🤷

packages/common/src/operations.graphql
joshxg: Yeah.. think it'll likely be a while

packages/system/src/Location/Components/LocationSearchInput.tsx
joshxg: Are ya saying to do this or to leave it as is?

packages/common/src/operations.graphql
joshxg: nice catch!

packages/system/src/Location/Components/LocationSearchInput.tsx
mark-prins: heh, pretty vague. Leave as is.

Issue #500, opened 2021-11-25
Data table keyboard navigation


joshxg:
Bit of a random idea..?

I thought it would be very useful to have at least up and down working for rows of the data table

I also thought having a consistent keybinding of something like.. "cmd+f" to focus the first cell in the data table on a page would be really cool

mark-prins:
Not that random! I have been wondering about keyboard shortcut support - and tables are so central to the whole thing, we really need some way of getting around them. navigating up and down rows, with enter for the onClick action would be super   if we could. First cell focus would be helpful - maybe we also need to have a way to focus back to other input elements too? swap between table and other bits?

joshxg:
Ah yes. Good incentive to keep pages consistent is that we can have nice and easy keybindings to 'target' an area.. i.e. d for the data table, p for the side panel and b for the button area or something like that. Surely that's super easy, right? 😁

CarlosNZ:
Do we want a generic `useKeyboardShortcuts` hook? 

You initialise the hook with an array of key definitions mapped to callbacks (and maybe some options defining focus areas or something), and the hook contains the event listeners / keyboard state and runs the callbacks accordingly.

I seem to remember seeing some discussion of this on here somewhere already, but can't find it now.

CarlosNZ:
Suggested key bindings:

In "Breadcrumbs" component
- `Escape` -- goes up one level in url/breadcrumb path

In "DataTable" component
- `Up`/`Down` arrows -- move focus up/down rows in table
- `Enter` -- open detail view of focused line
(I don't think we need `Cmd-F` to focus first line, as it can just start at the first line when you first press `Down` arrow (or start at the bottom if you first press `Up`)
_Eventually (but maybe not for initial prototype):_
- `Page Up`/`Page Down` -- navigate through pages
- `Shift`/`Ctrl-Enter` -- select/check rows/row range as per standard Finder/Explorer selection rules

There's currently no concept of a "focused" row in DataTable, only the "hover" effect for mouseover, so will need to add a `focusedRow` state variable probably (and then the conditional style can follow mouseover OR isFocused)

It looks like this _should_ be do-able just using Kbar's dynamic actions list using the `useRegisterActions` hook, but will need to confirm this.

PR #708, opened 2021-12-30
#505 inbound expiry date edit

### LINKED ISSUE #505: Inbound expiry date editing
Once we have a date picker from https://github.com/openmsupply/openmsupply-client/issues/504 - we need to enable editing of expiry date on rows in an inbound.

Need to add the date picker cell and update the row. Then on the mutation, ensure that the date is correctly parsed to a `NaiveDate` which I *think* is in the format `YYYY-MM-DD` - so we will need to make sure we have a special function to parse a date into a naive date correctly (the normal expiry dates might be entered as dd-mm-yyyy?)

joshxg:
Fixes #505 

- Added an expiry date cell to the inbound edit view
image
- Went a little offroad and added some simple keybindings (simple as in, I'm sure there are better ways to do this) for switching between tabs to get some feedback on. I spent a bit trying to figure out what the best key combo was, what the best way to show the keybindings (tooltip for example) and just landed on this as possibly? the worst example (you can't type '!' '@' '#' for example) which I am sure will result in a lot of feedback? 😁


packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEdit.tsx
```diff
@@ -208,9 +232,21 @@ export const InboundLineEdit: FC<InboundLineEditProps> = ({
...
+                    <Tab
+                      value={Tabs.Pricing}
+                      label={`${t('label.pricing')} (⇧+2)`}
+                      tabIndex={-1}
+                    />
+                    <Tab
+                      value={Tabs.Location}
+                      label={`${t('label.location')} (⇧+3)`}
+                      tabIndex={-1}
+                    />
```
joshxg, line 249: The main motivation for doing this now, was that the tab order is a little strange when there are multiple rows as the first two columns are a different table.

If we add a 'focus' to the first cell of the 'tabs table' when we change tab, it might be quite nice? Not really sure.. but the alternative is to bring the columns. back into the table which means we have to manually make sure that the widths and stuff are aligned and it's a bit tedious 😁

packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEditPanel.tsx
```diff
@@ -37,16 +38,22 @@ interface InboundLineEditPanel {
   updateDraftLine: (patch: Partial<DraftInboundLine> & { id: string }) => void;
 }
 
+const expiryInputColumn = getExpiryDateInputColumn<DraftInboundLine>();
```
joshxg, line 41: Bit of a gotcha here is that whenever the `updateDraftLine` function changes (When an update happens), we need to get new columns, which means if this is not a stable reference to an object, means the `Cell` will be re-mounted, rather than re-rendered which means it will lose all internal state

joshxg (commented):
Need to handle validation also. At the moment just a border when there's an error - would like to display feedback but not sure how. Wondering if we should ask Manu or we can just make something up.

I'm thinking because you could have an error on any tab, having a ❗ icon on the far left when the row is in an error state, which you can hover to see what the error is?

mark-prins:
general comment re the keybindings - awesome!!
agree, that there might be comments about #@! - but what I like is that SHIFT is fairly transferrable across windows & mac, if we used option or command symbols, that would mean nothing to our current users. 

It would be good to focus the first editable cell on the new tab, if that is at all possible(!) when using keys, I can switch to the location tab easily, but then I have to TAB a whole lot in order to use the drop down. or use the mouse..

packages/common/src/utils/formatters.ts
```diff
@@ -1,8 +1,12 @@
+import format from 'date-fns/format';
+import isValid from 'date-fns/isValid';
+
 export const formatDate = (date: Date): string =>
   `${date.getDate()}/${date.getMonth() + 1}/${date.getFullYear()}`;
 
-export const formatNaiveDate = (date: Date): string => {
-  return formatDate(date).replace(/\//g, '-');
+export const formatNaiveDate = (date: Date): string | null => {
```
mark-prins, line 7: thanks - had an issue with that in the other PR 😄

packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEdit.tsx
mark-prins: oh. heh. I commented about exactly that. yeah - a focus would be nice, I think.

mark-prins (approved):
👍  that is looking super slick! 

couple of :nits
- the expiry date formatting in the detail view column: in mSupply this is usually MM/yyyy and currently is d/M/yyyy. my understanding is that the day isn't super useful; I wonder about just having month & year?
- if the expiry is null from the server then today's date shows as the expiry in the edit modal, which made me think that a date was set for the expiry.. but it's just the date picker defaulting. can we have a blank or something?
- seems the API isn't saving the expiry, I had to update it manually to check things

joshxg:
Ah yes - have changed the column values.

For not saving, I'm saving against the demo server OK, were you doing something different this? 
https://user-images.githubusercontent.com/35858975/147724234-876583b2-5f2e-4e1e-ac3b-f41ae4d46f51.mp4


....  (I typed that and now an hour later....)

Aaaand now strangely I'm getting errors when I edit records through the API both on the demo server and a newly initialised instance locally. I'm not sure if this is a me issue or if there is something wrong. 

image



Ugh. Going to leave that to the side for the minute..


I've updated the column. Also, I've added a cheap hacky way to focus when changing tabs, which is annoying for the batch/expiry - I think doing a better/deliberate/specific implementation would take a bit of work but might be similar in idea to https://github.com/openmsupply/openmsupply-client/issues/500 where we could create a ref map for the columns... but hopefully this'll give an idea for now

mark-prins:
looks great - liking the autofocus change 😁  see what you mean about the refs now!
Having a further look into the API not saving the date - could just be a data thing, so am doing some more sniffing around while writing up the issue for the API team.

PR #710, opened 2021-12-30
Add simple locations list view

### LINKED ISSUE #555: Locations List view
Add a locations list view! 

- Under the `inventory` nav
- `locations` package to live in `system` with a `ListView`, `DetailView`, `Service` and `Components` folders or something 

Also realising that a detail view might not be what we want for this as there's not much to it. Maybe the detail view is just a modal?

joshxg:
Fixes #555 

- Should have worked on this before the location editing in inbounds. Trying to make it a little easier to test with etc.
- Will add a small edit/create modal next


packages/system/src/Location/ListView/api.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const locationsGuard = (locationsQuery: LocationsQuery) => {
+  if (locationsQuery.locations.__typename === 'LocationConnector') {
+    return locationsQuery.locations;
+  }
+
+  throw new Error(locationsQuery.locations.error.description);
+};
+
+export const useLocationList = (): UseQueryResult<
```
joshxg, line 26: I'm being lazy and haven't added pagination or filtering.. how many locations do people have? 🤷

joshxg:
image


There's a few more columns on msupply that could be added, but of course need the API 🙃

mark-prins:
oh. that reminds me.. in the inbound shipment location select, could we have the locations ordered? I added some to the demo server API and the default order isn't pretty:

Screen Shot 2021-12-31 at 2 37 56 PM

packages/system/src/Location/ListView/api.ts
mark-prins: oh! I was clicking on the pagination control and not noticing that the list stayed the same! 🙄 
Agree.. there shouldn't be that many locations.

mark-prins (approved):
👍  great! like having the menu fleshed out a bit, with more pages and stuff. 
Looks pretty solid to me - I've just added a commit, as I suspect the name & code should be the other way around. Just revert if you think differently.

packages/inventory/src/InventoryService.tsx
```diff
@@ -17,6 +17,7 @@ export const InventoryService: FC = () => {
     <Routes>
       <Route path={stocktakesRoute} element={<StocktakeListView />} />
       <Route path={stocktakeRoute} element={<StocktakeDetailView />} />
+      <Route path={stocktakeRoute} element={<StocktakeDetailView />} />
```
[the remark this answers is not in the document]
joshxg: whoops!

joshxg:
Heh, nice catch, thanks!

PR #713, opened 2022-01-04
#711 locations detail view

### LINKED ISSUE #711: Location detail view / edit modal
Want an easy way to add/edit locations.

A full page for a single location seems a bit much?

Will try a modal on the list view, I think!

joshxg:
Fixes #711 

image

- Really basic modal for editing a location
- Right now: Updates aren't working .. well, against the demo server as it's checking against the store ID. So I think I have to log in or something and then get the list of locations for 'my store' - though I'm not sure what 'my store' is there is only an endpoint for username and password? 🤔 
- Inserts aren't working: Getting an error for a foreign key violation (which I'm getting even on a fresh init local version)


packages/system/src/Location/ListView/ListView.tsx
```diff
@@ -1,26 +1,83 @@
...
+
+interface EditModalState<T> {
+  entity: T | null;
+  mode: ModalMode | null;
+  onOpen: (entity?: T | null) => void;
+  onClose: () => void;
+  isOpen: boolean;
+}
+
+const useEditModal = <T extends unknown>(): EditModalState<T> => {
```
joshxg, line 23: will look at utilising this in the other pages also, I think

packages/system/src/Location/ListView/LocationEditModal/LocationEditModal.tsx
```diff
@@ -0,0 +1,147 @@
...
+import { Location } from '../../types';
+import { useLocationInsert, useNextLocation, useLocationUpdate } from '../api';
+interface LocationEditModalProps {
+  mode: ModalMode | null;
+  isOpen: boolean;
+  onClose: () => void;
+
+  location: Location | null;
+}
+
```
joshxg, line 22: I .. think? this is all relatively straight forward. Pretty simple modal with the `useDraftLocation` hook pretty much controlling all the things.

packages/system/src/Location/ListView/api.ts
```diff
@@ -23,6 +28,56 @@ const locationsGuard = (locationsQuery: LocationsQuery) => {
   throw new Error(locationsQuery.locations.error.description);
 };
 
+const toInsertInput = (location: Location): InsertLocationInput => ({
+  id: location?.id,
+  name: location?.name,
+  code: location?.code,
+  onHold: location?.onHold,
+});
```
joshxg, line 36: these mappings seem redundant, but if we send *any* 'extra' fields through the API, there's an error and the request is rejected. So, for instance, if we have a `Location` object with an extra `__typename`, leftover from the request itself, then if we sent that back, the mutation would fail

packages/system/src/Location/ListView/api.ts
```diff
@@ -34,13 +89,27 @@ export const useLocationList = (): UseQueryResult<
...
 
-  const result = useQuery(['locations', 'list', queryParams], async () => {
-    const result = await api.locations({
+  const result = useQuery(['location', 'list', queryParams], async () => {
+    const response = await api.locations({
       sort: [toSortInput(queryParams.sortBy)],
     });
-    const locations = locationsGuard(result);
+    const locations = locationsGuard(response);
     return locations;
```
joshxg, line 97: bit of a problem here with sorting. The sorting works fine on the list view, however using this hook in the inbound shipment/`LocationSearchInput` component then the sort gets messed up, as the `useQueryParams` is taking state from the URL, which is for the DetailView - so the sort key is set as `itemName` which is also causing an error when against the demo server API, as `itemName` is not a sort key for location.

Thinking:
- Use another hook for the list which just uses `sortBy`
- Do something about `useQueryParams` and the url state

Unsure which right now

packages/system/src/Location/ListView/AppBarButtons.tsx
```diff
@@ -0,0 +1,43 @@
...
+        />
+        <ButtonWithIcon
+          Icon={<DownloadIcon />}
+          label={t('button.export')}
+          onClick={success('Downloaded successfully')}
+        />
+        <ButtonWithIcon
+          Icon={<PrinterIcon />}
+          label={t('button.print')}
+          onClick={info('No printer detected')}
```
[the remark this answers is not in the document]
mark-prins: I thought yes, though 🤷 
Since mSupply has print, might as well retain that. it doesn't have export though - so I think to remove that one

packages/system/src/Location/ListView/api.ts
mark-prins: ugh. am a little concerned about the `useQueryParams` and url state implementation now, partly due to the navigation getting messed up with it too. not sure at this stage.. two versions of `useQueryParams`? something to indicate whether the current page and the data for it is tied to the URL or not?

joshxg: Ah yes I had forgotten about this .. lucky you're here.

I'm thinking two versions would be easier. 

I'm hesitant, but I think that the benefit for the feature just isn't there right now and that it might just be better to cop it and back out of using the URL like this for now.. 😢 🤷

mark-prins: sad. but you are probably right and should just take it out

PR #726, opened 2022-01-11
#725 disabled state

### LINKED ISSUE #725: Disabled state isn't very clear
The input fields are using a default disabled state, and this looks pretty similar to the enabled version. 

A suggestion is to use `#FAFAFC` as the background colour, like this:

telegram-cloud-photo-size-1-5075985452598667767-x

mark-prins:
Fixes #725 

Added a different background for the disabled input states and updated the stories to demonstrate / test these.


joshxg (approved):
Nice!

PR #728, opened 2022-01-11
#727 spinner

### LINKED ISSUE #727: Add spinner
Would be nice to have a more interesting spinner now that #722 has been done.
Thought it might be fun to add an option or two, and see what is acceptable.

mark-prins:
Fixes #727 

added a couple of loader options and then couldn't decide which to use.. so added a randomiser 🤷 

The loaders are in storybook - we could just take the random one out and set Host back to basic, and get some opinions. I think the two loaders are pretty inoffensive though? Am very slightly bothered that the loader shows in the secondary colour and that the linear gradient wouldn't show on the mSupply guy.. but not bothered enough to look further into it

Also added the option to specify a 'saving' loader when using the basic one - I thought the 'loading' message when saving an inbound shipment item would be better as a 'saving' message.


joshxg (approved):
Love it. Torn between bouncy guy and pulse 😭

PR #729, opened 2022-01-11
#716 stocktake save button 2

### LINKED ISSUE #716: Remove save button on stocktake detail view
Epic for 'removing the save button' on stock take detail view!

This is still using 100% mock server. The backend is still a bit away.

joshxg:
Not quite enough to fix #716 

- This makes the lines/summary items etc consistent with inbound/requisitions
- Adds the many many hooks!
- Breaks a couple things - and adds a few TODO's (Will point out!)
- Sorting for the 'grouped' view is not working right now


packages/inventory/src/Stocktake/DetailView/ContentArea.tsx
```diff
@@ -1,21 +1,86 @@
...
+        borderRadius={4}
+        bgcolor="#c7c9d933"
+      >
+        <span style={{ whiteSpace: 'pre-wrap' }}>
+          {JSON.stringify(rowData, null, 2)}
+        </span>
+      </Box>
+    </Box>
+  );
+};
```
joshxg, line 33: TODO: Improved Expando

packages/inventory/src/Stocktake/DetailView/ContentArea.tsx
```diff
@@ -1,21 +1,86 @@
...
+  >({
+    key: 'itemName',
+  });
+  const { data: lines } = useStocktakeLines();
+  const { data: items } = useStocktakeItems();
+
+  const rows = isGrouped
+    ? items?.sort(
+        getDataSorter(sortBy.key as keyof StocktakeSummaryItem, !!sortBy.isDesc)
+      )
```
joshxg, line 47: TODO: This should be using the `getColumnSorter` but I had an issue with it that I'll need to address

packages/inventory/src/Stocktake/DetailView/DetailView.tsx
```diff
@@ -1,173 +1,63 @@
...
-  const { draft, onChangeSortBy, sortBy } = useDraftStocktake();
-
-  const onChangeItem = (item: StocktakeItem | null) => {
-    setModalState(state => ({ ...state, item }));
-  };
+  const [modalState, setModalState] = useState<{
+    item: Item | null;
+    mode: ModalMode;
+    isOpen: boolean;
+  }>({ item: null, mode: ModalMode.Create, isOpen: false });
```
joshxg, line 32: TODO:  Made a `useEditModal` hook in `location/ListView` which could be utilised for all of the edit modals

packages/inventory/src/Stocktake/DetailView/columns.ts
```diff
@@ -2,42 +2,166 @@ import {
...
+          }
+        },
       },
 
       GenericColumnKey.Selection,
     ],
-    { onChangeSortBy, sortBy },
-    [sortBy]
+    { sortBy, onChangeSortBy },
+    [sortBy, onChangeSortBy]
```
joshxg, line 166: TODO: I'm not sure if these columns are the right ones we want - so need to double check that

packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/StocktakeLineEdit.tsx
```diff
@@ -17,156 +11,126 @@ import {
...
-    ...stockLine,
-  };
-};
-
 export const StocktakeLineEdit: FC<StocktakeLineEditProps> = ({
   item,
-  draft,
-  onChangeItem,
   mode,
+  onClose,
```
joshxg, line 39: TODO: This is re-rendering a little much also - so need to address that!
TODO: The Panel should just use a `DataTable` like inbound - though the tabbing issue is a problem, so not sure if we need a different solution here.
TODO: Handle the "Next" button
TODO: Handle loading states after clicking OK

packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/StocktakeLineEditForm.tsx
```diff
@@ -47,24 +30,14 @@ export const StocktakeLineEditForm: FC<InboundLineEditProps> = ({
...
-              return !itemAlreadyInShipment;
-            }}
+            currentItem={item}
+            onChange={onChangeItem}
+            // extraFilter={item => {
+            //   const itemAlreadyInStocktake = draft.lines.some(
+            //     ({ id, isDeleted }) => id === item.id && !isDeleted
+            //   );
+            //   return !itemAlreadyInShipment;
+            // }}
```
joshxg, line 40: TODO: Correctly filter the list of all items currently already being counted

joshxg (commented):
Also need to handle deleting lines properly.

Bit of a long list of TODO's and this is a bit of a mess. I was getting overwhelmed and confused so wanted to have a bit of a restart and properly target smaller issues - though it makes your job harder .. sorry :(

mark-prins (approved):
👍  that's looking awesome! a lot of changes in a lot of files.. basically, it looks like a lot of work!!
Thanks for sorting it all out

Can I add a couple of extra todos:
- Add item > select item in modal > cancel > Add item : the same item is shown, would be good to clear the state on cancel
- If there is no stock, the form isn’t shown. I think the table should show so that rows can be added manually.. with the mock API, I can't actually add stocktake items!

packages/inventory/src/Stocktake/DetailView/columns.ts
joshxg: Yes to all! Thanks

packages/inventory/src/Stocktake/DetailView/DetailView.tsx
```diff
@@ -1,173 +1,63 @@
...
+import { ContentArea } from './ContentArea';
+
+export const toItem = (line: StocktakeLine | StocktakeSummaryItem): Item => ({
+  id: 'lines' in line ? line.lines[0].itemId : line.itemId,
+  name: 'lines' in line ? line.lines[0].itemName : line.itemName,
+  code: 'lines' in line ? line.lines[0].itemCode : line.itemCode,
+  isVisible: true,
+  availableBatches: [],
+  availableQuantity: 0,
+  unitName: 'bottle',
```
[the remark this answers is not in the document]
joshxg: Yes, thanks, you're right. Annoying that we can't access the unit through the invoice line.

packages/inventory/src/Stocktake/DetailView/columns.ts
```diff
@@ -2,42 +2,166 @@ import {
...
+  useColumns<StocktakeLine | StocktakeSummaryItem>(
     [
-      'itemCode',
-      'itemName',
-      'batch',
-      'expiryDate',
+      [
+        'itemCode',
+        {
+          getSortValue: row => {
```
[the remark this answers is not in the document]
joshxg: I don't think i'm following, but I certainly agree this area needs help, so am keen to find out!

- The shape: Yeah, I think having the basic shape is for the best and although adding the item name and code fields help for a couple of columns, a solution to make all of the columns better is surely possible.
- Helpers: Do you mean simply having `const getItemNameAccessor = (rowData) => ...` and using the same function for both the `getSortValue` and `accessor`?
- Constructor: Do you mean, for example, we could have in `useColumns`, the default `getSortValue` is equal to the `accessor`?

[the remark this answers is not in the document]
mark-prins: I'm not following either.. sorry, not sure what was in my head when I wrote that.
Looking at it now - there's the duplication of code for the `accessor` and `getSortValue`, as we've talked about before. Makes sense to me that the sort value is the accessor value unless explicitly set otherwise.

Would be useful to have a helper as well, which returns for a given key either the value or the `ifTheSameElseDefault` value from another property ( 'lines' in this case ). Easy enough to write, a tricky one to type correctly I expect.

PR #737, opened 2022-01-11
persist side panel state

### LINKED ISSUE #694: Should the side panel open/closed state persist to local storage?
Wondering if we should persist the side panel state to local storage similar to the grouped state?

(Also, side note: wonder if we should keep the 'grouped' state *only* in local storage, and not in the table context state? 🤔

mark-prins:
Fixes #694 

scratches an itch.. which, it turns out, shouldn't have been a problem 🙄 
The screen size detection was incorrect for the side panel, and on my little laptop screen it shouldn't have been showing anyway. Oh well, have persisted the state regardless.

Found that the panel should only show automatically at large screen sizes:

Screen Shot 2022-01-12 at 9 44 34 AM

### DIFF
--- a/packages/common/src/ui/components/inputs/Autocomplete/Autocomplete.tsx
+++ b/packages/common/src/ui/components/inputs/Autocomplete/Autocomplete.tsx
@@ -73,6 +73,9 @@ export function Autocomplete<T>({
       sx={{ width }}
     />
   );
+  const getOptionLabel = (option: { label?: string } & T): string => {
+    return option.label ?? '';
+  };
 
   return (
     <MuiAutocomplete
@@ -94,6 +97,7 @@ export function Autocomplete<T>({
       renderInput={renderInput || defaultRenderInput}
       renderOption={renderOption}
       onChange={onChange}
+      getOptionLabel={getOptionLabel}
     />
   );
 }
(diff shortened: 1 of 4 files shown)


packages/common/src/ui/components/inputs/Autocomplete/Autocomplete.tsx
```diff
@@ -73,6 +73,9 @@ export function Autocomplete<T>({
       sx={{ width }}
     />
   );
+  const getOptionLabel = (option: { label?: string } & T): string => {
```
mark-prins, line 76: this is just to fix a complaint when you navigate to the detail page of an invalid shipment, where you get this:

Screen Shot 2022-01-12 at 9 47 07 AM

because the Autocomplete `getOptionLabel` func defaults to `option.label ?? option` 

maybe not needed? should fix at the caller not here?

joshxg: Well I think I've seen this several times so obviously keep making the same mistake! Probably a good idea to just do it here, thanks!

joshxg (approved):
Yup that's great, thanks.

I have a feeling that we could/should have the `LocalStorage` class also be listening to the browser events so we can track changes from different tabs, but maybe not

mark-prins:
hm. I had a go at adding a listener for storage events and ended up a little tangled, so have decided not to go with it.
in the process found a bug where empty side panels are shown - if the panel is opened in the detail view, returning to the list view will keep the empty panel open 🙄

PR #742, opened 2022-01-12
Add filtering of items

### LINKED ISSUE #736: Stocktake add item modal - correctly filter out items already being counted
Currently when adding an item, you can select an item from the drop down. Would be consistent with the other pages to filter out any items already being counted within the stocktake

joshxg:
Fixes #736 

- Filters out the items that have already been counted from being selected again.


mark-prins (approved):
👍  very nice!
Now that I'm looking at the add item modal.. it would be good if the item autocomplete had focus on load, so I could start typing straight away 🤔

joshxg:
Yes! That's would be nice - will have a looksie

PR #743, opened 2022-01-12
Add update columns for detail view and add/edit item modal for stocktake

### LINKED ISSUE #739: Stocktake detail view columns
Some slight adjustments to the stocktake detail view columns:

- change the order of snapshot and counted; makes more sense to me having counted to the left of the snapshot
- could we add the row expando action column please?
- location is probably useful, I think after expiry
- pack size is also probably good. before snapshot?
- don't know about comment
- and not for now: Ivory Coast have the reason column as well

Additionally, the `columns.ts` for this view needs some tidying!

joshxg:
Fixes #739 

- Haven't added the location field - mainly because the mock hasn't got a location field. Thought to leave that for now until we integrate the backend API.


joshxg:
Not sure how to make the `getSortValue` better. Had a look at trying to make the default equal to the `accessor`. Forgot that `accessor` is returning `unknown` and the sort accessor needs number | string 🤔

packages/inventory/src/Stocktake/DetailView/columns.ts
```diff
@@ -114,52 +117,56 @@ export const useStocktakeColumns = ({
           },
         },
       ],
-
+      ['packSize', { width: 125 }],
```
mark-prins, line 120: :nit the 'Expiry date' column could be less wide, and the name wider

mark-prins (approved):
🎉  wicked! looks good!

PR #744, opened 2022-01-12
Add StocktakeDetailView row expansion

### LINKED ISSUE #730: Stocktake detail view expando
The current expando for the Stocktake DetailView is a placeholder. We need to implement the real thing.

joshxg:
Fixes #730 

- Added the expiry date formatter
- Changed some styles to improve the inbound shipment table and also ensure the expando is flexed (do we want this..? Seemed weird when the expand didn't fill the whole width, especially when the screen size was smaller and you were scrolling)

### DIFF
--- a/packages/inventory/src/Stocktake/DetailView/columns.ts
+++ b/packages/inventory/src/Stocktake/DetailView/columns.ts
@@ -110,14 +110,35 @@ export const useStocktakeColumns = ({
                 'expiryDate',
                 null
               );
-              return formatExpiryDate(expiryDate);
+              return expiryDate;
             } else {
-              return formatExpiryDate(rowData.expiryDate);
+              return rowData.expiryDate;
             }
           },
         },
       ],
-      ['packSize', { width: 125 }],
+      [
+        'packSize',
+        {
+          width: 125,
+          getSortValue: row => {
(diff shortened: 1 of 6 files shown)


packages/inventory/src/Stocktake/DetailView/columns.ts
```diff
@@ -172,3 +172,24 @@ export const useStocktakeColumns = ({
     { sortBy, onChangeSortBy },
     [sortBy, onChangeSortBy]
   );
+
+export const useExpansionColumns = (): Column<StocktakeLine>[] =>
```
mark-prins, line 176: In the non-expansionColumns.. the packSize needs to access data from `lines` rather than `rowData`.. currently the size only shows if non-grouped.

The expiry date isn't showing at all - needs to return a date rather than the formatted string

mark-prins (approved):
👍  sweet!

that was quick! 😁  looking good!

packages/inventory/src/Stocktake/DetailView/columns.ts
joshxg: Ah - sorry about that. Silly mistakes! Have made the changes and also fixed the pack size width in here also!

PR #746, opened 2022-01-13
Add aggregated row sorting

### LINKED ISSUE #731: Stocktake detailview sorting aggregated items
Currently the sorting on the Stocktake DetailView is not working for aggregated/grouped items.

The problem, I think, is that I used the `useSortBy` hook to help implement sorting. The callback doesn't have access to the `column` which is where the `getSortValue` fn is for the column, so will need a bit of juggling around.

joshxg:
Fixes #731 

- Made the parameter for the `sortBy` column be the `column`. Not sure if that's the best idea. Maybe I should also move the hook to the tables module since it's so coupled


packages/common/src/hooks/useSortBy/useSortBy.stories.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import { useSortBy } from './useSortBy';
 import { Story } from '@storybook/react';
-
+import { useColumns, Column } from '../../ui/layout';
```
mark-prins, line 4: curious - the `useSortBy` story doesn't run for me: am getting `useLocation() may be used only in the context of a <Router> component`

packages/common/src/hooks/useSortBy/useSortBy.ts
```diff
@@ -34,12 +35,13 @@ export const useSortBy = <T>({
     direction: getDirection(initialIsDesc),
   });
 
-  const onChangeSortBy = useCallback((newSortRule: SortRule<T>) => {
+  const onChangeSortBy = useCallback((newSortRule: Column<T>) => {
```
mark-prins, line 38: here, and in the interface - `column: Column<T>` makes more sense to me than `newSortRule` since there is a different shape for a sortRule object. but 🤷

packages/inventory/src/Stocktake/api/hooks.ts
```diff
@@ -107,14 +110,40 @@ export const useStocktakeRows = (isGrouped = true) => {
...
+    const currentColumn = columns.find(({ key }) => key === sortBy.key);
+    if (!currentColumn?.getSortValue) return items;
+    const sorter = getColumnSorter(
+      currentColumn?.getSortValue,
+      !!sortBy.isDesc
+    );
+    return items?.sort(sorter);
+  }, [items, sortBy.key, sortBy.isDesc]);
+
+  const sortedLines = useMemo(() => {
```
mark-prins, line 125: wasn't sure why the `sortedLines` aren't used below ( line 135 ). the lines are sorted again down there.
the sortedLines aren't used yet, so I couldn't guess intention from usage 😄

mark-prins (approved):
👍  works nicely!

have some thoughts about being tightly coupled to columns and tables.. but think it's fine. The use case for sorting data is to display the sorted data in tables!

joshxg:
I'm concerned about the coupling also .. just not sure how else to make it work well. Re the story - the `useSearchParameters` hook is messing with things so I added a testing router to the story book preview file 🤔

packages/common/src/hooks/useSortBy/useSortBy.ts
joshxg: Yeah you're right again!

packages/inventory/src/Stocktake/api/hooks.ts
joshxg: whoops 🙄

mark-prins:
storybook works like a charm - that darn `useSearchParameters`! starting to think that the query string url thing is more bad than good. still reluctant to remove it.. but..

PR #747, opened 2022-01-13
#732 use edit modal hook usage

### LINKED ISSUE #732: Use `useEditModal` hook in detail view pages
There is a hook here https://github.com/openmsupply/openmsupply-client/blob/f1f5644b106fd9d50b408e9869a44a5ad013a620/packages/system/src/Location/ListView/ListView.tsx#L23

which would be well suited to use in most/all detail views to manage the edit modal.

joshxg:
Fixes #732 

- Just using the `useEditModal` from location to stocktake and inbound. Missed outbound and requisitions (not done) so far


mark-prins (approved):
😁  that's very classy! very useful, nice and clean and easy to use.

PR #748, opened 2022-01-14
#709 dom violation

### LINKED ISSUE #709: DOM layout violation 😨
We have invalid dom structure. I believe here:

```
<TableBody sx={{ flex: 1, display: 'flex', flexDirection: 'column' }}>
          <TransitionGroup>
            {data.map((row, idx) => {
              return (
                <Collapse key={row.id}>
                  <DataRow
                    rows={data}
                    ExpandContent={ExpandContent}
                    rowIndex={idx}
                    columns={columns}
                    onClick={onRowClick}
                    rowData={row}
                    rowKey={String(idx)}
                    dense={dense}
                  />
                </Collapse>
              );
            })}
          </TransitionGroup>
        </TableBody>
```

- `tbody` needs `tr` as children and I believe the group is a `div`
- And `Collapse` is maybe another div?
- Haven't looked into it properly, but guessing there's a `component` prop

mark-prins:
Fixes #709

Had an issue with using `Collapse`, since it uses nested `div`s to get the animation effect.
So I've changed to use a `Fade` which affects the behaviour on first render of the table, I think it's kinda nice, so have left it. Can instead use props to only implement the fade when grouped if you prefer.

Couple of minor offroads - one was the table header change: the background was made white rather than transparent, which resulted in this:

Screen Shot 2022-01-14 at 10 13 31 AM

I've put a border around the mini table to soften the impact of the square top of the first table row:

Screen Shot 2022-01-14 at 1 53 33 PM

not sold on the big rounded corners.. but hopeful ok? 🤷

### DIFF
--- a/packages/common/src/intl/IntlProvider.tsx
+++ b/packages/common/src/intl/IntlProvider.tsx
@@ -13,8 +13,8 @@ export const IntlProvider: React.FC = ({ children }) => {
     const expirationTime =
       process.env['NODE_ENV'] === 'development'
         ? 0
-        : 7 * 24 * 60 * minuteInMilliseconds;
-
+        : // TODO: change back to a week when things are stable
+          60 * minuteInMilliseconds; // 7 * 24 * 60 * minuteInMilliseconds;
     i18next
       .use(initReactI18next) // passes i18n down to react-i18next
       .use(Backend)
--- a/packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
+++ b/packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
@@ -93,11 +93,8 @@ export const DataRow = <T extends DomainObject>({
           <Collapse
             sx={{
               flex: 1,
-              display: 'flex',
               '& .MuiCollapse-wrapperInner': {
-                flex: 1,
                 display: 'flex',
-                flexDirection: 'column',
               },
             }}
             in={isExpanded}
--- a/packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
+++ b/packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
@@ -4,7 +4,7 @@ import TableCell from '@mui/material/TableCell';
 import { Column } from '../../columns/types';
 import { DomainObject } from '@common/types';
 import { useExpanded, useDisabled } from '../../context';
-import { Collapse } from '@mui/material';
+import { Collapse, Fade } from '@mui/material';
 
 interface DataRowProps<T extends DomainObject> {
   columns: Column<T>[];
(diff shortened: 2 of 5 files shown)


packages/common/src/intl/IntlProvider.tsx
```diff
@@ -13,8 +13,8 @@ export const IntlProvider: React.FC = ({ children }) => {
     const expirationTime =
       process.env['NODE_ENV'] === 'development'
         ? 0
-        : 7 * 24 * 60 * minuteInMilliseconds;
-
+        : // TODO: change back to a week when things are stable
```
mark-prins, line 16: because the recent changes on the demo server don't have the translations showing yet.
thought that while we are changing often, an hour is probs enough of a balance between improving performance and actually showing the right words?

packages/common/src/ui/layout/tables/components/Expand/MiniTable.tsx
```diff
@@ -17,14 +17,16 @@ export const MiniTable = <T extends DomainObject>({
   JSXElementConstructor<MiniTableProps<T>>
 > => {
   return (
-    <Box p={1} style={{ padding: '0 100px', width: '100%' }}>
+    <Box p={1} style={{ padding: '0 100px 5px 100px', width: '100%' }}>
```
mark-prins, line 20: adds the padding below the row, just to space it out from the next row a little more

packages/invoices/src/OutboundShipment/DetailView/tabs/GeneralTab.tsx
```diff
@@ -59,13 +59,13 @@ export const GeneralTabComponent: FC<
...
   const { isGrouped, toggleIsGrouped } = useIsGrouped('outboundShipment');
 
-  const activeRows = useMemo(() => {
-    const x = data
-      .filter(({ isDeleted }) => !isDeleted)
-      .slice(pagination.offset, pagination.offset + pagination.first);
-
-    return x;
-  }, [data]);
+  const activeRows = useMemo(
```
mark-prins, line 62: one that will be replaced anyway I expect.
There's a funny bug with the outbound - if you view a detail page, go back to the list view and return to the same detail, the number of rows is added to. Thought that this will probably be fixed with the refactor 🤞

packages/common/src/intl/IntlProvider.tsx
joshxg: Yeah certainly

packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
```diff
@@ -36,66 +36,65 @@ export const DataRow = <T extends DomainObject>({
...
-                column={column}
-                rowKey={rowKey}
-                columnIndex={columnIndex}
-                rowIndex={rowIndex}
-              />
-            </TableCell>
-          );
-        })}
-      </TableRow>
+      <Fade in={true} timeout={500}>
```
joshxg, line 39: The fade I think is even nicer than the collapse!

packages/invoices/src/OutboundShipment/DetailView/tabs/GeneralTab.tsx
joshxg: Ah yes most likely

packages/common/src/ui/layout/tables/components/Expand/MiniTable.tsx
```diff
@@ -17,14 +17,16 @@ export const MiniTable = <T extends DomainObject>({
...
       <Box
         flex={1}
         display="flex"
         height="100%"
         borderRadius={4}
         sx={{
           backgroundColor: theme => alpha(theme.palette.gray.light, 0.2),
+          border: theme => `1px solid ${alpha(theme.palette.gray.light, 0.2)}`,
```
joshxg, line 28: I actually really like the look!

joshxg (approved):
Nice! Thanks for that it's awesome

joshxg:
> Can instead use props to only implement the fade when grouped if you prefer.

I reckon it's awesome like you have it!

Issue #715, opened 2022-01-06
Remove save button in outbound shipments


joshxg:
Epic for 'removing the save button' on outbound shipment detail view

PR #749, opened 2022-01-14
#715 Add API Folder

joshxg:
- An effort to try and break down the issue.
- Somewhat of an arbitrary split in code to PR this. It's essentially just a copy paste of the current queries into this folder similar to how the other packages are structured

### DIFF
--- a/packages/invoices/src/OutboundShipment/api/api.ts
+++ b/packages/invoices/src/OutboundShipment/api/api.ts
@@ -1,0 +1,209 @@
+import {
+  //   InsertOutboundShipmentLineInput,
+  //   DeleteOutboundShipmentLineInput,
+  //   UpdateOutboundShipmentLineInput,
+  RecordPatch,
+  OmSupplyApi,
+  UpdateOutboundShipmentInput,
+  InvoiceNodeStatus,
+  UpdateOutboundShipmentStatusInput,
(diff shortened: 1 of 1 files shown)


packages/invoices/src/OutboundShipment/api/api.ts
```diff
@@ -0,0 +1,209 @@
...
+
+  throw new Error('Unknown');
+};
+
+const itemGuard = (item: ItemResponse): ItemNode => {
+  if (item.__typename === 'ItemNode') {
+    return item;
+  }
+
+  throw new Error('Unknown');
```
joshxg, line 85: I wanted to make a `transform.ts` or something like that. However we run into a 'type' problem where determining the exact return value can often be a little murky. For instance I have here that this fn returns an `itemNode`, but that's because I have added in all the extra fields like `availableBatches` to the query.. if it wasn't a 'full' `ItemNode` then would have to create new types to add to the return value. 

Really just another thing to sort out soon with fragments

For instance we could create fragments like this,

```
fragment PartialItemNode {
    id
    code
    node
}.  
```

and use them in our queries, then we could generate types...

Though in saying that, hopefully one day we will have a less .. 'hardwork' api?

mark-prins (approved):
👍  that LGTM!

PR #753, opened 2022-01-14
#715 remove outbound shipment save button 3

joshxg:
- This is essentially updating all of the actual invoice fields to utilise hooks rather than the reducer/draft - but not the lines, yet.
- Couple of things that I'd prefer to fix a little later down the road: 
1. Deleting I've commented out for now. Will create a custom hook for this. It's essentially not implemented for all the pages, so I might as well leave it at this point and implement this for all pages at the same time.
2. The mock server/worker status changes aren't working well - need to check up on it, but I suspect the dates aren't being set correctly. However, changing status on the demo server is working. Since we need to address the invoice statuses (with adhoc/stocktransfers etc) anyway, I thought to bundle this problem with that instead 😛


packages/invoices/src/OutboundShipment/api/api.ts
```diff
@@ -0,0 +1,209 @@
+import {
```
joshxg, line 1: hmm. What did I break 🤔

joshxg:
- Hopefully by now this is looking all pretty familiar!
- Next step is to change the `GeneralTab` component to be as the other packages.
- Then, moving onto the `ItemModal` and updating that to hopefully 🤞 have less props and be a little simpler

packages/invoices/src/OutboundShipment/DetailView/SidePanel.tsx
```diff
@@ -101,45 +107,28 @@ const RelatedDocumentsSection: FC<SidePanelProps> = () => {
...
+        <DetailPanelAction
+          icon={<CopyIcon />}
+          title={t('link.copy-to-clipboard')}
+          onClick={copyToClipboard}
+        />
       }
     >
-      <AdditionalInfoSection draft={draft} />
-      <RelatedDocumentsSection draft={draft} />
+      <AdditionalInfoSection />
```
mark-prins, line 130: a minor thought.. the `dev.log-draft` entries could be removed from the translation files now

mark-prins (approved):
👍  that all looks good - thanks. yeah, starting to look familiar.. and much cleaner overall

packages/invoices/src/OutboundShipment/DetailView/SidePanel.tsx
joshxg: Ah yes! 

I will remember to remove this - for now I'd like to resolve the conflicts and merge as at a guess I'll just cause more conflicts by touching the translation files!

PR #754, opened 2022-01-14
Add updated outbound generaltab to use consistent pattern for content…

joshxg:
- Here we make the rows in the detail view consistent while breaking other things. Namely the row click.


mark-prins (approved):
🎉  LGTM!

yep! can confirm that the row click isn't popping the modal up. 
at this stage, I'm assuming issues are things in progress. 
What I've noticed is just
- the column sort throws an error
- for detail pages with multiple lines (e.g. #505402, using demo API).. viewing detail, going to list, viewing detail again: throws an error. invoices with single lines are fine 🤷

PR #755, opened 2022-01-14
default pack size

### LINKED ISSUE #752: Outbound edit modal : pack size state
there's an issue with the pack size state not being retained in the edit modal

- change pack size
- change issue quantity
- the pack size reverts to the original value

mark-prins:
Fixes #752 

couple of daft bugs 🐛  🙄 

As in the issue: changing the quantity would reset any changes to the selected pack size. The hook was happily resetting on every change. stopping that meant that the value was retained when editing a different line, which is a different problem. I've worked around by resetting the packSizeController state when you change the modal.


joshxg (approved):
That's great, thanks!

joshxg:
Will merge this one in :)

Issue #497, opened 2021-11-25
Redirect when an invoice can't be found


joshxg:
When navigating to `outbound-shipment/X` where X is an invoice which can't be found - redirect the user away .. or, show some sort of "can't find, go back" screen?

mark-prins:
Fixed with #760

PR #760, opened 2022-01-27
#620 detail loading

### LINKED ISSUE #620: Loading state of a detail view always says "no items"
When navigating to a detail view, when the draft hasn't been fetched yet, the data table uses the 'no data message' rather than an indication that the data is being fetched.

Need to use the `loading` state of the query, rather than just an empty array of lines.

Have been wondering about using https://mui.com/components/skeleton/#main-content - maybe a good place to make a 'data table skeleton' type component and use that as the loading indicator here.

Since most/all the pages are so similar, I imagine it would be quite nice to be able to make skeleton components for a 'detail view' or 'data table' etc

mark-prins:
Fixes #620 #497 

Working on 620 - but there was a bit of crossover with 497, so did that too.

Haven't finished off the outbound shipment, as that will be changing soon to a more consistent implementation.
Have removed the retries on the query client, as that was making the page super slow to load when you use an invalid ID - the retry x5 with a delay was a bit much. Maybe we should just retry for mutations 🤷  but thought that if the query fails, it is likely to fail again.. not seeing much value in retrying a bunch of times


packages/common/src/ui/layout/skeletons/ButtonSkeleton.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import { Skeleton } from '@mui/material';
+
+export const ButtonSkeleton: React.FC = () => {
+  return (
+    <Skeleton
+      variant="rectangular"
+      height={40}
+      width={125}
+      style={{ borderRadius: 24, transform: 'scale(1,1)' }}
```
mark-prins, line 10: by default the skeletons scale to `(1, 0.6)` in order to give some spacing. it means that the buttons aren't the right height, and I don't need it to scale / space here

packages/host/src/Host.tsx
```diff
@@ -47,6 +47,7 @@ const queryClient = new QueryClient({
   defaultOptions: {
     queries: {
       refetchOnWindowFocus: false,
+      retry: false,
```
mark-prins, line 50: this one - needs some further discussion

packages/inventory/src/Stocktake/DetailView/DetailView.tsx
```diff
@@ -40,5 +52,18 @@ export const DetailView: FC = () => {
         />
       )}
     </TableProvider>
+  ) : (
+    <AlertModal
```
joshxg, line 56: This is looking so good!

You likely already thought about it but I suppose in the future it would might nice for some sort of alerting 'service'.. I suppose similar to what you have done with the `useNotifications`..  some API like 

```
const { error, notification, warning } = useAlert(..)
```

Where we could just call the callbacks to open this type of modal. 

But I can see that being a decently big job!

joshxg (commented):
thats awesome! Wondering if we can use the same modal for alerting when statuses are changing which I remember being requested or if we should use another 🤔

PR #763, opened 2022-01-27
make kbar pretty again

### LINKED ISSUE #761: CMD+K layout
Something has gone wrong with the z index of the cmd K window
Possibly also time to upgrade and migrate the deprecated methods

Screen Shot 2022-01-28 at 9 34 46 AM

mark-prins:
Fixes #761 

- Updated kbar ( for no good reason really, there was an error, but it was due to our config )
- Changed the styles of a few parts
  - Search bar is now less orange ( opacity reduced to 0.2 so that the text is visible ), added a border
  - Opacity of background + background colour
  - Added z-index, though only required when shown over the data table due to the newly added zIndex of the header
  - Some padding around list results
  - border radius on results
- Modified the action list, added a bunch of routes, removed ones that were not working ( new outbound and navigation back )
- lifted kbar out to a sep file
- changed routes to use route builder
- added translation support for items

Screen Shot 2022-01-28 at 11 55 36 AM

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -27,7 +27,7 @@
     "i18next-http-backend": "^1.3.1",
     "i18next-localstorage-backend": "^3.1.3",
     "immer": "^9.0.7",
-    "kbar": "^0.1.0-beta.24",
+    "kbar": "^0.1.0-beta.27",
     "lodash": "^4.17.21",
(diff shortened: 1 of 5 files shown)


joshxg (commented):
Ahh, thanks so much for tidying this up!

PR #766, opened 2022-02-01
tidy autocomplete styles

### LINKED ISSUE #679: `NameSearchInput` on detail views could look better
Just taking a little walk around and being picky.

image

This input here has been annoying me somewhat 🤔 

Some possible things:
- Wider? Though, don't really want the input itself to be much wider.. the popover could be wider.. or the input itself could expand when you press on it, like the search bar on the list view pages
- More shadow around the popover?

mark-prins:
Fixes #679 

Added the option to expand the popper width as needed, when doing so have set the `minWidth` because I think it looks silly when the popper is not as wide as the input.

Added `boxShadow` to all Autocompletes as I think it looks nice.

Updated the name search input to use auto width; added a story to demonstrate the option.


joshxg (approved):
Very nice 😁

PR #767, opened 2022-02-02
remove batch & expiry table

### LINKED ISSUE #734: Detail view add/edit item modal: 'Static' panel of tabs
The tabbing of the 'static' panel/part of the tabs in a few of the edit item modals (The inbound shipment is the best to test with): 

image

is a little weird.

Because this is two tables, the tab order doesn't move through the row nicely and instead does just the static area. The motivation for implementing this as two tables is to ensure the LHS columns are a static width. Because we use flexing widths, if the widths differ between columns it's likely that the 'static' columns change also.

In some ways this is not too bad, because it means when you navigate to a tab, you're able to tab just the columns of that tab. However, the initial tab is then a little odd.

So, I'm not sure. Probably better to just have a standard & predictable tab order, moving from left to right, which would mean that this should become a single table. Then, we could look at redoing the original implementation using a background with an absolute position. Some other options could be to implement a way to 'style' a column (i.e. add a background to the column) - which might be useful. Also could look at implementing a 'maxWidth' property in addition to the current `minWidth` and `width` properties

mark-prins:
Fixes #734 

Removed the batch and expiry table and moved those columns into each of the other tables.
To support that -
- added a `maxWidth` prop
- added a `backgroundColor` prop : went with the specific use case, rather than a more helpful implementation of `style`. when we need that, we can add it in and remove the background colour
- adjusted the row padding: as the fixed height, combined with the really big padding meant that the background overlayed and looked too dark in between the rows. As it is, I've used slightly more padding than necessary to hide the expando row

### DIFF
--- a/packages/common/src/ui/layout/tables/columns/types.ts
+++ b/packages/common/src/ui/layout/tables/columns/types.ts
@@ -74,6 +74,8 @@ export interface Column<T extends DomainObject> {
 
   width: number;
   minWidth: number;
+  maxWidth?: number;
+  backgroundColor?: string;
 
   Cell: JSXElementConstructor<CellProps<T>>;
   Header: JSXElementConstructor<HeaderProps<T>>;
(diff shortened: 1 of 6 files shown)


joshxg (approved):
Sweet!

PR #775, opened 2022-02-07
#733 modal perf issues

### LINKED ISSUE #733: Modal perf issues
The modal in particular on detail views is lacking in performance!

Suspicion is it is re-rendering too many times.

joshxg:
Fixes #733 

image

- The main thing I concentrated on here was the heavy rendering. There were previously quite a few renders 200ms+. Now the heaviest is 40ms. It feels snappy!
- Memo'd the `DataTable` with a hacky solution (because of generic types)
- Removed the `useSearchParams` usage. Not sure if the right thing to do 😢 
- Made the data row expansions only render when expanded
- more memoising in various key places
- main one: removed the parsing of expiry dates into real dates in the query fn. This was making react-query think the data had changed, so was causing reference instability and re-rendering all of the things

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -34,7 +34,7 @@
     "react-currency-input-field": "^3.6.0",
     "react-hook-form": "^7.22.0",
     "react-i18next": "^11.15.1",
-    "react-query": "^3.34.7",
+    "react-query": "^3.34.12",
     "react-router-dom": "^6.1.1",
     "react-table": "^7.7.0",
     "stylis": "^4.0.13",
--- a/packages/common/src/api/OmSupplyApiContext.tsx
+++ b/packages/common/src/api/OmSupplyApiContext.tsx
@@ -1,4 +1,4 @@
-import React, { FC, useEffect, useState, useCallback } from 'react';
+import React, { FC, useMemo, useEffect, useState, useCallback } from 'react';
 import { createContext } from 'react';
 import { GraphQLClient } from 'graphql-request';
 import { getSdk } from '../types';
@@ -59,9 +59,17 @@ export const OmSupplyApiProvider: FC<ApiProviderProps> = ({
     setApi(createOmSupplyApi(url));
   }, [url]);
 
-  return (
-    <Provider value={{ api, client, setUrl, setHeader }}>{children}</Provider>
+  const val = useMemo(
+    () => ({
+      api,
+      client,
+      setUrl,
+      setHeader,
+    }),
+    [api, client, setUrl, setHeader]
   );
+
+  return <Provider value={val}>{children}</Provider>;
 };
 
 export const OmSupplyApiConsumer = Consumer;
--- a/packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
+++ b/packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
@@ -104,9 +104,7 @@ export const DataRow = <T extends DomainObject>({
             }}
             in={isExpanded}
           >
-            {ExpandContent && isExpanded ? (
-              <ExpandContent rowData={rowData} />
-            ) : null}
+            {ExpandContent ? <ExpandContent rowData={rowData} /> : null}
           </Collapse>
         </td>
       </tr>
--- a/packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
+++ b/packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
@@ -104,7 +104,9 @@ export const DataRow = <T extends DomainObject>({
             }}
             in={isExpanded}
           >
-            {ExpandContent ? <ExpandContent rowData={rowData} /> : null}
+            {ExpandContent && isExpanded ? (
+              <ExpandContent rowData={rowData} />
+            ) : null}
           </Collapse>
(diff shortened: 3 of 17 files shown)


packages/common/package.json
```diff
@@ -34,7 +34,7 @@
     "react-currency-input-field": "^3.6.0",
     "react-hook-form": "^7.22.0",
     "react-i18next": "^11.15.1",
-    "react-query": "^3.34.7",
+    "react-query": "^3.34.12",
```
joshxg, line 37: I bumped this as there was an issue with inlined selector functions. Thought this might be a problem but it wasn't. Kept the bump anyway.

packages/common/src/api/OmSupplyApiContext.tsx
```diff
@@ -59,9 +59,17 @@ export const OmSupplyApiProvider: FC<ApiProviderProps> = ({
...
+    () => ({
+      api,
+      client,
+      setUrl,
+      setHeader,
+    }),
+    [api, client, setUrl, setHeader]
   );
+
+  return <Provider value={val}>{children}</Provider>;
```
joshxg, line 72: Thought this might have been an issue when I was looking around for problems. It wasn't. Might be in the future, though, so kept it!

packages/invoices/src/InboundShipment/DetailView/api.ts
```diff
@@ -188,17 +188,12 @@ export const getInboundShipmentDetailViewApi = (
...
-      const expiryDate = line.expiryDate
-        ? new Date(line.expiryDate)
-        : undefined;
       const location = line.location ? locationGuard(line.location) : undefined;
 
       return {
         ...line,
         stockLine,
         location,
-        expiryDate,
```
joshxg, line 201: as explained, not converting to date objects which aren't json serialisable

packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
```diff
@@ -20,9 +20,12 @@ export const DetailView: FC = () => {
...
 
-  const onRowClick = (item: InvoiceLine | InvoiceItem) => {
-    onOpen(toItem(item));
-  };
+  const onRowClick = useCallback(
+    (item: InvoiceLine | InvoiceItem) => {
+      onOpen(toItem(item));
+    },
+    [toItem, onOpen]
+  );
```
joshxg, line 28: stable reference.. the content area component can re-render and the data table won't so the cost of the re-render in general is cheap

packages/mock-server/src/data/database.ts
```diff
@@ -608,7 +608,7 @@ export const insert = {
       itemUnit: item.unitName ?? '',
       itemId: item.id,
       expiryDate: invoiceLine?.expiryDate
-        ? parse(invoiceLine.expiryDate, 'dd-MM-yyyy', new Date()).toISOString()
+        ? parse(invoiceLine.expiryDate, 'yyyy-MM-dd', new Date()).toISOString()
```
joshxg, line 611: this was the issue I was having with expiry dates not updating, oops!

joshxg (commented):
Small change woo!

packages/invoices/src/InboundShipment/DetailView/columns.ts
```diff
@@ -87,9 +87,17 @@ export const useInboundShipmentColumns = (): Column<
                 'expiryDate',
                 null
               );
-              return formatExpiryDate(expiryDate);
+              const formatted =
```
mark-prins, line 90: a `formatExpiryDateString` perhaps?

packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
```diff
@@ -104,7 +104,9 @@ export const DataRow = <T extends DomainObject>({
             }}
             in={isExpanded}
           >
-            {ExpandContent ? <ExpandContent rowData={rowData} /> : null}
+            {ExpandContent && isExpanded ? (
```
mark-prins, line 107: this doesn't work for me 🤔 The row doesn't expand in the outbound or inbound detail views

mark-prins (approved):
👍  certainly feels snappier!

Have checked the rendering just to confirm what you say about the number of renders. and while sad to see the query params go.. they are a problem. we can't link to a list view with all the parameters, but we can go back now!

I'm having an issue with the row expando, as noted. otherwise functionality all looks good!

packages/invoices/src/InboundShipment/DetailView/columns.ts
joshxg: Yes! 2cb5517 

much better!

packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
joshxg: oops 🙄 

Yup it didn't actually work 😁 

I tried a few things.. adding the `exitOnMount` and `enterOnMount` and `enter` props etc etc but couldn't get it to work. Then I realised that this isn't an issue for what this is fixing as this was essentially fixed by not having the whole table rerender..

So i've reverted the change which makes no difference to the modal performance. Though, it would be best to not render all the mini tables without being expanded

Issue #777, opened 2022-02-08
Outbound TODOs


joshxg:
- [x] Expando component should check there is more than one row.
- [x] Enable deleting of items
- [x] Statuses: #686 
- [x] usePackSizeController tests

PR #779, opened 2022-02-08
#777 Add packSizeController tests

joshxg:
Part of #777 

- Just splitting the PR up
- This is adding tests for the packSizeController and a quick fix also

### DIFF
--- a/packages/invoices/src/OutboundShipment/DetailView/modals/hooks/useDraftOutboundLines.tsx
+++ b/packages/invoices/src/OutboundShipment/DetailView/modals/hooks/useDraftOutboundLines.tsx
@@ -32,11 +32,11 @@ const createPlaceholderRow = (invoiceId: string): DraftOutboundLine => ({
 
 interface DraftOutboundLineSeeds {
   invoiceId: string;
-  stockLine: StockLine;
-  invoiceLine?: InvoiceLine | null;
+  stockLine?: Partial<StockLine> | null;
+  invoiceLine?: Partial<InvoiceLine> | null;
 }
 
-const createDraftOutboundLine = ({
+export const createDraftOutboundLine = ({
   invoiceId,
   stockLine,
   invoiceLine,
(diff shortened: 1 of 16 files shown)


packages/invoices/src/OutboundShipment/DetailView/modals/hooks/useDraftOutboundLines.tsx
```diff
@@ -48,11 +48,19 @@ const createDraftOutboundLine = ({
...
+  stockLineId: stockLine?.id ?? '',
   location: invoiceLine?.location,
   expiryDate: invoiceLine?.expiryDate,
+  costPricePerPack: stockLine?.costPricePerPack ?? 0,
+  sellPricePerPack: stockLine?.sellPricePerPack ?? 0,
+  itemId: stockLine?.itemId ?? '',
+  packSize: stockLine?.packSize ?? 0,
+  onHold: stockLine?.onHold ?? false,
+  storeId: stockLine?.storeId ?? '',
+  totalNumberOfPacks: stockLine?.totalNumberOfPacks ?? 0,
```
joshxg, line 63: I changed this to be optionals so that I could utilise the same fn as a test util 🤔

packages/invoices/src/OutboundShipment/DetailView/modals/hooks/usePackSizeController.tsx
```diff
@@ -38,23 +39,24 @@ const usePackSizeOptions = (
...
+    (newValue: number) => {
+      const packSizeOption = options.find(({ value }) => value === newValue);
+      if (!packSizeOption) return;
+      setSelected(packSizeOption);
+    },
+    [options, setSelected]
+  );
 
   useEffect(() => {
-    if (selected.value !== 0) return;
```
joshxg, line 57: This was messing with the initial value.. the behaviour I think is correct now 🤔

mark-prins: with this removed, I cannot change to `Any` when editing an item which is currently using pack size `1`

The effect should only run on initial load - if the Select is manually changed, this block should not change the selected value back again.. but does!

mark-prins: Originally, I had the default value as `{label: '', value: 0}` and the effect selects `{label: 'label.any', value: -1}`. Now that the default is changed, this line would need to change - or a different way to detect initial state

mark-prins (approved):
👍  tests look good - the pack size select isn't currently working properly though 😢

joshxg:
Ugh thanks!

I didn't click that the effect was running after setting the pack size because the quantity was being changed, which would change the `lines` reference and reset the pack size - that's why I couldn't recreate it in tests..  Oops! 🤔 

So I changed a few things around because I was getting confused, hopefully simplifying the effect and getting the desired behaviour.. 🤞

mark-prins:
found another bug while testing this..
Screen Shot 2022-02-10 at 1 52 44 PM

In this example, I have 4 packs - when I load the modal I'm showing 40 packs of size 'Any' even though all my packs are size 10.

Two issues:
- The label shows `packs` when the value is `units`
- The select is defaulting to `Any` when all the rows are pack size 10

joshxg:
ah nooo 😢

joshxg:
So much for just adding some tests, right?

I see I missed this behaviour of assigning the pack size initially for the pack size that has quantity allocated - of course!  I've added a commit for the fix and a test case to try and capture the behaviour

PR #780, opened 2022-02-09
#777 os todo use next item tests

joshxg:
More on #777 

- I wanted to PR this to visualise the problem a little better.
- Technically, we have a code gen msw handler: `invoiceQueryMock`. However, if I use that, the amount of random fields I need to add is outrageous! So.. I went with this more hacky solution.
- Not 100% sure what the best thing to do is here but I do have some thoughts on what a great end goal could be..

1. Easily be able to attach a handler to a msw server running in test files, without having to think about or handle the restarting of the server etc.
2. Easily be able to generate simple mock objects, not having to specify every field
3. Easily be able to modify values of fields on those mock objects for a single test

I think if we had a way to have those 3 goals covered, it would be super nice and easy to write unit tests for any async work.

Right now, it's tedious and annoying which results in essentially.. not writing them 🤷 

Some extra nice to haves:

1. Allow for whatever data that is generated to be used in some sort of end to end testing suite? For example, maybe using https://github.com/sorry-cypress/sorry-cypress to run tests to insert the data 🤷  I don't really know how this works.
2. The backend team also have mock data. We could work together to create a suite of json files to share the work load? Unsure if this is more troublesome than helpful!


packages/invoices/src/OutboundShipment/DetailView/modals/hooks/useNextItem.tsx
```diff
@@ -9,15 +9,17 @@ export const useNextItem = (
...
+  const nextIdx = currentIdx + 1;
+  const nextItem = items[nextIdx % numberOfItems];
 
-  if (currentIdx === -1 || currentIdx === numberOfItems - 1 || !nextItem) {
+  if (currentIdx === -1 || !nextItem) {
     return { next, disabled };
   }
 
-  return { next: toItem(nextItem), disabled: false };
+  return { next: toItem(nextItem), disabled: nextIdx === numberOfItems - 1 };
```
joshxg, line 24: found a bug 👍

mark-prins: me too! Currently you can only go to the second-to-last item. I suggest:
```suggestion
  return { next: toItem(nextItem), disabled: currentIdx === numberOfItems - 1 };
```

packages/invoices/src/OutboundShipment/DetailView/modals/hooks/useNextItem.test.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+    return result.waitForNextUpdate().then(() => {
+      expect(result.result.current.next).toEqual(
+        expect.objectContaining({ id: invoice.lines.nodes[1]?.id })
+      );
+      expect(result.result.current.disabled).toEqual(false);
+    });
+  });
+
+  it('eventually equals an object where if the next item is the last, disabled is true', async () => {
```
mark-prins, line 140: I think it should be `disabled` if this is the last item. not sure why it should be disabled if the next item is the last one?

mark-prins (approved):
👍  The msw work is very neat! that looks super useful.
I may be confused about intended functionality.. I expected that the `useNext` hook returns disabled if we are on the final item. As it is, I can't `Next` to the last item line, it stops on the previous one 🤷

packages/invoices/src/OutboundShipment/DetailView/modals/hooks/useNextItem.test.tsx
joshxg: uhh.. wow! What a foooool! Fixed ! 😁

Issue #785, opened 2022-02-10
Outbound: Unable to add lines


mark-prins:
When trying to add lines to an invoice - can't. No items are returned.
Looking at the items query, this is sent with a filter of 

```
{
  "key": "name",
  "filter": {
    "code": {
      "equalTo": ""
    }
  },
  "first": 20,
  "offset": 0
}
```

which may explain why

joshxg:
Heh thanks - just was about to offroad this in my current task 😁

mark-prins:
oh! thanks - can probs just delete this issue then 😁

PR #786, opened 2022-02-10
Add correct filtering condition

joshxg:
#785 

The filter was incorrect and was showing only items that were already in the invoice, rather than the reverse.


mark-prins (approved):
🎉  yup! thanks - that's just what I had

PR #790, opened 2022-02-10
Add enabling status changes for outbound shipments

### LINKED ISSUE #777: Outbound TODOs

joshxg:
Fixes #777 

- This enables using the mock server to change statuses
- Also disables on the correct status
- Off road fix for the status text on the popover

### DIFF
--- a/packages/common/src/ui/components/steppers/VerticalStepper/VerticalStepper.tsx
+++ b/packages/common/src/ui/components/steppers/VerticalStepper/VerticalStepper.tsx
@@ -136,10 +136,10 @@ export const VerticalStepper: FC<StepperProps> = ({ activeStep, steps }) => (
                 justifyContent="space-between"
               >
                 <Typography
+                  color="text.primary"
                   variant="body2"
                   lineHeight={0}
                   fontSize="12px"
-                  fontWeight="bold"
                 >
                   {step.label}
                 </Typography>
--- a/packages/invoices/src/OutboundShipment/DetailView/Footer.tsx
+++ b/packages/invoices/src/OutboundShipment/DetailView/Footer.tsx
@@ -116,8 +116,8 @@ export const Footer: FC = () => {
                   variant="contained"
                   color="secondary"
                   onClick={async () => {
-                    success('Saved invoice! 🥳 ')();
                     await update({ status: getNextOutboundStatus(status) });
+                    success('Saved invoice! 🥳 ')();
                   }}
                 />
               )}
--- a/packages/invoices/src/OutboundShipment/api/hooks.ts
+++ b/packages/invoices/src/OutboundShipment/api/hooks.ts
@@ -53,6 +53,7 @@ export const useOutboundFields = <KeyOfInvoice extends keyof Invoice>(
 export const useIsOutboundDisabled = (): boolean => {
   const { status } = useOutboundFields('status');
   return (
+    status === InvoiceNodeStatus.Shipped ||
     status === InvoiceNodeStatus.Verified ||
     status === InvoiceNodeStatus.Delivered
   );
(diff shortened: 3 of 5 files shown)


packages/common/src/ui/components/steppers/VerticalStepper/VerticalStepper.tsx
```diff
@@ -136,10 +136,10 @@ export const VerticalStepper: FC<StepperProps> = ({ activeStep, steps }) => (
                 justifyContent="space-between"
               >
                 <Typography
+                  color="text.primary"
```
joshxg, line 139: was white 🤔

packages/invoices/src/OutboundShipment/DetailView/Footer.tsx
```diff
@@ -116,8 +116,8 @@ export const Footer: FC = () => {
                   variant="contained"
                   color="secondary"
                   onClick={async () => {
-                    success('Saved invoice! 🥳 ')();
                     await update({ status: getNextOutboundStatus(status) });
+                    success('Saved invoice! 🥳 ')();
```
joshxg, line 120: showing after it's saved. Not sure what to *actually* display?

packages/invoices/src/OutboundShipment/api/hooks.ts
```diff
@@ -53,6 +53,7 @@ export const useOutboundFields = <KeyOfInvoice extends keyof Invoice>(
 export const useIsOutboundDisabled = (): boolean => {
   const { status } = useOutboundFields('status');
   return (
+    status === InvoiceNodeStatus.Shipped ||
```
joshxg, line 56: should be disabled for shipped also?

packages/mock-server/src/api/mutations/invoice.ts
```diff
@@ -49,7 +49,16 @@ export const InvoiceMutation = {
...
+          InvoiceNodeType.OutboundShipment,
+          {
+            status: InvoiceNodeStatus.New,
+            allocatedDatetime: null,
+            pickedDatetime: null,
+            shippedDatetime: null,
+            deliveredDatetime: null,
+            verifiedDatetime: null,
+            comment: null,
+          }
```
joshxg, line 61: This is for creating an outbound. It is utilising the same `createInvoice` as the random data generator, so nulling out all the fields from random values.

joshxg (commented):
I wanted to change the footer based on the shipment being an adhoc invoice or not, but I don't think the field has made it into the api yet

packages/invoices/src/OutboundShipment/api/hooks.ts
mark-prins: yeah, I would think so.
Also - the inbound version is `useIsInboundEditable` and one of them should change to be consistent

packages/invoices/src/OutboundShipment/DetailView/Footer.tsx
joshxg: Haha, yes. I suppose my question was more in regard to if there should even be a notification. I originally added it just for a visual indicator while developing. Though, if there is a confirmation modal before status changes (requested by craig) .. then maybe there is a better solution (i.e. should the confirmation modal just stay open and show a loading indicator..?)

packages/invoices/src/OutboundShipment/api/hooks.ts
joshxg: Yeah you're right

packages/invoices/src/OutboundShipment/DetailView/Footer.tsx
mark-prins: oh, right. if there's a confirmation modal then that could stay open and spin, like you say. then close if the save is ok, or show an error if not? That ties in with the other issue of showing a blocking modal to prevent changes while updating status

PR #796, opened 2022-02-11
#788 localisation file format

### LINKED ISSUE #788: Change localisation file format
I would like to support plurals in the localisation, specifically for units in the outbound shipment edit modal.
This means changing from the flat json structure to the JSON v4 format

mark-prins:
Fixes #788 

I wanted to use the standard format 
```
{
    "key": {
        "message": "blah",
        "description": "blah"
    }
...
}
```

but, now that I look - it isn't supported by i18next, so have just added an example of a plural and battered the `LocaleKey` to allow typing of the locale keys to work


mark-prins:
oh. actually - this needs another PR merged to main - when that is done I'll change the base of this PR

joshxg (approved):
Nice! 

I agree the format you have in your comment seems easier to understand, but this works well and is easy!

Issue #789, opened 2022-02-10
Outbound: Extract FEFO logic


mark-prins:
Would like to lift out the FEFO allocation logic and wrap it up in some nice tests

PR #797, opened 2022-02-11
check expiry

### LINKED ISSUE #652: Don't FEFO allocate to expired batches?
Something in a meeting the other day about mSupply auto allocating to expired batches and that being a no-no.  At the moment we allocate FEFO, but we're not completely disregarding expired batches. Maybe we should be?

mark-prins:
Fixes #652 

I checked - I'd assumed that the API wouldn't provided expired batches.. but it does. So I've prevented FEFO allocation for those. Probably good to merge this before doing #789 - or just pulling the changes in, there aren't many.


packages/invoices/src/OutboundShipment/DetailView/modals/hooks/useDraftOutboundLines.tsx
```diff
@@ -52,7 +52,7 @@ const createDraftOutboundLine = ({
   invoiceId,
   stockLineId: stockLine.id,
   location: invoiceLine?.location,
-  expiryDate: invoiceLine?.expiryDate,
+  expiryDate: invoiceLine?.expiryDate || stockLine?.expiryDate,
```
mark-prins, line 55: `expiryDate` was only being shown for invoice lines.. which wasn't very helpful 😉

joshxg (approved):
Yes! thanks.. I'd forgotten 🙄

PR #809, opened 2022-02-11
#686 allocate quantities tests

### LINKED ISSUE #686: Invoice statuses
The current statuses  mechanism isn't really making sense.

This is where I would usually put a suggestion as to what to do here. However, the statuses in general don't make sense to me.. so for now this is a placeholder 😁

joshxg:
Fixes #686 

- Resisted the urge to refactor. The allocateQuantities function in general seems a little confusing and some breaking up into smaller functions would likely be beneficial for my reading of the code. Have left it for now though. It's working well and with tests in place it will hopefully be easy to change when needed - if you think that time is now, I'm happy to spend a bit of time on it.
- Added config changes for something i've been meaning to figure out for awhile. This makes the source maps correct when debugging / using breakpoints for jest tests. This means that you can put a break point in vscode and use the `testing` side bar tab to quickly run and breakpoint on code. Wasn't sure if I should commit my launch.json - I thought not as it's specific to vscode.
- Added tests which cover at least most of the behaviours of the function

### DIFF
--- a/.swcrc
+++ b/.swcrc
@@ -8,6 +8,7 @@
       "decorators": true
     }
   },
+  "sourceMaps": "inline",
   "module": {
     "type": "commonjs"
   }
--- a/packages/invoices/src/OutboundShipment/DetailView/modals/OutboundLineEdit.tsx
+++ b/packages/invoices/src/OutboundShipment/DetailView/modals/OutboundLineEdit.tsx
@@ -62,11 +62,13 @@ export const OutboundLineEdit: React.FC<ItemDetailsModalProps> = ({
     return true;
   };
 
-  const onAllocate = allocateQuantities(
-    status,
-    draftOutboundLines,
-    setDraftOutboundLines
-  );
+  const onAllocate = (newVal: number, packSize: number | null) => {
+    const newAllocateQuantities = allocateQuantities(
+      status,
+      draftOutboundLines
+    )(newVal, packSize);
+    setDraftOutboundLines(newAllocateQuantities ?? draftOutboundLines);
+  };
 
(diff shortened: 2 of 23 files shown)


packages/invoices/src/OutboundShipment/DetailView/modals/OutboundLineEdit.tsx
```diff
@@ -62,11 +62,13 @@ export const OutboundLineEdit: React.FC<ItemDetailsModalProps> = ({
...
-    draftOutboundLines,
-    setDraftOutboundLines
-  );
+  const onAllocate = (newVal: number, packSize: number | null) => {
+    const newAllocateQuantities = allocateQuantities(
+      status,
+      draftOutboundLines
+    )(newVal, packSize);
+    setDraftOutboundLines(newAllocateQuantities ?? draftOutboundLines);
+  };
```
joshxg, line 71: I made the `allocateQuantities` fn just return a new array rather than pass in the setState fn to make testing easier

joshxg:
Hm wrong branch name 🤔

.swcrc
```diff
@@ -8,6 +8,7 @@
       "decorators": true
     }
   },
+  "sourceMaps": "inline",
```
mark-prins, line 11: ah. thanks! was going to look into that too..

mark-prins (approved):
🎉  woo! that's great, thanks!

mark-prins:
I think you're right with the refactor - good to do, and it would be great to have your hand in it. best not to do it right now, and wait till more of the v1 work is done

PR #812, opened 2022-02-11
#811 Add pack size controller bug fix

### LINKED ISSUE #811: Stale state with Pack size controller and `ok & next`
This is a test case that fails:

```
  it('resets the selected if the set of lines changes to a set with a different itemID', async () => {
    const { result, rerender } = renderHook(
      ({ lines }) => usePackSizeController(lines),
      {
        initialProps: { lines: singlePackSizeLines },
      }
    );

    expect(result.current.selected).toEqual({
      label: '1',
      value: 1,
    });

    rerender({ lines: multiplePackSizeLines });

    expect(result.current.selected).toEqual({ label: 'label.any', value: -1 });
  });
```

Where the mutiple pack size lines should be setting the pack size to `any` when re-rendering with that set of lines, instead the state is not changing.

joshxg:
#811 

- resets the selected state when the item changes
- Also adds some styling changes in an attempt to stop the janky height changes to the modal when clicking ok next

### DIFF
--- a/packages/invoices/src/OutboundShipment/DetailView/modals/OutboundLineEditTable.tsx
+++ b/packages/invoices/src/OutboundShipment/DetailView/modals/OutboundLineEditTable.tsx
@@ -14,6 +14,7 @@ import {
   useTranslation,
   ReadOnlyInput,
   InvoiceNodeStatus,
+  Box,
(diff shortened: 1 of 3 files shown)


mark-prins (approved):
🎉  very nice

PR #820, opened 2022-02-15
#813 skipping outbound statuses

### LINKED ISSUE #813: Allow 'skipping' status levels when changing status
It's been requested that a user be able to 'skip' status levels. For example, allow a user to go from `new` to `verified` in one press, rather than having to send multiple requests.

joshxg:
Fixes #813 

- Not really sure if the UX for this is the best, but thought to leave it here and start getting feedback

image

### DIFF
--- a/packages/common/src/ui/components/buttons/ButtonShowcase.stories.tsx
+++ b/packages/common/src/ui/components/buttons/ButtonShowcase.stories.tsx
@@ -62,7 +62,7 @@ const Template: Story<{ color: 'primary' | 'secondary' }> = ({ color }) => {
   return (
     <>
       <Grid container gap={2}>
-        <Wrapper text="Base Button: Outlined variant, primary color">
+        <Wrapper text="Base Button: Outlined variant">
           <BaseButton
             variant="outlined"
             color={color}
@@ -72,7 +72,7 @@ const Template: Story<{ color: 'primary' | 'secondary' }> = ({ color }) => {
           </BaseButton>
         </Wrapper>
 
-        <Wrapper text="Base Button: Contained variant, primary color">
+        <Wrapper text="Base Button: Contained variant">
           <BaseButton
             variant="contained"
             color={color}
@@ -82,7 +82,7 @@ const Template: Story<{ color: 'primary' | 'secondary' }> = ({ color }) => {
           </BaseButton>
         </Wrapper>
 
-        <Wrapper text="Button with Icon, contained & primary">
+        <Wrapper text="Button with Icon, contained">
           <ButtonWithIcon
             variant="contained"
             color={color}
@@ -92,7 +92,7 @@ const Template: Story<{ color: 'primary' | 'secondary' }> = ({ color }) => {
           />
         </Wrapper>
 
-        <Wrapper text="Button with Icon, outlined & primary">
+        <Wrapper text="Button with Icon, outlined">
           <ButtonWithIcon
             variant="outlined"
             color={color}
--- a/packages/common/src/ui/components/buttons/ButtonShowcase.stories.tsx
+++ b/packages/common/src/ui/components/buttons/ButtonShowcase.stories.tsx
@@ -11,6 +11,17 @@ import { ToggleButton } from './ToggleButton';
 import { ColorSelectButton } from './ColorSelectButton';
 import { useTranslation } from '@common/intl';
 import { StoryProvider } from '../../../utils/testing';
+import { SplitButtonOption } from '@common/components';
+
+const ops: [
+  SplitButtonOption<string>,
+  SplitButtonOption<string>,
+  SplitButtonOption<string>
+] = [
+  { label: 'Create a merge commit', value: 'createAndMerge' },
+  { label: 'Squash and merge', value: 'squashAndMerge' },
+  { label: 'Rebase and merge', value: 'rebaseAndMerge' },
+];
 
 const getOnClick = (someText: string) => () => {
   alert(someText);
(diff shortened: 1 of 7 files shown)


packages/common/src/ui/components/buttons/ButtonShowcase.stories.tsx
```diff
@@ -10,7 +10,17 @@ import { Color } from '../menus';
 import { ToggleButton } from './ToggleButton';
 import { ColorSelectButton } from './ColorSelectButton';
 import { useTranslation } from '@common/intl';
-import { StoryProvider } from '../../../utils/testing';
+import { SplitButtonOption } from '@common/components';
+
```
joshxg, line 14: the change in here is to account for now having to pass in the state. Also changed up the primary/secondary to pass a param for the colors for the buttons - though not all buttons support the color changing

packages/common/src/ui/components/buttons/SplitButton.stories.tsx
```diff
@@ -1,28 +1,85 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
```
joshxg, line 1: again change here is to support the change to the internal state. Additionally added the disabled showcase and RTL support

packages/common/src/ui/components/buttons/SplitButton.tsx
```diff
@@ -6,109 +6,105 @@ import { ChevronDownIcon } from '../../icons';
...
-}: SplitButtonProps) => {
+  selectedOption,
+  onSelectOption,
+}: SplitButtonProps<T>) => {
+  console.log(selectedOption);
   const [anchorEl, setAnchorEl] = React.useState<null | HTMLElement>(null);
-  const [selectedIndex, setSelectedIndex] = React.useState(0);
-
-  const buttonLabel = options[selectedIndex]?.label ?? '';
+  const buttonLabel = selectedOption.label;
```
joshxg, line 39: the state change here is not ideal. I was thinking to create a wrapper `StatefulSplitButton` so there is still the option of using the more simple implementation.. 🙄  

Material UI usually allows for their components to be either - i.e. autocomplete - if you pass `undefined` for `value` .. it's a bit of work to implement but not too bad, so could also go that route, though personally I feel like having two explicit components is easier

packages/invoices/src/OutboundShipment/DetailView/Footer/StatusChangeButton.tsx
```diff
@@ -0,0 +1,122 @@
+import React, { useState, useEffect } from 'react';
```
joshxg, line 1: don't particularly like how this turned out.. other ways I tried with arrays of statuses and indexes to try and make it more dry just seemed to be more confusing than helpful

packages/common/src/ui/components/buttons/ButtonShowcase.stories.tsx
```diff
@@ -72,17 +75,7 @@ const Template: Story = () => {
...
-          >
-            Base Button
-          </BaseButton>
-        </Wrapper>
-
-        <Wrapper text="Base Button: Contained variant, secondary color">
-          <BaseButton
-            variant="contained"
-            color="secondary"
+            color={color}
```
mark-prins, line 78: Should probably change the description for the buttons, to remove `primary` now that they aren't hard coded to the primary colour

packages/common/src/ui/components/buttons/SplitButton.stories.tsx
```diff
@@ -1,28 +1,85 @@
...
+
+  return (
+    <Box>
+      <SplitButton
+        ariaLabel="Split button"
+        ariaControlLabel="open split button menu"
+        options={ops}
+        onClick={option => alert(JSON.stringify(option))}
+        selectedOption={selectedOption}
+        onSelectOption={setSelectedOption}
```
mark-prins, line 50: um.. this one needs the `isDisabled` prop 🙂

packages/common/src/ui/components/buttons/SplitButton.stories.tsx
```diff
@@ -1,28 +1,85 @@
...
+      />
+    </Box>
+  );
+};
+
+const RTLComponent = () => {
```
mark-prins, line 56: very nice! looks good RTL too

packages/invoices/src/OutboundShipment/DetailView/Footer/StatusChangeButton.tsx
mark-prins: I can feel the temptation to make a tricksy array based method - and expect that it would be much harder to understand. what you've got is clear and easy to follow; it's not too big so I think it's great like it is

mark-prins (approved):
👍  that's working really nicely - and I think the design is fine as it is

packages/common/src/ui/components/buttons/SplitButton.stories.tsx
joshxg: One of the options was disabled 😁  Have added an option disabled story and added the is disabled prop to this one [74cd53f](https://github.com/openmsupply/openmsupply-client/pull/820/commits/74cd53f26f7a521bdbb475e8883d122da13fcf0a)

mark-prins: I thought that might be it.. but the first story also had a disabled option and I wasn't sure what the difference was supposed to be! thanks - I think I get it now :)

Issue #792, opened 2022-02-10
Status changes should have a confirmation dialog


joshxg:
Has been requested that a confirmation dialog is shown before statuses are changed.

I think this is likely a common thing. So.. generic?!

Thinking a hook concept similar to `useNotification` where we can pass a message and callback for the `ok`.

An api I think would be good would be something like..

```
const { onGetConfirmation } = useConfirmation({
    message: "This will change the status to Verified. Are you sure?",
    onConfirm: () => // do the update
})
```

1. the `onGetConfirmation` naming is poor
2. The message could have some nicer dx. passing a label, or not having to put "Are you sure?" for example.
3. I think an "Are you sure?" confirmation is pretty standard. Are there other variants of this type of modal? The `useNofications` is awesome because you can choose a variant but have the same api, so if there was different variants of this problem, we could make this hook have multiple variants also. Maybe too far.

An alternative api could be something promise based. I think this might be a little harder to implement. Though I think the api is quite nice to use 🤔 

```
const { open } = useConfirmationModal({
    message: "This will do things...."
})

...

onClick={async () => {
    const result = await open();
    if (result.ok) // update
}}

```

PR #822, opened 2022-02-15
Add ConfirmationModal component

joshxg:
Starts #792 

- Adds a confirmation modal component
image
- Handles the confirmation callback being an async func or returning a promise also
- Next step I think is to put this component at the root level, likely with a context provider and make a hook `useConfirmationModal` to be able to open the modal from anywhere.

### DIFF
--- a/packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModal.tsx
+++ b/packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModal.tsx
@@ -1,0 +1,78 @@
+import React, { useState } from 'react';
+import { Grid, Typography } from '@mui/material';
+import { BasicModal } from '../BasicModal';
+import { AlertIcon } from '@common/icons';
+import { DialogButton, LoadingButton } from '../../buttons';
+
+interface ConfirmationModalProps {
+  open: boolean;
+  width?: number;
+  height?: number;
+  onConfirm: (() => void) | (() => Promise<void>);
+  onCancel: () => void;
+  title: string;
+  message: string;
+  Icon?: React.ReactNode;
+}
+
+export const ConfirmationModal = ({
+  open,
+  width = 400,
+  height = 200,
+  onConfirm,
+  title,
+  message,
+  onCancel,
+  Icon = <AlertIcon color="primary" />,
+}: ConfirmationModalProps) => {
+  const [loading, setLoading] = useState(false);
+
+  return (
+    <BasicModal width={width} height={height} open={open}>
+      <Grid container gap={1} flex={1} padding={4} flexDirection="column">
+        <Grid container gap={1} flexDirection="row">
+          <Grid item>{Icon}</Grid>
+          <Grid item>
+            <Typography variant="h6">{title}</Typography>
+          </Grid>
+        </Grid>
+        <Grid item>
+          <Typography>{message}</Typography>
+        </Grid>
+        <Grid
+          container
+          gap={1}
+          flexDirection="row"
+          alignItems="flex-end"
+          justifyContent="flex-end"
+          flex={1}
+          display="flex"
+        >
+          <Grid item>
+            <DialogButton
+              variant="cancel"
+              disabled={loading}
+              onClick={onCancel}
+            />
+          </Grid>
+          <Grid item>
+            <LoadingButton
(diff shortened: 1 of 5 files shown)


packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -0,0 +1,78 @@
...
+  onCancel: () => void;
+  title: string;
+  message: string;
+  Icon?: React.ReactNode;
+}
+
+export const ConfirmationModal = ({
+  open,
+  width = 400,
+  height = 200,
```
joshxg, line 21: not really sure if this is a good idea. I suspect not

packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -0,0 +1,78 @@
...
+              onClick={async () => {
+                const result = onConfirm();
+                if (result instanceof Promise) {
+                  setLoading(true);
+                  await result;
+                  setLoading(false);
+                }
+              }}
+            >
+              OK
```
joshxg, line 71: Hm. Not sure 😁  The `DialogButton` uses `useTranslation` for the `OK`, but I think we want to move away from that. So maybe button labels as props? This is potentially better anyway .. might want to put `save` instead of `ok`, sometimes?

packages/common/src/ui/notifications/AlertModal.tsx
```diff
@@ -17,55 +18,29 @@ export const AlertModal: React.FC<AlertModalProps> = ({
...
-                </Typography>
-              </Grid>
-            </Grid>
-            <Grid item>
-              <Typography>{message}</Typography>
-            </Grid>
-            <Grid item display="flex" justifyContent="flex-end" flex={1}>
-              <DialogButton variant="ok" onClick={onOk} />
-            </Grid>
+    <BasicModal open={open} width={400} height={150}>
```
joshxg, line 21: uhh I removed the props and then when I did the confirmation modal I used the `BasicModal` which seems to do what we need.. looks the same to me.. but it likely isn't and I can revert

packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModal.tsx
mark-prins: it does need something for the dimensions..

packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+export const ConfirmationModal = ({
+  open,
+  width = 400,
+  height = 200,
+  onConfirm,
+  title,
+  message,
+  onCancel,
+  Icon = <AlertIcon color="primary" />,
```
mark-prins, line 26: wonder if the icon should be a prop; or an enum for alert, info, error 🤔

packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModal.tsx
mark-prins: would be better as a prop - ideally defaulting to OK but still translatable, though not sure how to pull that off. it would be adding `useTranslation` in the common component and I don't think that's a great idea. maybe it's ok?

could be an enum of possible options? couldn't be that many.. surely?

packages/common/src/ui/notifications/AlertModal.tsx
mark-prins: looks fine to me!

mark-prins (approved):
🎉  that looks great!

packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModal.tsx
joshxg: that's a good idea!

joshxg:
Added an icon type prop for `alert` and `info` - I thought `error` might be the other `AlertModal`? Also wasn't sure what to use for an alert icon 😁

PR #823, opened 2022-02-15
#792 status confirmation

### LINKED ISSUE #792: Status changes should have a confirmation dialog

joshxg:
Fixes #792 

- Adds a provider with `ConfirmationModal` and a hook `useConfirmationModal`
- Adds use of the above hook in the status change area of the outbound shipment detail view
- Unsure of the copy but .. 🤷

### DIFF
--- a/packages/common/src/intl/locales/en/distribution.json
+++ b/packages/common/src/intl/locales/en/distribution.json
@@ -32,5 +32,8 @@
   "message.deleted-requisitions": "Deleted {{number}} requisitions",
   "message.select-rows-to-delete": "Select rows to delete",
   "messages.click-to-return-to-requisitions": "Unable to find a requisition with that ID. Click OK to return to the requisition list",
-  "message.deleted-lines": "Deleted {{number}} lines"
+  "message.deleted-lines": "Deleted {{number}} lines",
+  "message.confirm-status-as": "Confirm status as $t({{status}})?",
+  "message.shipment-saved": "Shipment saved 🥳",
+  "message.error-saving-shipment": "Error saving shipment 🥺"
 }
--- a/packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModal.stories.tsx
+++ b/packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModal.stories.tsx
@@ -4,6 +4,7 @@ import { Story } from '@storybook/react';
 import { ConfirmationModal } from './ConfirmationModal';
 import { BaseButton } from '../../buttons';
 import { useToggle } from '../../../../hooks';
+import { useConfirmationModal } from '@common/components';
 
 export default {
   title: 'Modals/ConfirmationModal',
@@ -51,5 +52,16 @@ const Loading: Story = () => {
   );
 };
 
+const UseConfirmationModalHook: Story = () => {
+  const getConfirmation = useConfirmationModal({
+    title: 'Are you sure?',
+    message: 'This will delete all your data.',
+    onConfirm: () => alert('confirmed!'),
+  });
+
+  return <BaseButton onClick={getConfirmation}>Open Modal</BaseButton>;
+};
+
 export const Primary = Template.bind({});
 export const WithAsyncConfirmation = Loading.bind({});
+export const WithContextAndHook = UseConfirmationModalHook.bind({});
--- a/packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModal.stories.tsx
+++ b/packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModal.stories.tsx
@@ -4,7 +4,7 @@ import { Story } from '@storybook/react';
 import { ConfirmationModal } from './ConfirmationModal';
 import { BaseButton } from '../../buttons';
 import { useToggle } from '../../../../hooks';
-import { useConfirmationModal } from '@common/components';
+import { useConfirmationModal } from './ConfirmationModalProvider';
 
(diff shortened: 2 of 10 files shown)


packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModal.stories.tsx
```diff
@@ -51,5 +52,16 @@ const Loading: Story = () => {
...
 };
 
+const UseConfirmationModalHook: Story = () => {
+  const getConfirmation = useConfirmationModal({
+    title: 'Are you sure?',
+    message: 'This will delete all your data.',
+    onConfirm: () => alert('confirmed!'),
+  });
+
+  return <BaseButton onClick={getConfirmation}>Open Modal</BaseButton>;
```
joshxg, line 62: should I also put a story in the `hooks` section 🤔

packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModalProvider.tsx
```diff
@@ -0,0 +1,104 @@
...
+      setState,
+      ...confirmationModalState,
+    }),
+    [setState, confirmationModalState]
+  );
+
+  return (
+    <Context.Provider value={confirmationModalController}>
+      {children}
+      <ConfirmationModal
```
joshxg, line 67: technically this is hitting against the general context problem of state changing this far up the tree. it doesn't seem to have a performance hit - profiling doesn't show anyheavy renders, all below 10ms so it seems fine..

packages/invoices/src/OutboundShipment/DetailView/Footer/StatusChangeButton.tsx
```diff
@@ -72,35 +73,73 @@ const getStatusOptions = (
   return options;
 };
 
-export const StatusChangeButton = () => {
-  const { success } = useNotification();
-  const isDisabled = useIsOutboundDisabled();
-  const { status, update } = useOutboundFields('status');
-  const t = useTranslation('distribution');
+const getNextStatusOption = (
```
joshxg, line 76: I basically took all the things in the component and extracted it to the `useStatusChangeButton` because that makes me feel better.... then just added the call to `useConfirmationModal` and it works pretty well!

packages/common/src/intl/locales/en/distribution.json
```diff
@@ -32,5 +32,8 @@
   "message.deleted-requisitions": "Deleted {{number}} requisitions",
   "message.select-rows-to-delete": "Select rows to delete",
   "messages.click-to-return-to-requisitions": "Unable to find a requisition with that ID. Click OK to return to the requisition list",
-  "message.deleted-lines": "Deleted {{number}} lines"
+  "message.deleted-lines": "Deleted {{number}} lines",
+  "message.confirm-status-as": "Confirm status as $t({{status}})?",
+  "message.shipment-saved": "Shipment saved 🥳",
+  "message.error-saving-shipment": "Error saving shipment 🥺"
```
mark-prins, line 38: I do like the emoji!

packages/common/src/intl/locales/en/distribution.json
```diff
@@ -32,5 +32,8 @@
   "message.deleted-requisitions": "Deleted {{number}} requisitions",
   "message.select-rows-to-delete": "Select rows to delete",
   "messages.click-to-return-to-requisitions": "Unable to find a requisition with that ID. Click OK to return to the requisition list",
-  "message.deleted-lines": "Deleted {{number}} lines"
+  "message.deleted-lines": "Deleted {{number}} lines",
+  "message.confirm-status-as": "Confirm status as $t({{status}})?",
+  "message.shipment-saved": "Shipment saved 🥳",
```
mark-prins, line 37: With the modal and the loading spinner this could work well as just 'Success' or some such.
As you've already said.. I don't think we need the `Shipment saved` notification so much now.

packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModal.stories.tsx
mark-prins: not essential, but it is a hook - so probably yes.

mark-prins (approved):
👍  that is very elegant! the implementation of the hook and provider is very neat

If I can make a minor request.. could we have the info icon in blue? - the secondary colour would be good. The orange is a little startling

mark-prins (approved):
🎉  awesome!

PR #824, opened 2022-02-16
Add status change confirmation

### LINKED ISSUE #792: Status changes should have a confirmation dialog

joshxg:
Fixes #792 

- This implements the confirmation for inbound shipments confirmation modal

### DIFF
--- a/packages/invoices/src/InboundShipment/DetailView/Footer/StatusChangeButton.tsx
+++ b/packages/invoices/src/InboundShipment/DetailView/Footer/StatusChangeButton.tsx
@@ -1,0 +1,148 @@
+import React, { useMemo, useState, useEffect } from 'react';
+import {
+  ArrowRightIcon,
+  useTranslation,
+  useNotification,
+  InvoiceNodeStatus,
+  SplitButton,
+  SplitButtonOption,
+  useConfirmationModal,
+} from '@openmsupply-client/common';
+import { getStatusTranslation, getNextInboundStatus } from '../../../utils';
+import { useIsInboundEditable, useInboundFields } from '../api';
+
+const getStatusOptions = (
+  currentStatus: InvoiceNodeStatus,
+  getButtonLabel: (status: InvoiceNodeStatus) => string
+): SplitButtonOption<InvoiceNodeStatus>[] => {
+  const options: [
+    SplitButtonOption<InvoiceNodeStatus>,
+    SplitButtonOption<InvoiceNodeStatus>,
+    SplitButtonOption<InvoiceNodeStatus>,
+    SplitButtonOption<InvoiceNodeStatus>,
+    SplitButtonOption<InvoiceNodeStatus>
+  ] = [
+    {
+      value: InvoiceNodeStatus.New,
+      label: getButtonLabel(InvoiceNodeStatus.New),
+      isDisabled: true,
+    },
+    {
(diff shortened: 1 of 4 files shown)


joshxg (commented):
I've not had a chance to test this on the real api yet as there are a lot of changes to make after merges which i'll look at next

packages/invoices/src/InboundShipment/DetailView/Footer/StatusChangeButton.tsx
```diff
@@ -0,0 +1,148 @@
...
+  }, [status, options]);
+
+  return { options, selectedOption, setSelectedOption, onGetConfirmation };
+};
+
+export const StatusChangeButton = () => {
+  const { options, selectedOption, setSelectedOption, onGetConfirmation } =
+    useStatusChangeButton();
+  const isDisabled = !useIsInboundEditable();
+  console.log(isDisabled, selectedOption);
```
mark-prins, line 134: debug code?

packages/invoices/src/InboundShipment/DetailView/Footer/StatusChangeButton.tsx
```diff
@@ -0,0 +1,148 @@
...
+  if (currentStatus === InvoiceNodeStatus.New) {
+    // When the status is new, delivered and verified are available to
+    // select.
+    options[3].isDisabled = false;
+    options[4].isDisabled = false;
+  }
+
+  // When the status is delivered, only verified is available to select.
+  if (currentStatus === InvoiceNodeStatus.Delivered) {
+    options[3].isDisabled = false;
```
mark-prins, line 61:
```suggestion
    options[4].isDisabled = false;
```

mark-prins (approved):
🎉  awesome!

packages/invoices/src/InboundShipment/DetailView/Footer/StatusChangeButton.tsx
joshxg: 😁

packages/invoices/src/InboundShipment/DetailView/Footer/StatusChangeButton.tsx
joshxg: argh! thanks

Issue #825, opened 2022-02-16
API integration is broken


joshxg:
There are a few changes which are breaking the API integration in places that are a little unexpected. Specifically, the invoice/invoices queries aren't working because the field `color` has been renamed `colour`. 

Just changing the query isn't quite enough - would want to run the code gen again. But that is having issues as there are now duplicates in the `additions` schema.

First step for getting this all integrated nicely again is to be a bit hacky and just get things running!

PR #826, opened 2022-02-16
Add hack updates to get things working :grin

joshxg:
#825 

- There's a lot of work here
- I by-passed the husky check, with this branch `yarn compile` doesn't work
- I believe *most* of the errors are isolated to the requisition and stocktake mock server and api modules, though there will likely be a fair amount of other changes - there are a lot!


joshxg:
This is getting a bit tricky. With the addition of `stats` to `item`, the `ItemNode` is a little more complex - though, now it doesn't return an error which is nice. 

Requisition is a bit different so haven't really handled that - same with stocktake!

packages/mock-server/src/worker/handlers/invoice/index.ts
```diff
@@ -306,35 +344,36 @@ const upsertOutboundShipmentMutation = mockUpsertOutboundShipmentMutation(
     // Then, the graphql code gen types generally have `__typename` as an optional field as it's
     // not always queried for - but in the types returned by queries, when you specify the type name,
     // it becomes mandatory to have this in the response, so we have to manually add it.
+    console.log('req.variables.input', req.variables);
```
mark-prins, line 347: mark as to be removed later?

packages/mock-server/src/worker/handlers/invoice/index.ts
```diff
@@ -199,102 +204,135 @@ const insertOutboundShipmentMutation = mockInsertOutboundShipmentMutation(
...
+//           }[];
+//           updateInboundShipmentLines: {
+//             __typename: 'UpdateInboundShipmentLineResponseWithId';
+//             id: string;
+//           }[];
+//         };
+//       }) => any;
+//     }
+//   ) => {
+//     // This whole thing is a little unfortunate.
```
mark-prins, line 242: yeah - ugh.

packages/invoices/src/OutboundShipment/DetailView/api.ts
```diff
@@ -1,258 +0,0 @@
-import {
```
mark-prins, line 1: 🤔  moved elsewhere?

mark-prins (approved):
wow. so many changes. thanks!

packages/common/src/operations.graphql
```diff
@@ -402,152 +395,154 @@ mutation insertStocktake($input: InsertStocktakeInput!) {
...
-          }
-        }
         ... on RequisitionLineConnector {
           totalCount
           nodes {
             id
-            itemName
-            itemCode
-            itemUnit
+            # itemName
```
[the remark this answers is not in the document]
joshxg: Yeah.

I think we're really going to need to look at fragments now. I.e. a fragment `BasicItemNode`

```
fragment Item on ItemNode {
    id
    name
    code
  }
```

for example this will generate a new type - I believe `ItemFragment` is the default name - then I believe we should be able to make all of this a lot easier in some ways

packages/invoices/src/OutboundShipment/DetailView/api.ts
joshxg: Ah yes this was moved previously and forgot to remove this

PR #828, opened 2022-02-16
#825 remove mock server

joshxg:
Removes the mock server 🤷 

- Just removed it completely and the couple of tests which were using msw
- Have left the codegen for msw etc as this will still be useful for returning static data in unit tests


joshxg:
Wow that's a lot of code!

joshxg:
Wow that's a massive bundle size differnence 🤣

mark-prins (approved):
😢  bye bye mock server

shame to see it go - there's a lot of good work in there.
it does make the bundle smaller though..

PR #829, opened 2022-02-17
#631 split gql files

### LINKED ISSUE #631: Split GQL operations file
The GQL Operations file is a bit too big and annoying to work with.

You can generate using a glob pattern `**/*.graphql` which means we could have a .. `queries`? `operations`? file next to the `api` file in each of the sub packages?

joshxg:
Fixes #631 

- kind of
- I don't actually really know what I'm doing, and this is more of an example / just the config while trying to leave everything in place so that we can migrate a little easier.
- Idea being, that next to or in the api modules we current have for each of the different features, we can create an `operations` file and run the code gen for each of the files.
- This also means we would be able to write fragments in each of the operation files that we can use in the modules also


codegen.yml
```diff
@@ -1,10 +1,21 @@
 overwrite: true
 schema: ["https://demo-open.msupply.org:8000/graphql", "./packages/common/src/additions.schema.graphql"]
-documents: './packages/common/src/operations.graphql'
 generates:
   ./packages/common/src/types/schema.ts:
     plugins:
       - typescript
+    config:
```
mark-prins, line 7: what do you think about adding:

```
config:
  nonOptionalTypename: true
```

to enforce `__typename`?

mark-prins (approved):
🎉 

that looks super! have read through the docs and don't have anything to add. it's working :)

joshxg:
Nice! Added. Though I think that's putting the required typename in every type that is generated, even if we don't actually query for it. Still going to be way nicer, but should probably be a little cautious, I suppose

Issue #831, opened 2022-02-17
Update Requisition queries


joshxg:
- Update the requisition queries & mutations to use the api correctly
- Extract the queries into an operations file in the requisitions module

PR #833, opened 2022-02-17
Add new requisition queries in requisition module

joshxg:
#831 - 1!

- This mainly adds the new queries and gets the pages working, though there is a lot more work to do!

### DIFF
--- a/packages/requisitions/src/SupplierRequisition/api/operations.graphql
+++ b/packages/requisitions/src/SupplierRequisition/api/operations.graphql
@@ -1,0 +1,124 @@
+mutation insertRequestRequisition(
+  $storeId: String!
+  $input: InsertRequestRequisitionInput!
+) {
+  insertRequestRequisition(input: $input, storeId: $storeId) {
+    ... on RequisitionNode {
+      __typename
+      id
+    }
+  }
+}
+
+mutation updateRequestRequisition(
+  $storeId: String!
+  $input: UpdateRequestRequisitionInput!
+) {
+  updateRequestRequisition(input: $input, storeId: $storeId) {
+    ... on RequisitionNode {
+      __typename
+      id
+    }
+  }
+}
+
+# mutation deleteSupplierRequisitions($ids: [DeleteSupplierRequisitionInput!]) {
+#   batchSupplierRequisition(deleteSupplierRequisitions: $ids) {
+#     __typename
+#     deleteSupplierRequisitions {
+#       __typename
+#       id
+#     }
+#   }
+# }
+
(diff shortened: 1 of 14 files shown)


packages/requisitions/src/SupplierRequisition/api/operations.graphql
```diff
@@ -0,0 +1,124 @@
...
+
+# mutation deleteSupplierRequisitions($ids: [DeleteSupplierRequisitionInput!]) {
+#   batchSupplierRequisition(deleteSupplierRequisitions: $ids) {
+#     __typename
+#     deleteSupplierRequisitions {
+#       __typename
+#       id
+#     }
+#   }
+# }
```
joshxg, line 33: there's not batch mutation, so mass-deleting requisitions isn't ideal

packages/requisitions/src/SupplierRequisition/api/operations.graphql
```diff
@@ -0,0 +1,124 @@
...
+  id
+  itemId
+  requestedQuantity
+  calculatedQuantity
+  itemStats {
+    averageMonthlyConsumption
+    stockOnHand
+    monthsOfStock
+  }
+}
```
joshxg, line 45: not sure about this, but not bad to gen the type from this subset of fields. Though, the type to use is `RequestRequisitionLineFragment` (and the same for the other fragments below) - I hate the suffix. Can use a config option to remove *all* suffixes, which would take a bit of time to rename the whole code base (i..e  `requisitionQuery` would be `requisition`). I also think fighting against this has always just ended up in more time spent, so thought to just go with it 🤷

packages/requisitions/src/utils.ts
```diff
@@ -1,25 +1,25 @@
-import { SupplierRequisitionNodeStatus } from '@openmsupply-client/common';
+import { RequisitionNodeStatus } from '@openmsupply-client/common';
```
joshxg, line 1: Changes to get rendering to work, not exactly the right logic

joshxg:
Have a list here of all the things to be done for request requisitions. Then repeat for response: https://github.com/openmsupply/openmsupply-client/issues/831

mark-prins (approved):
👍  looks good. maybe also rename the `SupplierRequisition` and `CustomerRequisition` folders? Is that too much?

joshxg:
> Looks good. maybe also rename the SupplierRequisition and CustomerRequisition folders? Is that too much?

Will do - will just do a new PR to try make the diff easier to view. Unsure if the 'internal order' naming plays a part in the naming at all

PR #835, opened 2022-02-17
#831 Add feature to only show  stores within the name search modal and requisitions statuses

joshxg:
#831 

- Requisitions should only be made to a store, not just any supplier - we get errors otherwise!
- Also added requisitions status changes - should have been in a new branch ugh.


packages/system/src/Name/Components/NameSearchModal/NameSearchModal.tsx
```diff
@@ -7,7 +7,7 @@ interface NameSearchProps {
   open: boolean;
   onClose: () => void;
   onChange: (name: Name) => void;
-
+  onlyShowStores?: boolean;
```
joshxg, line 10: there's no filter in the API to do this, so we need to do it in JS. Not such a big deal. Though, this boolean annoys me.

An `extraFilter` prop where we can pass a more generic filtering function was another thought, though getting a list of 'Internal suppliers' is relatively common.

I had a thought that there should just be more components. `CustomerSearchModal`, `SupplierSearchModal`, `InternalSupplierSearchModal` etc instead

packages/requisitions/src/SupplierRequisition/DetailView/Footer/Footer.tsx
```diff
@@ -0,0 +1,69 @@
+import {
```
joshxg, line 1: shifted this file and essentially made it actually work

packages/requisitions/src/SupplierRequisition/DetailView/Footer/StatusChangeButton.tsx
```diff
@@ -0,0 +1,133 @@
+import React, { useMemo, useState, useEffect } from 'react';
```
joshxg, line 1: All very similar/exactly the same as the other status change buttons

packages/requisitions/src/SupplierRequisition/DetailView/DetailView.tsx
```diff
@@ -10,7 +10,7 @@ import {
 } from '@openmsupply-client/common';
 import { useRequestRequisition, useIsRequestRequisitionDisabled } from '../api';
 import { Toolbar } from './Toolbar';
-import { Footer } from './Footer';
+import { Footer } from './Footer/Footer';
```
mark-prins, line 13: line 29 - the disabled state on AppBarButtons is reversed

packages/system/src/Name/Components/NameSearchModal/NameSearchModal.tsx
mark-prins: agree - creating more components would help here.

mark-prins (approved):
👍  LGTM!

packages/requisitions/src/SupplierRequisition/DetailView/DetailView.tsx
joshxg: Ah thanks, oops! Lucky it doesn't do anything yet 😁

Issue #830, opened 2022-02-17
Shipments API issues


mark-prins:
Part of #825 - the API for invoices is broken with the latest changes

- [x] List view for inbound and outbound
- [x] Detail view for inbound and outbound
- [x] Inbound mutations
- [x] Outbound mutations
- [x] Dashboard request to invoices

PR #836, opened 2022-02-17
#830 shipments api issues

mark-prins:
Fixes part of #830 

Have got the list and detail views working, with a few things to sort out

- I want to shift the token out, probably to a cookie and set expiry and handle refresh
- And, in the same way.. would like to move the `storeId` out somewhere
- The `operations.graphql` - local versions have been moved into common again.. as the `OmSupplyApiContext` is only loading the common `operations.generated.ts`
- The outbound batch edit isn't showing correctly


packages/common/src/hooks/useQueryParams/useQueryParams.ts
```diff
@@ -28,6 +30,7 @@ export const useQueryParams = <T extends DomainObject>({
   const filter = useFilterBy(initialFilterBy);
   const sort = useSortBy(initialSortBy);
   const pagination = usePagination();
+  const [storeId] = useLocalStorage('/authentication/storeid', '');
```
mark-prins, line 33: concerned that this crosses the line between pure, common functions and host concerns. The `storeId` is now a parameter though, so logically it made sense to have it in here..

packages/common/src/localStorage/keys.ts
```diff
@@ -20,6 +20,7 @@ export type LocalStorageRecord = {
   '/theme/logo': string;
   '/mru/credentials': AuthenticationCredentials;
   '/authentication/token': string;
+  '/authentication/storeid': string;
```
mark-prins, line 23: would like to lift the auth aspects out to a cookie - mainly to set expiry, but also for a separation between the rest of the local storage concerns.

packages/invoices/src/OutboundShipment/DetailView/DetailView.tsx
```diff
@@ -15,23 +19,27 @@ import { Footer } from './Footer';
...
+  const t = useTranslation('distribution');
+  const navigate = useNavigate();
   const onRowClick = useCallback(
     (item: InvoiceLine | InvoiceItem) => {
       onOpen(toItem(item));
     },
     [toItem, onOpen]
   );
 
+  if (isLoading) return <DetailViewSkeleton hasGroupBy={true} hasHold={true} />;
```
mark-prins, line 36: this was an offroad - it bothered me that the alert / redirect wasn't enabled for outbound

packages/invoices/src/OutboundShipment/DetailView/Footer/StatusChangeButton.tsx
```diff
@@ -130,7 +130,10 @@ const useStatusChangeButton = () => {
   // When the status of the invoice changes (after an update), set the selected option to the next status.
   // It would be set to the current status, which is now a disabled option.
   useEffect(() => {
-    setSelectedOption(() => getNextStatusOption(status, options));
+    const nextSelectedOption = getNextStatusOption(status, options);
```
mark-prins, line 133: clumsy attempt to work around this:

Screen Shot 2022-02-18 at 10 12 17 AM
 
The `setState` was having trouble comparing the selection option against the options, and I didn't see why they were being recreated.

packages/common/src/hooks/useQueryParams/useQueryParams.ts
joshxg: I suppose the intention was more of a simplified way of using filtering, sorting & pagination for some queries. The storeID is needed in a lot of places so I see the motivation.

I can see how all this auth stuff isn't really fitting our current structure.. I wonder about another `auth`  package.. seems excessive, but I can see it being helpful

packages/common/src/localStorage/keys.ts
joshxg: I'm not following this point, might have to get you to spell it out to me 😁  🤔

packages/host/src/components/Login/hooks.ts
```diff
@@ -71,6 +72,8 @@ export const useLoginForm = (
     setAuthToken(token);
     setMRUCredentials({ username: username, store: store });
     setUser({ id: '', name: username });
+    setStoreId(store?.id ?? '');
```
joshxg, line 75: ah yes - I guess this is where we diverged on the store ID a bit. I used the `store` object in `hostContext` - the problem with that approach is that it's not set in the host context when refreshing the page etc like it is for local storage

packages/invoices/src/InboundShipment/DetailView/api.ts
```diff
@@ -244,7 +248,8 @@ export const getInboundShipmentDetailViewApi = (
 export const useInboundShipment = (): UseQueryResult<Invoice, unknown> => {
   const { id = '' } = useParams();
   const { api } = useOmSupplyApi();
-  const queries = getInboundShipmentDetailViewApi(api);
+  const { storeId } = useQueryParams({ initialSortBy: { key: 'id' } });
```
joshxg, line 251: I suppose this is the downside of using that hook for the store id - annoying!

packages/invoices/src/OutboundShipment/DetailView/Footer/StatusChangeButton.tsx
joshxg: Ah yes I just ran up against this. 

turns out `t` from `useTranslation` is a new reference each render, so using it in a deps array causes some unwanted behaviour!

i.e. in here:

```
const options = useMemo(
    () => getStatusOptions(status, getButtonLabel(t)),
    [status, getButtonLabel, t]
  );
```

Taking the `t` out stops the infinite re-rendering (Though, doesn't solve the root cause)

joshxg (approved):
nice!

packages/invoices/src/OutboundShipment/DetailView/Footer/StatusChangeButton.tsx
mark-prins: aha! didn't notice that .. well spotted

PR #840, opened 2022-02-18
#831 Add other party / supplier input for supplier requisition

joshxg:
#831 - update detail view Toolbar component part1..

- Started this and can't finish it all, but thought to just commit this for now
- There is no requisition date in the api
- There is no bulk delete lines
- There is no updating of other party - thought to put the input there anyway 🤷 
- These are the main missing pieces so far of the API, so this comment likely disproportionately seems negative 😁


mark-prins (approved):
🎉  LGTM!

PR #842, opened 2022-02-18
Add pagination, filtering & sorting

joshxg:
#831 

- Adds sorting/pagination/filtering to internal orders list view
- Didn't use `useListData` as there is no delete function yet and this was easier


mark-prins (approved):
👍  looks great

Issue #843, opened 2022-02-18
Update Response requisition queries


joshxg:
- [ ] Create response requisition operations file
- [ ] Add hook for querying for list data
- [ ] Update detail view with new fields and types
- [ ] Rename everything from `CustomerRequisiton` -> `ResponseRequisition` though I believe the UI will use `Customer Requisition`

PR #844, opened 2022-02-19
Add response requisition queries

joshxg:
#843 

- Starting by creating the operations file and hooks.. nothing hooked up really


packages/requisitions/src/CustomerRequisition/api/api.ts
```diff
@@ -1,96 +1,102 @@
...
 export const requisitionToInput = (
-  requisition: Partial<Requisition> & { id: string }
-): UpdateCustomerRequisitionInput => {
+  requisition: Partial<ResponseRequisitionFragment> & { id: string }
+): UpdateResponseRequisitionInput => {
   return {
     id: requisition.id,
-    orderDate: requisition.requisitionDate?.toISOString(),
-    otherPartyId: requisition.otherParty?.id,
+    // otherPartyId: requisition.otherParty?.id,
```
joshxg, line 24: the other party ID is not supported currently in the API.. though I'm not sure it should be.. maybe just for request requisitions

packages/requisitions/src/CustomerRequisition/api/api.ts
```diff
@@ -1,96 +1,102 @@
...
     comment: requisition.comment,
     theirReference: requisition.theirReference,
-    color: requisition.color,
-    status: requisition.status,
+    colour: requisition.colour,
+    status: requisition.status
+      ? UpdateResponseRequisitionStatusInput.Finalised
+      : undefined,
```
joshxg, line 30: we can only change a response to finalised

packages/requisitions/src/CustomerRequisition/api/hooks.ts
```diff
@@ -1,5 +1,6 @@
 import { useMemo } from 'react';
 import {
+  useHostContext,
```
joshxg, line 3: ah.. didn't we speak about this? Was I meant to use the store ID from elsewhere..? Local storage?

packages/requisitions/src/CustomerRequisition/api/operations.graphql
```diff
@@ -0,0 +1,114 @@
...
+    ... on RequisitionNode {
+      __typename
+      id
+    }
+  }
+}
+
+fragment ResponseRequisitionLine on RequisitionLineNode {
+  id
+  itemId
```
mark-prins, line 15: what about adding in 
```
item {
  id
  code
  name
}
```

instead of itemId - to use item name and code in the list? 
either add a custom accessor or flatten the response object..

mark-prins (approved):
👍  LGTM!

Ignore comments that are covered in upcoming PRs.. I'm working my way from old to new.. so could well be commenting on things which have been changed already!

side comment: the method `useFieldsSelector` in `@common/api/hooks/useFieldSelector.ts` has a variable called `requisitions` which should be more generic (`entity`?)

packages/requisitions/src/CustomerRequisition/api/operations.graphql
joshxg: I took the easy route and made an issue https://github.com/openmsupply/openmsupply-client/issues/853

In general I suppose this will be a little easier to develop around when we can add items to a request requisition

joshxg:
> side comment: the method useFieldsSelector in @common/api/hooks/useFieldSelector.ts has a variable called requisitions which should be more generic (entity?)

Ha! woops!

PR #846, opened 2022-02-20
#843 response requisition queries 3

joshxg:
#843 

- Adds some updates for compiling - think everything is compiling except for stocktakes right now
- Adds updates to the detail view components to use the new hooks/queries


codegen.yml
```diff
@@ -15,7 +15,7 @@ generates:
     preset: near-operation-file
     presetConfig:
       extension: .generated.ts
-      baseTypesPath: ./common/src/types/schema.ts
+      baseTypesPath: ~@openmsupply-client/common
```
joshxg, line 18: The path for the base paths (i.e. schema.ts) becomes relative in the operation fils, which typescript doesn't like as it's not within the dir specified. This fixes that.. though it also uses this same import path for within `common`.. which seems to work ok right now, but it doesn't seem right. However, it's better than the original problem so..

packages/requisitions/src/CustomerRequisition/DetailView/Toolbar.tsx
```diff
@@ -11,42 +11,27 @@ import {
...
-  } = useCustomerRequisitionFields([
-    'lines',
-    'otherParty',
-    'orderDate',
-    'requisitionDate',
-    'theirReference',
-  ]);
-  const [bufferedReqDate, setBufferedReqDate] = useBufferState(requisitionDate);
-  const [bufferedOrderDate, setBufferedOrderDate] = useBufferState(orderDate);
+  const isDisabled = useIsResponseRequisitionDisabled();
```
joshxg, line 26: dates aren't handled by the API right now so have removed

joshxg (commented):
not too much to say.. just utilising the hooks that have been written previously

mark-prins (approved):
👍 

One question.. the list view is navigating to `id` but the detail view is querying by `requisitionNumber`. Is that changing later?

I've had to use this for the onClick: `navigate(String(row.requisitionNumber));`
( I then get blank lines 😁  but that's just the columns / props requested etc )

joshxg:
Ugh yes, thanks - have added another issue 🙄

Issue #847, opened 2022-02-20
Update stocktake queries for new API


joshxg:
- [ ] Create a stocktake's operations file and update
- [ ] Create hooks to utilise new operation file generated code/queries
- [ ] Update stocktake list view to utilise new hooks
- [ ] Update stocktake detail view to utilise new hooks

PR #855, opened 2022-02-21
#847 Stocktake query updates for DetailView

joshxg:
#847 

- Changes to `DetailView` to start using the new queries/types
- More slight backward steps to try and move forward. Am not getting anymore `inventory`/`stocktake` type errors, only a couple invoice ones, now


packages/inventory/src/Stocktake/DetailView/DetailView.tsx
```diff
@@ -15,21 +15,29 @@ import { Toolbar } from './Toolbar';
...
-    onOpen(toItem(item));
+  const onRowClick = (item: StocktakeLineFragment | StocktakeSummaryItem) => {
+    if ('lines' in item) return onOpen(toItem(item));
+    onOpen(
+      toItem({
+        itemId: item.item?.id ?? '',
+        itemName: item.item?.name ?? '',
+        itemCode: item.item?.code ?? '',
+      })
+    );
```
joshxg, line 40: This is a little annoying and will need to revisit how this works properly rather than the hack here. Essentially the `toItem` was using a `ItemLike` parameter which was a `line` that had properties `itemId`, `itemName`, `itemCode` and turning it into an `Item` type - now that `StocktakeLine` doesn't have those properties, rather just the `item` field, it's a little different. 

`RequisitionLineNode` and `StocktakeLineNode` don't have those properties, but `InvoiceLineNode` does - all I want is consistency so I can use the same function for them all! 

Thinking we should just go with using an `item` field - though, for invoices, that means using the guard and things as the invoice types are using the 'older' pattern

packages/inventory/src/Stocktake/DetailView/Toolbar.tsx
```diff
@@ -7,22 +7,17 @@ import {
...
 export const Toolbar: FC = () => {
   const t = useTranslation(['distribution', 'common', 'inventory']);
   const { success, info } = useNotification();
   const isDisabled = useIsStocktakeDisabled();
-  const { stocktakeDatetime, description, update } = useStocktakeFields([
-    'description',
-    'stocktakeDatetime',
-  ]);
+  const { description, update } = useStocktakeFields('description');
   const [descriptionBuffer, setDescriptionBuffer] = useBufferState(description);
```
joshxg, line 21: stocktake date time is not a thing - there's no editable dates right now

packages/inventory/src/Stocktake/DetailView/columns.ts
```diff
@@ -8,40 +8,47 @@ import {
   Column,
   SortBy,
 } from '@openmsupply-client/common';
-import { StocktakeLine, StocktakeSummaryItem } from './../../types';
+import { StocktakeSummaryItem } from './../../types';
```
joshxg, line 11: this is not ideal and might be a reason why having the flattened item fields could be a better way to go. Alternatively, making this function be able to handle nested paths. For now, I've hacked away in favour of getting your opinion/doing a proper fix in a dedicated issue

packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/hooks.ts
```diff
@@ -58,14 +59,12 @@ const stockLineToDraftLine = (
...
     stocktakeId,
-    itemCode: '',
-    itemName: '',
     countThisLine: false,
     isCreated: false,
     isUpdated: false,
     ...line,
     snapshotNumberOfPacks: line.totalNumberOfPacks,
-    expiryDate: line.expiryDate ? new Date(line.expiryDate) : null,
+    expiryDate: line.expiryDate ? line.expiryDate : null,
```
joshxg, line 67: I removed the parsing of dates in the query into date objects to avoid the performance problem we were seeing with invoices and non-serialisable data

packages/inventory/src/Stocktake/DetailView/columns.ts
mark-prins: I was wondering about extracting an array of items and using that
```
              const items = getItems(row);
              return ifTheSameElseDefault(items, 'code', '');
```

with 
```
const getItems = (row: StocktakeSummaryItem) => {
  const { lines } = row;
  const items =
    lines?.filter(line => !!line.item).map(line => line.item as ItemNode) ?? [];
  return items;
};
```

similar to supporting nested paths - just making it explicit at this level 🤷 

I quite like the `item` object.. but it is much more annoying to actually use

mark-prins (approved):
🎉  that looks great - nothing very helpful to add!
Nice to have the stocktakes working 😁

packages/inventory/src/Stocktake/DetailView/DetailView.tsx
joshxg: sounds good!

packages/inventory/src/Stocktake/DetailView/columns.ts
joshxg: Ah yes, thanks, that's a very nice approach.

PR #865, opened 2022-02-21
#854 use requisition numbers for responses

### LINKED ISSUE #854: Response requisition navigation to detail view should use the requisition number not id
I updated request requisitions to use the `requisitionNumber` but forgot to update response requisitions

joshxg:
Fixes #854 

- Adds use of `requisitionNumber` for navigating to response reqs (I had to use a local copy of remote-server to test this, rather than the demo server)
- Off-road added internal orders to commandK - though I didn't add a shortcut as all the letters are taken!!

### DIFF
--- a/packages/host/src/CommandK.tsx
+++ b/packages/host/src/CommandK.tsx
@@ -184,6 +184,17 @@ export const CommandK: FC = ({ children }) => {
         ),
     },
     {
+      id: 'navigation:internal-order',
+      name: `${t('cmdk.goto-internal-order')}`,
+      keywords: 'replenishment',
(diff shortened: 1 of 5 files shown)


mark-prins (approved):
👍  very nice!
I thought about using I+O as a shortcut but then we'd have to change inbound shipments to I+S.. and really don't think it's necessary having shortcuts on everything anyway!

joshxg:
Though it is nice to have the shortcuts, I do find them very useful. I couldn't figure out how to use special keys for shortcuts, which would maybe make it a bit easier :(

PR #866, opened 2022-02-21
Add response requisition status updates

### LINKED ISSUE #849: Update response requisition statuses
Currently response requisitions are using request requisition statuses which isn't correct

joshxg:
Fixes #849 

- Puts the confirmation modal and correct status changes into response
- Off road clean ups of the requisition utils file


packages/requisitions/src/RequestRequisition/DetailView/Footer/Footer.tsx
```diff
@@ -7,35 +7,21 @@ import {
...
 
-  if (statusIdx >= 0) {
-    statusLog[RequisitionNodeStatus.Draft] = requisition.createdDatetime;
-  }
-  if (statusIdx >= 1) {
-    statusLog[RequisitionNodeStatus.Sent] = requisition.sentDatetime;
-  }
-  if (statusIdx >= 2) {
-    statusLog[RequisitionNodeStatus.Finalised] = requisition.finalisedDatetime;
-  }
```
joshxg, line 37: This was a bit funny because of the mock server and making sure the dates were set properly. In reality the dates would just be `null` anyway, so we don't need to do all this

packages/requisitions/src/utils.ts
```diff
@@ -3,15 +3,21 @@ import {
   LocaleKey,
   useTranslation,
 } from '@openmsupply-client/common';
-import { Requisition } from './types';
 import { ResponseRequisitionRowFragment } from './ResponseRequisition/api';
+import { RequestRequisitionRowFragment } from './RequestRequisition/api';
```
joshxg, line 8: misc cleaning up of this file. Think I'm happy with it in general, though there is a lot of duplication and it could be more DRY I don't think it's worthwhile

packages/requisitions/src/utils.ts
```diff
@@ -95,6 +56,18 @@ export const getNextRequestRequisitionStatus = (
...
+  );
+  const nextStatus = responseRequisitionStatuses[currentStatusIdx + 1];
+  return nextStatus ?? null;
 };
+
+export const canDeleteRequisition = (
+  requisitionRow: ResponseRequisitionRowFragment | RequestRequisitionRowFragment
+): boolean =>
+  requisitionRow.status === RequisitionNodeStatus.Draft ||
+  requisitionRow.status === RequisitionNodeStatus.New;
```
joshxg: Can only delete requests, at the moment. In the future, should be able to delete adhoc/manual response reqs with no linked shipments..

packages/requisitions/src/utils.ts
```diff
@@ -95,6 +57,18 @@ export const getNextRequestRequisitionStatus = (
...
+  );
+  const nextStatus = responseRequisitionStatuses[currentStatusIdx + 1];
+  return nextStatus ?? null;
 };
+
+export const canDeleteRequisition = (
+  requisitionRow: ResponseRequisitionRowFragment | RequestRequisitionRowFragment
+): boolean =>
+  requisitionRow.status === RequisitionNodeStatus.New &&
+  requisitionRow.type === RequisitionNodeType.Request;
```
joshxg, line 74: Only deleting request reqs. In the future, manual response reqs with no linked shipments

packages/requisitions/src/ResponseRequisition/DetailView/Footer/StatusChangeButton.tsx
```diff
@@ -0,0 +1,128 @@
...
+};
+
+const getNextStatusOption = (
+  status: RequisitionNodeStatus,
+  options: SplitButtonOption<RequisitionNodeStatus>[]
+): SplitButtonOption<RequisitionNodeStatus> | null => {
+  if (!status) return options[0] ?? null;
+
+  const nextStatus = getNextResponseRequisitionStatus(status);
+  console.log('nextStatus', nextStatus);
```
mark-prins, line 54: not required?

packages/requisitions/src/ResponseRequisition/DetailView/Footer/StatusChangeButton.tsx
```diff
@@ -0,0 +1,128 @@
...
+  currentStatus: RequisitionNodeStatus,
+  getButtonLabel: (status: RequisitionNodeStatus) => string
+): SplitButtonOption<RequisitionNodeStatus>[] => {
+  const options: [
+    SplitButtonOption<RequisitionNodeStatus>,
+    SplitButtonOption<RequisitionNodeStatus>
+  ] = [
+    {
+      value: RequisitionNodeStatus.New,
+      label: getButtonLabel(RequisitionNodeStatus.Draft),
```
mark-prins, line 30: there's a disconnect between the crumbs and the split button - one has `Draft > Finalised` and the other `New > Finalised` 

I'd run with Draft for both 🤷 


Screen Shot 2022-02-22 at 9 40 04 AM
Screen Shot 2022-02-22 at 9 40 10 AM

mark-prins (approved):
🎉  ah - thanks! that's good to have tidied up!

One concern.. I think the status options are the wrong way around - for an internal order, the user is receiving stock, but at the moment they can 'Confirm Sent'. Should just be 'Draft' & 'Finalised' for that one, and have the option of confirming 'Sent' for a response requisition / customer requisition

joshxg:
I've tried followed what's on here:

image

I'll try to describe fully my understanding so we can pinpoint where my confusion is

- So we start with an internal order, which is manually created so the status is `DRAFT`. 
- To 'move forward' with that, the status is set to `SENT` (which I can see is confusing, as this is more like 'the order
 itself has been sent and not 'the medicine' has been sent?). 
- Once `SENT`, a response requisition is created on the suppliers side, which is not manually created so the status is `NEW`. 
- The supplier then can make customer invoices from that requisition, and once there are no more invoices to create, the requisition can be set to `FINALISED`, which will also trigger a status change on the customers request requisition

joshxg:
@mark-prins Ugh, sorry, I merged a PR into this one and messed the diff - I've made a new issue to chase up the status problem here: https://github.com/openmsupply/openmsupply-client/issues/878

mark-prins:
> * To 'move forward' with that, the status is set to `SENT` (which I can see is confusing, as this is more like 'the order

Oh I see.. I'm thinking of `SENT` as 'sending stock' but it is really `CONFIRMED`. huh. the terms used aren't working for me 🤷

joshxg:
> Oh I see.. I'm thinking of SENT as 'sending stock' but it is really CONFIRMED. huh. the terms used aren't working for me

Yeah I agree, it took me a bit to wrap my head around

Issue #857, opened 2022-02-21
Update `system`/`item`/`toItem` fn to work with `item` fields rather than `itemName`, `itemCode` etc fields


joshxg:
https://github.com/openmsupply/openmsupply-client/pull/855#discussion_r810741048

PR #868, opened 2022-02-21
#857 Update Stocktakes to use item field rather than flattened item fields

joshxg:
#857 

Essentially the `StocktakeEditModal` was expecting an `Item` type, which is essentially a full `ItemNode`, even though the only fields actually being utilised was the item `id`/`code`/`name` (should probably be using `unitName` also). 

We were creating a pseudo-Item object from the `itemId`, `itemName` and `itemCode` fields on a `StocktakeLine`. However the new API queries do not have these fields, instead opting for an `item` field.

A simple fix would have been to just query for the full `ItemNode` when fetching a `StocktakeLine`, however that's a little excessive given we only need 3 simple fields. Instead, I created an `ItemRowFragment` (bad name?) and tried to utilise this 'smaller' item fields subset type.

- Bit of offroad work here in the `item` package which I'll comment on..
- Haven't touched the invoice part of it, didn't want to cause conflicts


packages/common/src/operations.generated.ts
```diff
@@ -4,17 +4,6 @@ import { GraphQLClient } from 'graphql-request';
 import * as Dom from 'graphql-request/dist/types.dom';
```
joshxg, line 4: this is a big diff for this file! I just took *out* the `itemListView` query and put it in the `system/item` package operations file.

There is another item related query here: `itemsWithStockLines` which I *duplicated* into the `system/item` package as it is also used in `system/stock`. These queries would likely differ in the future so I think duplicating the query in *each* package is likely the way to go.

packages/common/src/operations.graphql
```diff
@@ -1,54 +1,3 @@
-query names(
```
joshxg, line 1: Again a big diff, what was changed here was simply the removal of the `itemsListViw` generated code

packages/inventory/src/Stocktake/DetailView/DetailView.tsx
```diff
@@ -24,20 +23,14 @@ import { AppRoute } from '@openmsupply-client/config';
...
   const onRowClick = (item: StocktakeLineFragment | StocktakeSummaryItem) => {
-    if ('lines' in item) return onOpen(toItem(item));
-    onOpen(
-      toItem({
-        itemId: item.item?.id ?? '',
-        itemName: item.item?.name ?? '',
-        itemCode: item.item?.code ?? '',
-      })
-    );
+    if (item.item) onOpen(item.item);
```
joshxg, line 33: What I have changed is that a `StockLineFragment` has an `item` field with `{ id, code, name }`, which is equal to an `ItemRowFragment`. Then, the `SummaryItem` also has a `item` field with the same shape, so instead of trying to 'fix' the `toItem` function which the original issue was for, I just removed the need to use it

packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/StocktakeLineEdit.tsx
```diff
@@ -1,4 +1,5 @@
 import React, { FC, useState } from 'react';
+import { ItemRowFragment } from '@openmsupply-client/system';
```
joshxg, line 2: Then, I just updated the need to use the `Item` type in the `StocktakeLineEdit` modal to use the `ItemRowFragment` instead

packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/StocktakeLineEditForm.tsx
```diff
@@ -1,20 +1,19 @@
...
-  Item,
   ModalRow,
   ModalLabel,
   Grid,
   useTranslation,
   BasicTextInput,
   ModalMode,
 } from '@openmsupply-client/common';
-import { ItemSearchInput } from '@openmsupply-client/system';
+import { ItemSearchInput, ItemRowFragment } from '@openmsupply-client/system';
```
joshxg, line 10: more modal changes to go from `Item` -> `ItemRowFragment`

packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/hooks.ts
```diff
@@ -1,13 +1,13 @@
...
 import React, { useEffect, Dispatch, SetStateAction } from 'react';
 import {
   useParams,
   StockLine,
   RecordPatch,
   generateUUID,
-  Item,
 } from '@openmsupply-client/common';
-import { toItem, useStockLines } from '@openmsupply-client/system';
+import { useStockLines } from '@openmsupply-client/system';
```
joshxg, line 10: more modal changes to go from `Item` -> `ItemRowFragment`

packages/inventory/src/Stocktake/api/hooks.ts
```diff
@@ -128,9 +128,7 @@ export const useStocktakeItems = (): UseQueryResult<StocktakeSummaryItem[]> => {
       ([itemId, lines]) => {
         return {
           id: itemId,
-          itemId,
-          itemName: lines[0].item?.name ?? '',
-          itemCode: lines[0].item?.code ?? '',
+          item: lines[0].item,
```
joshxg, line 131: this is changing the summary item creation to use the item field

packages/inventory/src/types.ts
```diff
@@ -1,9 +1,8 @@
+import { ItemRowFragment } from '@openmsupply-client/system';
 import { StocktakeLineFragment } from './Stocktake/api/operations.generated';
 
 export interface StocktakeSummaryItem {
   id: string;
-  itemId: string;
-  itemName: string;
-  itemCode: string;
+  item?: ItemRowFragment | null;
```
joshxg, line 6: For some reason a stocktake line has an `Item` as optional, which is a little odd!

packages/system/src/Item/Components/ItemSearchInput/ItemSearchInput.tsx
```diff
@@ -4,9 +4,33 @@ import {
   useFormatNumber,
   useTranslation,
   styled,
+  Autocomplete,
```
joshxg, line 7: This got a little hairy!

The `ItemSearchInput` was expecting an `Item` to be passed as the 'current item'. The stocktake is now using an `ItemRowFragment`!

So.. went back to the `toItem` casting .. hack? which just creates the `Item` type to make typescript happy. In general I think this component is confusing and over complicated for what it needs to do.

Really, just passing an `id` for the `item` and doing a `.find` in the items returned would be a bit easier

packages/system/src/Item/ListView/api.ts
```diff
@@ -1,57 +0,0 @@
-import {
```
joshxg, line 1: we can't do anything other than read system data, so using this interface seemed a bit silly?

packages/system/src/Item/api/api.ts
```diff
@@ -0,0 +1,89 @@
...
+
+const itemsGuard = (itemsQuery: ItemsListViewQuery) => {
+  if (itemsQuery.items.__typename === 'ItemConnector') {
+    return itemsQuery.items;
+  }
+
+  throw new Error(itemsQuery.items.error.description);
+};
+
+export const ItemQueries = {
```
joshxg, line 23: looking at this now, maybe it should be

```
const getItemQueries = (api, storeID) => ({ ... })..
```

And then 

```
const useItemApi = () => {
    const storeId = getStoreId()
    const  {client} = getClient()
    const sdk = getSdk(client)
    return getItemQueries(storeId, sdk)
}
```


and is probably a pattern to use elsewhere

packages/system/src/Item/api/hooks/useItem/useItem.ts
```diff
@@ -1,25 +1,24 @@
+import { ItemFragment } from './../../operations.generated';
```
joshxg, line 1: Changes to use the code generated `ItemFragment` rather than the `Item` type and also use the co-located generated operations query rather than from common

packages/system/src/Item/api/hooks/useItemListView/useItemListView.ts
```diff
@@ -0,0 +1,23 @@
+import { useQueryParams, useQuery } from '@openmsupply-client/common';
```
joshxg, line 1: This is the hook made instead of using the `ListApi` interface which had stubs for 3/4 methods

packages/system/src/Item/api/hooks/useItemsList/useItemsList.ts
```diff
@@ -8,11 +7,14 @@ import {
   UseQueryResult,
   useQueryClient,
 } from '@openmsupply-client/common';
-import { getItemSortField, mapItemNodes } from '../../utils';
+import { useItemApi } from './../useItemApi';
```
joshxg, line 10: Changes to use the code generated `ItemFragment` rather than the `Item` type and also use the co-located generated operations query rather than from common

packages/system/src/Item/api/api.ts
mark-prins: yes, that's the pattern I'm ending up with - I think it's about the best so far

packages/system/src/Item/Components/ItemSearchInput/ItemSearchInput.tsx
mark-prins: what about going the other way, and changing the list query to return `ItemRowFragment[]`?
Would that help? It would mean adding `availableQuantity` to the fragment.. the only props used are `code` `name` and `availableQuantity` in here.. and I think consumers of the component would only need the `id`. 

Just a thought - the current `toItem` is all good!

mark-prins (approved):
🎉  brilliant! lots of nifty work going on in there.. nice use of fragments 😄

packages/system/src/Item/Components/ItemSearchInput/ItemSearchInput.tsx
joshxg: Yeah that's far better. I've had a quick look at changing that around and it will require changes in multiple places which are using the search input and a pretty big diff. 

So..  new issue to revisit when things are a little more stable.. 🙄

PR #875, opened 2022-02-22
#807 delete location

### LINKED ISSUE #807: Location: Delete
It's not possible to delete a location in the UI - there was a requirement line in the spreadsheet though.
We can make new ones, I think we should allow deletes.

mark-prins:
Fixes #807 

Adds the delete option for locations, using the same pattern as in the other list views.
Trouble is the lack of a batch mutation.. so I've hacked up an alternative. I doubt that location delete will be used often, so probably ok in this case.
Also added a filter for the list and then removed it after realising that there is only an `equalTo` filter available.


packages/common/src/intl/locales/en/distribution.json
```diff
@@ -27,15 +27,15 @@
   "label.select-rows-to-delete-them": "Select rows to delete them",
   "label.units-in-pack-size-of_one": "unit in pack size of",
   "label.units-in-pack-size-of_other": "units in pack size of",
-  "message.cant-delete-invoices": "Can only delete invoices with a status of 'New' or 'Allocated'",
-  "message.cant-delete-requisitions": "Can only delete requisitions with a status of 'Draft'",
-  "message.deleted-invoices": "Deleted {{number}} invoices",
-  "message.deleted-requisitions": "Deleted {{number}} requisitions",
-  "message.select-rows-to-delete": "Select rows to delete",
+  "messages.cant-delete-invoices": "Can only delete invoices with a status of 'New' or 'Allocated'",
```
mark-prins, line 30: yeah, sorry. I wandered offroad with this one.. noticed that some were `message` and some `messages` so I standardised.

packages/common/src/operations.graphql
```diff
@@ -317,6 +317,59 @@ mutation updateLocation($input: UpdateLocationInput!) {
...
+        ... on RecordBelongsToAnotherStore {
+          __typename
+          description
+        }
+        ... on LocationInUse {
+          __typename
+          description
+          stockLines {
+            ... on StockLineConnector {
+              __typename
```
mark-prins, line 344: am returning a little bit of the stock lines and invoice lines though I doubt that they are useful

packages/common/src/ui/components/portals/AppFooter/AppFooter.tsx
```diff
@@ -53,7 +53,9 @@ export const AppFooterPortal: FC<AppFooterPortalProps> = ({
         <Box {...boxProps}>{Content}</Box>
       </Portal>
       <Portal container={appSessionDetailsRef.current}>
-        <Box {...boxProps}>{SessionDetails}</Box>
+        <Box {...boxProps} style={{ paddingBottom: 3, paddingTop: 5 }}>
```
mark-prins, line 56: another offroad - it was a little cramped in the footer

packages/common/src/ui/notifications/AlertModal.tsx
```diff
@@ -5,7 +5,7 @@ import { AlertIcon } from '@common/icons';
 import { BasicModal } from '@common/components';
 
 export interface AlertModalProps {
-  message: string;
+  message: React.ReactNode;
```
mark-prins, line 8: just so I could use a `<ul/>` in the message. They can't be a child of `<p>` which the `<Typography/>` uses by default. Could specify a different component for that.. but since there is a div rendered anyway and it doesn't affect the look, I've just removed the `<Typography/>`

packages/system/src/Location/Components/Toolbar.tsx
```diff
@@ -0,0 +1,106 @@
...
+      .filter(Boolean) as Location[],
+  }));
+
+  const deleteAction = () => {
+    const numberSelected = selectedRows.length;
+    if (selectedRows && numberSelected > 0) {
+      const errors: DeleteError[] = [];
+      Promise.all(
+        selectedRows.map(async location => {
+          await deleteLocation(location).then(data => {
```
mark-prins, line 42: *cough* yes, not ideal.

joshxg: only for a bit till we can get a batch mutation!

With the batch mutation, I think it might also be easier to just throw an error within the mutation function and have `react-query` catch and set the errors in the `error` prop returned from `useMutation` also.

certainly something I've been neglecting so good to see some work put into error handling

packages/system/src/Location/Components/Toolbar.tsx
```diff
@@ -0,0 +1,106 @@
+import React, { FC, useEffect, useRef } from 'react';
```
joshxg, line 1: since this is only used in the list view, maybe should live in that folder?

packages/system/src/Location/ListView/api.ts
```diff
@@ -78,6 +79,24 @@ export const useLocationUpdate = (): UseMutationResult<
...
+  unknown
+> => {
+  const queryClient = useQueryClient();
+  const { api } = useOmSupplyApi();
+  const { storeId } = useQueryParams({ initialSortBy: { key: 'id' } });
+  return useMutation(
+    async (location: Location) =>
+      api.deleteLocation({ input: { id: location.id }, storeId: storeId }),
+    {
+      onSettled: () => queryClient.invalidateQueries(['location', 'list']),
```
joshxg, line 95: I wonder if this should be just `['location']` which would also invalidate some query like `['location', id]`

packages/common/src/intl/locales/en/distribution.json
joshxg: haha all good, prefer to get em done even if it bloats the diff a bit!

joshxg (approved):
awesome! 😁

packages/system/src/Location/Components/Toolbar.tsx
mark-prins: Good point - ta

packages/system/src/Location/ListView/api.ts
mark-prins: cool - might as well

PR #881, opened 2022-02-22
Set Authorization header

### LINKED ISSUE #838: Put Token as authentication header

mark-prins:
Fixes #838 

Stores the token in a plain text cookie 🤔  and adds the token to the Authoriztion header for all requests


packages/common/src/hooks/useAuthContext.ts
```diff
@@ -0,0 +1,54 @@
...
+  const [, setMRUCredentials] = useLocalStorage('/mru/credentials');
+  const { setHeader } = useOmSupplyApi();
+  const { token, store, user } = getAuthCookie();
+  const storeId = store?.id ?? '';
+
+  const login = (user: User, token: string, store?: Store) => {
+    setMRUCredentials({ username: user.name, store: store });
+
+    const authCookie = { store, token, user };
+    const expires = addMinutes(new Date(), COOKIE_LIFETIME_MINUTES);
```
mark-prins, line 42: will need some attention when the token refresh is done - at the moment you are logged out regardless of how active you are! Was thinking the refresh would update the token and set the expiry to that time + 60 mins, effectively keeping the user logged in

packages/common/src/hooks/useAuthContext.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    Cookies.set('auth', JSON.stringify(authCookie), { expires });
```
mark-prins, line 44: I decided that all of the cookie content was easily viewable and so no need to encrypt. open to suggestions

packages/common/src/hooks/useAuthContext.ts
```diff
@@ -0,0 +1,54 @@
...
+    setUser(user);
+    if (store) setStore(store);
+  };
+
+  useEffect(() => {
```
mark-prins, line 49: otherwise the header is only set when you login

packages/common/src/hooks/useHostContext.ts
```diff
@@ -46,8 +48,8 @@ export const useHostContext = create<HostContext>(set => ({
   detailPanelRef: null,
 
   setStore: store => set(state => ({ ...state, store })),
-  store: { id: '', code: '' },
+  store: authCookie.store ?? { id: '', code: '' },
```
mark-prins, line 51: set the host context correctly when reloading the page

packages/host/src/components/Navigation/RequireAuthentication.tsx
```diff
@@ -1,17 +1,18 @@
...
 } from '@openmsupply-client/common';
 import { AppRoute } from '@openmsupply-client/config';
 import React from 'react';
 
 export const RequireAuthentication = ({ children }) => {
-  const [authToken] = useLocalStorage('/authentication/token');
+  const { token } = useAuthContext();
   const location = useLocation();
-  const isProduction = process.env['NODE_ENV'] === 'production';
+  // const isProduction = process.env['NODE_ENV'] === 'production';
```
mark-prins, line 12: I thought it might be useful to be logged out, even when developing, now that the API requires valid credentials for most things, and there isn't a mock server for dev work

packages/common/src/hooks/useAuthContext.ts
joshxg: Nice

Should this check `if (token)`? It's currently setting a header `Authorization: Bearer` which isn't stopping anything, but seems unintentional

packages/common/src/hooks/useAuthContext.ts
```diff
@@ -0,0 +1,54 @@
...
+      const parsed = JSON.parse(authString) as AuthCookie;
+      return parsed;
+    } catch {
+      return emptyCookie;
+    }
+  }
+  return emptyCookie;
+};
+
+export const useAuthContext = () => {
```
joshxg, line 31: Calling this `AuthContext` confused me a bit - I was looking for the `Provider` component - maybe `useAuthState`?

packages/host/src/components/Navigation/RequireAuthentication.tsx
joshxg: Yes, it's actually a little confusing to be logged in and not be logged in!

joshxg (approved):
Nice! Super nice to be able to refresh and be logged in still 😁  

I've made a suggestion about changing `useQuery` -> `useMutation` which I think might work well and seems to suit the flow a nicely, though this might be too much of a change so feel free to ignore

packages/common/src/hooks/useAuthContext.ts
mark-prins: yeah. thanks

packages/common/package.json
```diff
@@ -28,6 +28,7 @@
     "i18next-http-backend": "^1.3.1",
     "i18next-localstorage-backend": "^3.1.3",
     "immer": "^9.0.7",
+    "js-cookie": "^3.0.1",
```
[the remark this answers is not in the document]
mark-prins: no particular reason - small, popular, did what I wanted 🤷

mark-prins:
@joshxg is this what you meant? 34b99af6

wasn't sure about the timeout - sounded like you were wanting it to bypass the need for a refresh token with a sneaky login. perhaps you should adjust if I've got that wrong?

Still needs an inactivity timeout of some kind.. that can come later.

Issue #889, opened 2022-02-23
API Updates: nested union responses have been removed


joshxg:
There are a bunch of changes in the API which mean a bunch of changes we need to make:

The union responses for fields of objects have been removed

PR #891, opened 2022-02-24
#889 nested union responses

joshxg:
#889 

- 🤷


packages/common/src/operations.graphql
```diff
@@ -11,84 +11,38 @@ query itemsWithStockLines(
     sort: { key: $key, desc: $desc }
     filter: $filter
   ) {
-    ... on ConnectorError {
```
joshxg, line 14: Updating the common operations. Main things here are that there is no longer a `ConnectorError` for queries which return array of things. We're still getting an `XResponse` object returned - i.e. `ItemsResponse` for `items` query, but that union has only one member.

That means we definitely get the `ItemConnector` object back, so we don't need to use a guard. However, it means we still keep the `... on ItemConnector` syntax..

So instead of just doing

```
query items {
    totalCount
    nodes { ...}
}
```

```
query items {
    ... on ItemConnector { 
        totalCount
        nodes { ...}
    }
}
```

But effectively what we get returned is the same, so we don't need to worry about checking what type it is anymore (at the moment?)

packages/common/src/operations.graphql
```diff
@@ -159,41 +92,20 @@ query locations($sort: [LocationSortInput!]) {
...
-          }
-        }
-      }
     }
   }
 }
 
-mutation insertLocation($input: InsertLocationInput!) {
-  insertLocation(input: $input) {
+mutation insertLocation($input: InsertLocationInput!, $storeId: String!) {
```
joshxg, line 107: couple of storeIDs added, but not for everything. I believe that's coming a bit later.

packages/common/src/operations.graphql
```diff
@@ -394,77 +312,28 @@ query masterLists(
...
-                    ... on DatabaseError {
-                      __typename
-                      description
-                      fullError
-                    }
-                    description
-                  }
-                }
-                ... on StockLineConnector {
+              availableBatches(storeId: $storeId) {
```
joshxg, line 315: `ConnectorError` gone for nested fields also, but the nested fields just return the `XConnector`, not the `XResponse`.. which means we don't use the `... on XConnector` syntax and we also don't need the guard

packages/common/src/types/index.ts
```diff
@@ -25,13 +25,12 @@ export interface ItemRow extends DomainObject {
   unitName?: string;
 }
 
-export interface Item extends Omit<ItemNode, 'availableBatches'>, DomainObject {
+export interface Item extends ItemNode, DomainObject {
```
joshxg, line 28: Best to just remove this type soon

packages/invoices/src/InboundShipment/DetailView/api.ts
```diff
@@ -201,31 +127,31 @@ export const getInboundShipmentDetailViewApi = (
...
+
+        return {
+          ...line,
+          stockLine,
+          location,
+          stockLineId: stockLine?.id ?? '',
+          invoiceId: invoice.id,
+          unitName: line.item?.unitName ?? '',
+        };
+      });
```
joshxg, line 146: this isn't ideal and it would be good to revisit and pay a little closer attention, probably look at removing the `InvoiceLine` in favour of a fragment

packages/invoices/src/InboundShipment/DetailView/api.ts
```diff
@@ -201,31 +127,31 @@ export const getInboundShipmentDetailViewApi = (
...
-    const lineNodes = linesGuard(invoice.lines);
-    const lines: InvoiceLine[] = lineNodes.map(line => {
-      const stockLine = line.stockLine
-        ? stockLineGuard(line.stockLine)
-        : undefined;
-      const location = line.location ? locationGuard(line.location) : undefined;
-      const item = line.item ? itemGuard(line.item) : undefined;
+    const invoice = result.invoice;
+
+    if (invoice.__typename === 'InvoiceNode') {
```
joshxg, line 132: this is the `invoiceGuard` fn, just inlined as queries for specific entities - i.e. `query invoice` return a union of two members: `NodeError | InvoiceNode` for errors like `RecordNotFound`.

Seems fair enough! The fields however, all are returned as concrete types, so none of the other guards are needed

packages/invoices/src/OutboundShipment/ListView/ListView.tsx
```diff
@@ -17,7 +17,7 @@ import { getStatusTranslator } from '../../utils';
 import { Toolbar } from './Toolbar';
 import { AppBarButtons } from './AppBarButtons';
 import { getOutboundShipmentListViewApi } from './api';
-import { pricingGuard, useOutboundShipmentApi } from '../api';
+import { useOutboundShipmentApi } from '../api';
```
joshxg, line 20: aaand from here down is the same thing: Remove handling of unions for nested fields, remove handling of `ConnectorError`

packages/invoices/src/InboundShipment/ListView/api.ts
```diff
@@ -86,15 +63,8 @@ export const getInboundShipmentListViewApi = (
...
-
-      const invoices = invoicesGuard(result);
-
-      const nodes = invoices.nodes.map(invoice => ({
-        ...invoice,
-        pricing: pricingGuard(invoice.pricing),
-      }));
-
-      return { nodes, totalCount: invoices.totalCount };
+      const invoices = result.invoices;
```
mark-prins, line 66: not 
```
return result.invoices;
```

or

```
const { invoices } = result;
return invoices;
```
? I feel that Will would want me to ask about the destructuring 😉

packages/system/src/MasterList/ListView/ListView.tsx
```diff
@@ -7,13 +7,15 @@ import {
...
   useNavigate,
+  useHostContext,
 } from '@openmsupply-client/common';
 import { Toolbar } from './Toolbar';
 import { getMasterListListViewApi } from './api';
 import { AppBarButtons } from './AppBarButtons';
 import { MasterListRow } from '../types';
 
 export const MasterListListView: FC = () => {
+  const { store } = useHostContext();
```
mark-prins, line 18:
```suggestion
  const { store } = useAuthContext();
```

yeah, I plan to update them all anyway, so not a biggie. more concerned that we'll copy paste the wrong version.  I think the distinction between host and auth concerns is important, better to have the auth version used when using for auth purposes

mark-prins (approved):
😮  that's a lot of change! uh, thanks. sorry you had to wade through all of that!
It has made things simpler though - nice to remove the guards!

codegen.yml
```diff
@@ -1,5 +1,5 @@
 overwrite: true
-schema: ["https://demo-open.msupply.org:8000/graphql", "./packages/common/src/additions.schema.graphql"]
+schema: ["http://localhost:8000/graphql", "./packages/common/src/additions.schema.graphql"]
```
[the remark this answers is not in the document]
joshxg: Yup!

packages/invoices/src/InboundShipment/ListView/api.ts
joshxg: The legacy of william thomson! Thanks 😁

packages/system/src/MasterList/ListView/ListView.tsx
joshxg: Yup, done!

PR #893, opened 2022-02-24
Remove drawer hover expand

### LINKED ISSUE #815: Navigation: hover behaviour
The hover behaviour in the left nav is driving me nuts. Someone should really sort it out.

mark-prins:
Fixes #815 

Removes the dodgy hover / expand function on the drawer. Wasn't working terribly well. 
Took the hover off the menu elements and moved to the drawer itself, no need to be on elements with the move to an onClick for the parent nav items.

Also shifted master lists to Catalogue rather than Inventory as suggested by Richard - made sense to me.

### DIFF
--- a/packages/common/src/styles/RTLProvider.tsx
+++ b/packages/common/src/styles/RTLProvider.tsx
@@ -4,6 +4,10 @@ import { useRtl } from '@common/intl';
 export const RTLProvider: FC = props => {
   const isRtl = useRtl();
 
+  React.useLayoutEffect(() => {
+    document.body.setAttribute('dir', isRtl ? 'rtl' : 'ltr');
+  }, [isRtl]);
+
   return (
     <div
       style={{
--- a/packages/common/src/ui/components/navigation/AppNavLink/AppNavLink.tsx
+++ b/packages/common/src/ui/components/navigation/AppNavLink/AppNavLink.tsx
@@ -3,7 +3,6 @@ import {
   ListItem,
   ListItemIcon,
   ListItemText,
-  Tooltip,
   ListItemButton,
   Box,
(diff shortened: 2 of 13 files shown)


packages/common/src/styles/RTLProvider.tsx
```diff
@@ -4,6 +4,10 @@ import { useRtl } from '@common/intl';
 export const RTLProvider: FC = props => {
   const isRtl = useRtl();
 
+  React.useLayoutEffect(() => {
```
mark-prins, line 7: was looking at the arrow on autocomplete issue and one suggestion was to set the `dir` on body / containing element. don't think it helped, but since it was here I left it in 🤷

packages/common/src/ui/components/navigation/AppNavLink/AppNavLink.tsx
```diff
@@ -91,75 +84,40 @@ export const AppNavLink: FC<AppNavLinkProps> = props => {
...
-        >
-          <ListItemIcon sx={{ minWidth: 20 }}>{icon}</ListItemIcon>
-          <Box className="navLinkText">
-            <Box width={10} />
-            <ListItemText primary={text} />
-          </Box>
-        </ListItemButton>
-      </StyledListItem>
-    </Tooltip>
+    <StyledListItem isSelected={selected} to={to}>
```
mark-prins, line 99: no need for the tooltip now - the drawer expands on hover to show the full width nav items.
was useful for the hover, but that isn't required either now.

joshxg (approved):
I have no useless or useful suggestions - it's amazing, thanks so much for spending the time to do it ❤️ 😀

PR #896, opened 2022-02-24
Add stock level widget to dashboard

### LINKED ISSUE #892: Dashboard: Show number of items with low MOS
Unsure what 'low' is in this context. 3 months for now?
Is it possible to use the `items` query to start with?

mark-prins:
Fixes #892 

Moved the stock query out of the common `operations.graphql` and put in a quick query to show stock levels. The request was to show the number of items with 'low months of stock'. I've made a guess as to what that means.


packages/common/src/styles/RTLProvider.tsx
```diff
@@ -4,12 +4,9 @@ import { useRtl } from '@common/intl';
 export const RTLProvider: FC = props => {
   const isRtl = useRtl();
 
-  React.useLayoutEffect(() => {
-    document.body.setAttribute('dir', isRtl ? 'rtl' : 'ltr');
-  }, [isRtl]);
-
   return (
     <div
+      dir={isRtl ? 'rtl' : 'ltr'}
```
mark-prins, line 9: same effect as adding to the body (probably nothing), just feels better to use declaratively rather than an effect.

packages/common/src/styles/RTLProvider.tsx
```diff
@@ -4,10 +4,6 @@ import { useRtl } from '@common/intl';
 export const RTLProvider: FC = props => {
   const isRtl = useRtl();
 
-  React.useLayoutEffect(() => {
```
mark-prins, line 7: I now think it is the direction being set to `rtl` in the theme.. and this has no effect

joshxg: are you able to repro? it's still just random for me 🙄

packages/dashboard/src/widgets/StockWidget.tsx
```diff
@@ -2,24 +2,31 @@ import React from 'react';
...
+  const lowStockItemsCount =
+    itemStatsData?.filter(
+      item =>
+        item.stats.availableStockOnHand > 0 &&
+        item.stats.availableMonthsOfStockOnHand < LOW_MOS_THRESHOLD
+    ).length || 0;
+
+  const noStockItemsCount =
+    itemStatsData?.filter(item => item.stats.availableStockOnHand === 0)
+      .length || 0;
```
joshxg, line 26: Could have another `useItemCounts` hook and put this in there.. though that's pretty excessive 😁

joshxg (approved):
It's great 🎉 

What about a little button that for now just navigated to `internal-orders` ?

image

hehe

packages/dashboard/src/widgets/StockWidget.tsx
mark-prins: I was thinking to have the hook do the work.. was hoping this might be temporary anyway - something that the back end could simply return 🤷

mark-prins:
> are you able to repro? it's still just random for me 🙄

the closest I've got is changing the direction in the theme - setting that to `ltr` when using Arabic makes the autocomplete behave strangely. though different to the original problem 🙄

mark-prins:
heh - good suggestion with the order more button! I've popped in a chunky one at the bottom.. to keep the design consistent. I don't expect it to stay though..

PR #897, opened 2022-02-25
#895 internal order columns

### LINKED ISSUE #895: Supplier requisition columns
- The `RequisitionLineNode` type differs from the mock API so the columns are showing the correct values anymore so this should be updated.
- The column headers: Used the headers from zepplin. Also remember craig saying to just use the column header names from mSupply. Not sure if these should change. However, they've also always looked average on smaller screens as they're quite verbose and end up overlapping.. unsure what to do about this exactly.. maybe removing the `forecast method` column will fix all the problems?

joshxg:
Fixes #895

image

- Have updated the columns for request and the item columns for responses. I'm not sure about the columns for responses I don't think the spreadsheet is accurate https://docs.google.com/spreadsheets/d/1H0Rv1DEOgW0cE07wpfc_KivLf8TcNhYDqlwRp92x-To/edit#gid=320810900&range=A233:S233 and haven't seen any mock ups 🤔

### DIFF
--- a/packages/common/src/intl/locales/en/common.json
+++ b/packages/common/src/intl/locales/en/common.json
@@ -113,6 +113,7 @@
   "label.selected": "Selected",
   "label.sell": "Sell",
   "label.shipped": "Shipped",
+  "label.months": "Months",
   "label.status": "Status",
   "label.supplier": "Supplier",
   "label.total": "Total",
--- a/packages/common/src/intl/locales/en/common.json
+++ b/packages/common/src/intl/locales/en/common.json
@@ -113,7 +113,7 @@
   "label.selected": "Selected",
   "label.sell": "Sell",
   "label.shipped": "Shipped",
-  "label.months": "Months",
+  "label.month": "month",
   "label.status": "Status",
   "label.supplier": "Supplier",
   "label.total": "Total",
--- a/packages/common/src/ui/layout/tables/utils/ColumnDefinitionSetBuilder.ts
+++ b/packages/common/src/ui/layout/tables/utils/ColumnDefinitionSetBuilder.ts
@@ -39,46 +39,19 @@ export type ColumnKey =
   | 'stocktakeNumber'
   | 'description'
   | 'stocktakeDatetime'
-  | 'monthlyConsumption'
-  | 'previousStockOnHand'
-  | 'calculatedQuantity'
-  | 'requestedQuantity'
-  | 'forecastMethod';
+  | 'monthlyConsumption';
 
 const getColumnLookup = <T extends DomainObject>(): Record<
   ColumnKey,
   ColumnDefinition<T>
 > => ({
-  requestedQuantity: {
-    key: 'requestedQuantity',
-    label: 'label.requested-quantity',
-    align: ColumnAlign.Right,
-    width: 100,
-  },
   monthlyConsumption: {
     key: 'monthlyConsumption',
     label: 'label.average-monthly-consumption',
     align: ColumnAlign.Left,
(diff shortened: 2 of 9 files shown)


packages/common/src/ui/layout/tables/utils/ColumnDefinitionSetBuilder.ts
```diff
@@ -39,46 +39,19 @@ export type ColumnKey =
...
-    align: ColumnAlign.Left,
-    width: 125,
-  },
-  forecastMethod: {
-    key: 'forecastMethod',
-    label: 'label.forecast-method',
-    align: ColumnAlign.Left,
-    width: 55,
-  },
+
```
joshxg, line 54: I removed these columns as they're only really used in request reqs, so doesn't seem useful to put them in a shared space

packages/common/src/utils/NumUtils.ts
```diff
@@ -0,0 +1,5 @@
+export const NumUtils = {
+  isPositive: (num: number): boolean => {
+    return num > 0;
+  },
+};
```
joshxg, line 5: Is this silly?

Idea being colocation of general number utils 🤷

mark-prins: it's a good pattern - not silly I think.
I do wonder about moving the 'is low months of stock' into here actually - could be useful to have in one place

packages/common/src/intl/locales/en/common.json
```diff
@@ -113,6 +113,7 @@
   "label.selected": "Selected",
   "label.sell": "Sell",
   "label.shipped": "Shipped",
+  "label.months": "Months",
```
mark-prins, line 116:
```suggestion
  "label.months_one": "month",
  "label.months_other": "months",
```

to allow for plurals ( and lowercase the months 😉  )

packages/requisitions/src/RequestRequisition/DetailView/columns.ts
```diff
@@ -17,44 +20,72 @@ export const useRequestRequisitionColumns = ({
...
+        width: 200,
+        accessor: ({ rowData }) => {
+          const { itemStats } = rowData;
+          const { availableStockOnHand, availableMonthsOfStockOnHand } =
+            itemStats;
+
+          const monthsString = availableMonthsOfStockOnHand
+            ? `${availableMonthsOfStockOnHand} ${t('label.months')}`
+            : '';
+          return `${availableStockOnHand} ${monthsString}`;
```
mark-prins, line 58: should that have brackets around it?
```suggestion
          return `${availableStockOnHand} (${monthsString})`;
```

packages/requisitions/src/RequestRequisition/DetailView/columns.ts
```diff
@@ -17,44 +20,72 @@ export const useRequestRequisitionColumns = ({
...
+
+        width: 200,
+        accessor: ({ rowData }) => {
+          const { itemStats } = rowData;
+          const { availableStockOnHand, availableMonthsOfStockOnHand } =
+            itemStats;
+
+          const monthsString = availableMonthsOfStockOnHand
+            ? `${availableMonthsOfStockOnHand} ${t('label.months')}`
+            : '';
```
mark-prins, line 57:
```suggestion
            ? `${availableMonthsOfStockOnHand} ${t('label.months', {
                count: availableMonthsOfStockOnHand,
              })}`
            : '';
```

mark-prins (approved):
🎉  awesome, looks good

PR #898, opened 2022-02-25
Set theme direction

### LINKED ISSUE #886: Autocomplete caret has jumped over the fence to the other side
The little caret arrow icon of the autocomplete component has jumped onto the left hand side!

image
image

What's that about?

mark-prins:
Fixes #886 ( maybe!! )

Setting the direction to `ltr` and reloading in Arabic makes the autocomplete behave incorrectly in a different way ( the caret is incorrect and moves when the select is expanded ).. so I figure this is similar.

The actual problem was when using an ltr language, the mui classes `rtl-` were applied to all the things. so there was something amiss with the mui theme. Hoping this helps 🤞

### DIFF
--- a/packages/common/src/styles/theme.ts
+++ b/packages/common/src/styles/theme.ts
@@ -113,7 +113,7 @@ export const themeOptions = {
       xl: 1537,
     },
   },
-  direction: 'rtl' as Direction,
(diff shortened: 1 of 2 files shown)


joshxg (approved):
Yup LGTM 😁

PR #899, opened 2022-02-25
Prompt for unsaved shipments

### LINKED ISSUE #498: Warn the user when navigating a way from an invoice that hasn't been saved
When navigating away from an invoice which needs to be saved, stop the user and warn them.

- Will require managing some sort of state for knowing when the invoice needs to be saved

mark-prins:
Fixes #498 

There is an issue when unloading the event listener.. for me at least. It doesn't - so once the listener is set, if you refresh the page you are warned, even though you reset the lines back to their original state. This isn't a big deal, but it bugs me.

Other issues

- I found that the use of storeId in `useQueryParams` was problematic, as it cached the value and caused problems when switching stores and trying to test this thing.
- Had trouble running storybook until I upgraded react-refresh-webpack-plugin 🤷


packages/common/src/hooks/useListData/useListData.ts
```diff
@@ -57,8 +55,9 @@ export const useListData = <T extends DomainObject>(
   onError?: (e: ClientError) => void
 ): ListDataState<T> => {
   const queryClient = useQueryClient();
-  const { filterBy, queryParams, first, offset, sortBy, storeId } =
-    useQueryParams(initialListParameters);
+  const { filterBy, queryParams, first, offset, sortBy } = useQueryParams(
```
mark-prins, line 58: so I went a big crazy and did some of these, so I could test stuff.

packages/common/src/ui/components/navigation/NavigationPrompt/NavigationPrompt.tsx
```diff
@@ -0,0 +1,35 @@
...
+  useEffect(() => {
+    if (isUnsaved) {
+      window.addEventListener('beforeunload', promptUser, { capture: true });
+    } else {
+      window.removeEventListener('beforeunload', promptUser, { capture: true });
+    }
+  }, [isUnsaved]);
+
+  return <></>;
+  // Prompt, usePrompt and useBlocker are not currently implemented in the current v6 release
```
mark-prins, line 27: it is in v5 - they said they did not include in v6 in order to get the release out.
I don't think it matters - in our use case, you cannot navigate without closing the modal which cancels changes. So the window event is the only thing which can catch the navigation anyway.

packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEdit.tsx
```diff
@@ -98,12 +102,18 @@ const useDraftInboundLines = (itemId: string) => {
     }
   }, [lines, itemId]);
 
+  useEffect(() => {
+    if (!initialLines && draftLines.length > 0) {
+      setInitialLines([...draftLines]);
+    }
+  }, [draftLines]);
```
joshxg, line 109: Could be simpler to just use a ref

```
const isDirty = useRef(false)
...
```

in the `updateDraftLine` and `addBatch` functions

```
isDirty.current = true
```

I think?

joshxg: If that does work - maybe a custom hook in common `const {isDirty, markDirty } = useDirtyCheck()` or something?

packages/common/src/ui/components/navigation/NavigationPrompt/NavigationPrompt.tsx
```diff
@@ -0,0 +1,35 @@
+import React, { useEffect } from 'react';
+// import { Prompt } from 'react-router-dom';
+
+export interface NavigationPromptProps {
+  isUnsaved: boolean;
+}
+
+export const NavigationPrompt: React.FC<NavigationPromptProps> = ({
```
joshxg, line 8: Is this a component because `react-router-dom` might eventually provide a component and so it would make it easy to change later?

I wondered if it would be a little easier if this was a hook

```
const { isDirty, ..., ..., ..., } = useDraftOutboundLines();
useAreYouSureYouWantToLeave(isDirty)
```

something like that? 🤷

packages/common/src/ui/components/navigation/NavigationPrompt/NavigationPrompt.tsx
```diff
@@ -0,0 +1,35 @@
...
+  isUnsaved,
+}) => {
+  const promptUser = (e: BeforeUnloadEvent) => {
+    // Cancel the event
+    e.preventDefault(); // If you prevent default behavior in Mozilla Firefox prompt will always be shown
+    // Chrome requires returnValue to be set
+    e.returnValue = '';
```
joshxg, line 15: nice! this is awesome

joshxg (approved):
Nice!

packages/common/src/ui/components/navigation/NavigationPrompt/NavigationPrompt.tsx
mark-prins: it is - if there is a situation where there isn't a blocking modal, like we currently have, then the navigation with `react-router` won't trigger the browser prompt. I don't know when that might happen, or if we'll actually need it. I'm thinking now that I'll use a hook and comment to warn people instead

packages/invoices/src/InboundShipment/DetailView/modals/InboundLineEdit/InboundLineEdit.tsx
mark-prins: the intention was to check the condition of returning to the original state. if you changed, then reverted, you shouldn't get prompted. Given that I've not been able to remove the confirmation once set, this is a bit pointless though

joshxg: Ah! Of course!

Also by using a state variable, we can react to the 'dirtiness' if ever needed - i.e. if we needed to enable a 'save' button or show a 'not saved' / 'saved' label

packages/invoices/src/OutboundShipment/DetailView/modals/hooks/useDraftOutboundLines.tsx
```diff
@@ -104,10 +108,9 @@ export const useDraftOutboundLines = (
     if (!data) return;
 
     setDraftOutboundLines(() => {
-      if (!lines) return [];
       const rows = data
         .map(batch => {
-          const invoiceLine = lines.find(
+          const invoiceLine = (lines ?? []).find(
```
mark-prins, line 113: this and the above - part of a fix for no stocklines showing, when I know that there are some!

packages/system/src/Item/api/hooks/useItem/useItem.ts
```diff
@@ -9,19 +9,19 @@ import { ItemQueries } from './../../api';
 // TODO: Use itemID to filter when possible.
 export const useItem = (itemCode: string): UseQueryResult<Item> => {
   const api = useItemApi();
-  const { filterBy, filter, queryParams, first, offset, sortBy, storeId } =
+  const { storeId } = useAuthState();
+  const { filterBy, filter, queryParams, first, offset, sortBy } =
```
mark-prins, line 13: was using an incorrect `storeId` and as a result, not showing me my stock

joshxg:
Awesome, sorry about the conflicts 😁

PR #902, opened 2022-02-25
#590 Response req line 1

### LINKED ISSUE #590: Add/Edit item functionality for customer requisition
Epic for the add/edit item functionality for a customer requisition

Currently you add an item like:

image

- There is no edit modal.
- You can edit the cells in the main list view

Personally I find the inconsistency of which detail view has editable cells and which has modals etc quite confusing so would prefer to just stick with click to edit 🤷

joshxg:
Fixes #590

- Just adds the bulk of the response requisition line edit modal
- This is quite a bit simpler than others, as we don't need to account for adding items, only editing. Also since there is only one line per item, this is a lot easier! So we can just deal with requisition items


packages/requisitions/src/ResponseRequisition/DetailView/AppBarButtons.tsx
```diff
@@ -1,34 +1,16 @@
...
 
-interface AppBarButtonProps {
-  onAddItem: (newState: boolean) => void;
-}
-
-export const AppBarButtonsComponent: FC<AppBarButtonProps> = ({
-  onAddItem,
-}) => {
-  const isDisabled = useIsResponseRequisitionDisabled();
+export const AppBarButtonsComponent: FC = () => {
```
joshxg, line 8: No adding items, so removed

packages/requisitions/src/ResponseRequisition/DetailView/DetailView.tsx
```diff
@@ -23,15 +22,15 @@ import {
 import { ResponseLineEdit } from './ResponseLineEdit';
 
 export const DetailView: FC = () => {
-  const { onOpen, onClose, mode, entity, isOpen } =
-    useEditModal<ItemRowFragment>();
+  const { onOpen, onClose, entity, isOpen } =
+    useEditModal<ResponseRequisitionLineFragment>();
```
joshxg, line 26: Since we don't need to add items we can just use a req line

packages/requisitions/src/ResponseRequisition/DetailView/ResponseLineEdit/ResponseLineEditForm.tsx
```diff
@@ -1 +1,151 @@
...
+  Right,
+}: ResponseLineEditFormLayoutProps) => {
+  return (
+    <Grid
+      container
+      spacing={2}
+      direction="row"
+      justifyContent="space-between"
+      bgcolor="background.toolbar"
+      padding={2}
```
mark-prins, line 61:
```suggestion
      padding={4}
```
is more like the zeplin version - looks better with a bit more padding to me

packages/requisitions/src/ResponseRequisition/api/hooks.ts
```diff
@@ -54,7 +56,7 @@ export const useResponseRequisition =
     const { requisitionNumber = '' } = useParams();
     const { store } = useHostContext();
```
mark-prins, line 57: could you change to `useAuthContext` please?

mark-prins (approved):
👍  looking good thanks!

PR #913, opened 2022-02-27
#882 remaining operations from common

### LINKED ISSUE #882: Shift remaining operations from common
There are only a couple of operations left in the `common` package operations file.

- Locations
- MasterList
- itemsWithStock
- names
- stockCounts
- authToken
- stores

To me, these have clear places to live: in an operations file in the packages for the feature.

There should be more API changes coming - in particular adding `storeID` to a few of these queries. It would be good when those changes land to shift these queries

joshxg:
Fixes #882  (except not completely)

- Bit big!
- Shifts all the name, store, stock and master list queries out of common and into their own packages
- This is all system data which is read only and so using the `useListData` hook wasn't useful as the interface for that had 3/4 queries as stubs
- The types could still be worked on but generally was going to be a bit more work - thought to just scope this specifically to be the same thing for each of the features - moving the query, creating the pattern we've been using elsewhere
- Location still to do - it's a bit more work as it's store data
- AuthToken is also left. Personally think an `auth` package is the way to go..  but possibly we could move the query into `host`? Or just leave it in common


packages/system/src/Name/api/hooks.ts
```diff
@@ -0,0 +1,55 @@
...
+  const { client } = useOmSupplyApi();
+  const queries = getNameQueries(getSdk(client), storeId);
+  return { ...queries, storeId };
+};
+
+export const useNamesSearch = ({
+  isCustomer,
+}: {
+  isCustomer?: boolean;
+  isSupplier?: boolean;
```
mark-prins, line 23: not required?

mark-prins (approved):
👍 another biggie! thanks tho - that's looking much more consistent and overall a lot clearer.

PR #914, opened 2022-02-27
#882 Location query relocation

### LINKED ISSUE #882: Shift remaining operations from common

joshxg:
Fixes #882 

- Just the location queries out of common


mark-prins (approved):
🎉  looks great!
One suggestion - is it possible to focus the `Name` field on the edit modal on load? When adding a new one it's a bit annoying..

PR #915, opened 2022-02-28
#494 Add foreign currency support

### LINKED ISSUE #494: Support for different currencies and number formats
Need to add support for different currencies: prefixes mainly? Or just don't use prefixes anywhere 😁 

Also need to add support for `,` v `.` in numbers. 

These maybe are separate issues. Wanted to make an issue as to not forget!

joshxg:
Fixes #494 

- I think I'm ok with this, but.. undecided 😁 
- Have added intl helpers for `useCurrency` / `useCurrencyFormat` similar to the other format hooks. This allows for the currency config to update with language state changes, though it might be nice to be able to work with currencies and have the option update outside of react, I wasn't sure of the need right now
- Update currency input quite a lot.. think it's simplified and easier to use
- Don't remember any other money values anywhere 🤔


packages/common/src/intl/intlHelpers.ts
```diff
@@ -24,6 +26,81 @@ export const useTranslation = (ns?: Namespace): TypedTFunction<LocaleKey> => {
...
+// would be 12.00. With a precision of 2, then currency values would be
+// truncated to be a maximum of two decimal places.
+const formatter: currency.Format = (currency?, options?): string => {
+  const { decimal = '.' } = options ?? {};
+  if (!currency) return '';
+  if (!currency.cents()) {
+    return `${currency.value}${decimal}${'0'.repeat(2)}`;
+  }
+  return String(parseFloat(String(currency.value)));
+};
```
joshxg, line 42: I don't know if this is needed, but wanting to handle the case of tiny cent values, but also have normal looking money values

packages/common/src/intl/intlHelpers.ts
```diff
@@ -24,6 +26,81 @@ export const useTranslation = (ns?: Namespace): TypedTFunction<LocaleKey> => {
...
+  },
+  ar: {
+    symbol: 'ر.ق.',
+    separator: ',',
+    decimal: '.',
+    precision: 10,
+    pattern: '!#',
+    negativePattern: '-!#',
+    format: formatter,
+  },
```
joshxg, line 71: seems that if we were to use `intl` we would need *something* like this anyway, but would also maybe need to do more work. I.e. I can't get `Intl` to format with a `XOF` suffix, so might need a customer`format` method to handle that.

I don't know what the future holds for currencies, but I imagine either having these functions also convert to the actual currency from some base, or only displaying a single currency per deployment, in which case this config could come from the server on login

packages/common/src/ui/components/inputs/CurrencyInput/CurrencyInput.tsx
```diff
@@ -24,49 +25,18 @@ const StyledCurrencyInput = styled(RCInput)(({ theme }) => ({
...
-  // to a number. In this case, we want to keep the value as '42.' in the input
-  // but not persisted. Though, whenever I have an uncontrolled component I
-  // generally will always regret it, so sync the passed value up with the buffer.
-  const [buffer, setBuffer] = React.useState<string | undefined>(String(value));
-  useEffect(() => {
-    if (value !== Number(buffer)) {
-      setBuffer(String(value));
-    }
-  }, [value]);
+  const { c, options, language } = useCurrency();
```
joshxg, line 36: I think we're preferring not to use `intl` in `ui`, but I didn't see another way of encapsulating this in the component. I really didn't think any consumers of the component should really have to handle all this stuff

packages/common/src/intl/intlHelpers.ts
```diff
@@ -24,6 +26,81 @@ export const useTranslation = (ns?: Namespace): TypedTFunction<LocaleKey> => {
...
+  fr: {
+    symbol: 'XOF',
+    separator: '.',
+    decimal: ',',
+    precision: 10,
+    pattern: '# !',
+    negativePattern: '-!#',
```
mark-prins, line 60: I would have thought `-# !` to be consistent

packages/common/src/intl/intlHelpers.ts
mark-prins: when using `intl` this wouldn't be required for most locales - only the edge cases. 
It is a lot more painful to customise when required though, and so this is better long-term I think.

My main concern is that the options are country-specific and not language-specific, as they are here. We do need a way to have the a language translation set to be used, with a different currency format option.

packages/common/src/intl/intlHelpers.ts
mark-prins: Some tests would be helpful to see it in action

packages/common/src/ui/layout/tables/components/Cells/CurrencyInputCell/CurrencyInputCell.tsx
```diff
@@ -23,7 +23,7 @@ export const CurrencyInputCell = <T extends DomainObject>({
     <CurrencyInput
       autoFocus={autoFocus}
       maxWidth={column.width}
-      value={buffer}
+      defaultValue={buffer}
```
mark-prins, line 26: this does look pretty bad with `xof` in it - any thoughts about having the currency symbol outside the input?

Screen Shot 2022-03-01 at 4 49 06 PM

Screen Shot 2022-03-01 at 4 49 55 PM

Screen Shot 2022-03-01 at 4 50 12 PM

Looks ok with a numeric input too..

Screen Shot 2022-03-01 at 4 45 39 PM

Screen Shot 2022-03-01 at 4 46 41 PM

mark-prins (approved):
👍 

I jumped in to see what the shipments would look like.. any thoughts about doing those at the same time?

Outbounds:

Screen Shot 2022-03-01 at 4 26 09 PM

Screen Shot 2022-03-01 at 4 26 25 PM

packages/common/src/ui/layout/tables/components/Cells/CurrencyInputCell/CurrencyInputCell.tsx
joshxg: I've no preference either way. Do you have a branch with the changes you have in the pics? Can we just commit that 🤷

packages/common/src/intl/intlHelpers.ts
joshxg: oops

packages/common/src/intl/intlHelpers.ts
joshxg: oops - I wrote some and they were failing as this was silly. Not quite as simplistic as that 😁 

To get to where I wanted it would require a bit more work - specifically need to get the pattern from `options` and do a regex replace with the value, symbol and also handle the group separator and decimal 🙃

joshxg:
I've updated outbound.

Some out standings, but will get this in for now:

- Better use of `useColumns` default format options - specifically, use `useCurrency` or some other way when the column format is currency.
- Put currency symbols outside of inputs
- Use a custom formatter to be able to use larger precisions

PR #916, opened 2022-02-28
#888 refresh token

### LINKED ISSUE #888: Refresh token
Following on from #838 - implement the refresh token and session timeout

thinking something like this to refresh the token regularly:
```
const { mutate: refresh, data: refreshResponse } = useMutation<
    AuthenticationResponse,
    unknown,
    string,
    unknown
  >(async token => {
    const result = await api.refreshToken(token);
    return refreshTokenGuard(result);
  });

  const login = (
    credentials: LoginCredentials,
    options: MutateOptions<
      AuthenticationResponse,
      unknown,
      LoginCredentials,
      unknown
    >
  ) => {
    if (timeoutRef.current) clearTimeout(timeoutRef.current);
    mutate(credentials, options);
    timeoutRef.current = setTimeout(() => {
      refresh(rest.data?.token ?? '');
    }, REFRESH_INTERVAL);
  };
  return { login, ...rest };
```

though that probably isn't needed. The expired token response will need to be caught for all requests, and the token refreshed. The above only helps when continuing a session from a login request. Refreshing the page or opening a new tab will require something else.

mark-prins:
Fixes #888 

Having trouble testing the refresh as the strict same site policy means that I cannot test locally against demo server, and when running locally I'm having issues with my local certs. 
The refresh mutation is being sent, but failing due to the lack of refresh token cookie.


packages/common/src/api/OmSupplyApiContext.tsx
```diff
@@ -6,7 +6,7 @@ import { getSdk } from '../operations.generated';
 export const createOmSupplyApi = (
   url: string
 ): { api: OmSupplyApi; client: GraphQLClient } => {
-  const client = new GraphQLClient(url);
+  const client = new GraphQLClient(url, { credentials: 'same-origin' });
```
mark-prins, line 9: include same origin cookies in the request

packages/common/src/authentication/AuthContext.tsx
```diff
@@ -0,0 +1,105 @@
+import React, { createContext, FC, useEffect, useMemo, useRef } from 'react';
```
mark-prins, line 1: the `useState` hook, but in Context form. and then the refresh added in

packages/common/src/authentication/AuthContext.tsx
```diff
@@ -0,0 +1,105 @@
...
+};
+
+const AuthContext = createContext<AuthControl>({
+  onLoggedIn: () => {},
+});
+
+const { Provider } = AuthContext;
+
+export const AuthProvider: FC = ({ children }) => {
+  const { setStore, setUser } = useHostContext();
```
joshxg, line 71: thinking we could just use `useAuthContext` for these values rather than duplicate them in the host context?

packages/common/src/authentication/AuthContext.tsx
joshxg: nice, it's awesome!

joshxg (approved):
Nice! you're right it's hard to test.. maybe we can ask the guys if they can make it easier - maybe not using a secure cookie 🤷

packages/common/src/authentication/AuthContext.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const [, setMRUCredentials] = useLocalStorage('/mru/credentials');
```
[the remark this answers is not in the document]
mark-prins: neat. thanks

packages/common/src/authentication/AuthContext.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const { expires, token } = getAuthCookie();
+  const cookieLifetimeInMinutes = differenceInMinutes(
+    new Date(expires ?? ''),
+    new Date()
+  );
+  const timeoutInterval =
+    (cookieLifetimeInMinutes > 0 ? cookieLifetimeInMinutes : 0.3) * 60 * 1000;
+
+  React.useEffect(() => {
+    if (token) {
```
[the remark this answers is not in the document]
mark-prins: I thought it would reschedule because the token changes - can't say for certain because I still don't have fetch sending the cookie!!
It is a very neat idea! My concern is that we refresh immediately on login and that's not very neat. Only a minor concern, I'll change anyway..

mark-prins:
implemented suggestions; merged main back in and have the refresh token cookie working. The refresh query was failing locally.. let's get this in and retry.

PR #919, opened 2022-02-28
#917 shipment number for navigating

### LINKED ISSUE #917: shipments should use shipment numbers for queries not IDs
Currently using `id` for navigating to detail views. Should use `shipmentNumber`

joshxg:
Fixes #917  (Inbound only)

- Ended up going quite offroad on this one..
- Shifts all the api stuff to be consistent with everything else
- Then makes the small change the issue wanted for using shipment numbers for navigation rather than IDs


packages/invoices/src/InboundShipment/ListView/ListView.tsx
```diff
@@ -3,55 +3,38 @@ import {
...
-    getInboundShipmentListViewApi(api, storeId)
-  );
+  } = useInbounds();
+
   const t = useTranslation();
 
   const columns = useColumns<InvoiceRow>(
     [
-      [getNameAndColorColumn(), { setter: onUpdate }],
+      [getNameAndColorColumn(), { setter: () => {} }],
```
joshxg, line 37: Will make an issue for this - similar is done for the requisition list views, so will do them all

packages/invoices/src/InboundShipment/api/api.ts
```diff
@@ -1,3 +1,261 @@
-import { getSdk } from './operations.generated';
+import {
```
joshxg, line 1: these are all copy pastes from where they lived, grouped and renamed

packages/invoices/src/InboundShipment/api/hooks.ts
```diff
@@ -1,7 +1,255 @@
-import { useOmSupplyApi } from '@openmsupply-client/common';
+import { useCallback } from 'react';
```
joshxg, line 1: these have also been copy pasted from where they lived, renamed and changed slightly to use the `api` object and `invoiceNumber` rather than `id`

packages/invoices/src/InboundShipment/api/operations.graphql
```diff
@@ -209,6 +209,34 @@ query invoice($id: String!, $storeId: String!) {
   }
 }
 
+query inboundByNumber($invoiceNumber: Int!, $storeId: String!) {
+  invoiceByNumber(
+    invoiceNumber: $invoiceNumber
+    storeId: $storeId
+    type: INBOUND_SHIPMENT
```
joshxg, line 216: invoice list views aren't filtering by their type 🤣

mark-prins (approved):
👍  very nice to have the number in the URL - that's much more usable!

Issue #911, opened 2022-02-26
Invoice type cleanup


joshxg:
Requisitions and stocktakes are using GQL fragments - invoices aren't.  Would be good to consolidate

- Use `OutboundFragment`/`OutboundRowFragment`/`OutboundLineFragment` (and same for inbound)
- Start to use `OutboundLineFragment.item` rather than `itemId`/`itemCode`/`itemName` to be consistent

PR #927, opened 2022-03-01
#911 invoice type cleanup

joshxg:
#911 (Outbound only)

- Removes the use of the extra types in outbound and use generated fragment types instead


packages/invoices/src/InboundShipment/api/operations.graphql
```diff
@@ -161,6 +80,81 @@ fragment InboundShipmentLine on InvoiceLineNode {
   }
 }
 
+fragment InboundShipment on InvoiceNode {
```
joshxg, line 83: Not using these in this PR but added them 😁

packages/invoices/src/OutboundShipment/DetailView/ContentArea.tsx
```diff
@@ -8,16 +8,17 @@ import {
   MiniTable,
   useIsGrouped,
 } from '@openmsupply-client/common';
-import { InvoiceLine, InvoiceItem } from '../../types';
+import { OutboundItem } from '../../types';
```
joshxg, line 11: essentially just replacing the `InvoiceItem` with `OutboundShipmentLineFragment` (Mayeb that should be `OutboundLineFragment` 🤔 

This is so we don't have to manage the `InvoiceLine` type which is a pain. An example of annoyance is that because it extends `InvoiceLineNode`, the `item` field is an `ItemNode`.. so that means it has the required fields `stats` and `availableBatches`, when all you really want is a more simple subset

packages/invoices/src/OutboundShipment/DetailView/modals/OutboundLineEditTable.tsx
```diff
@@ -50,7 +50,10 @@ const BatchesRow: React.FC<BatchesRowProps> = ({
         <NumericTextInput
           onChange={event => {
             const value = Math.max(Number(event.target.value), 0);
-            const newValue = Math.min(value, batch.availableNumberOfPacks);
+            const newValue = Math.min(
+              value,
+              batch.stockLine?.availableNumberOfPacks ?? 0
```
joshxg, line 55: Previously, the `DraftOutboundLine`  was extending `InvoiceLine` which had no `stockLine` or `item` fields, specifically because of the problem above - having annoying types with a bunch of required fields we didn't actually want or need.

Now, the draft type is extending `PartialOutboundLineFragment` type, which has subset of fields on `stockLine` and `item` - the ones that are needed.

So, instead of creating a `DraftOutboundLine` and adding in a bunch of fields and reassigning (i.e. putting `availableNumberOfPacks` on `DraftOutboundLine` and taking it from the `stockLine`, we can just use the `stockLine` directly

packages/invoices/src/OutboundShipment/DetailView/modals/hooks/useDraftOutboundLines.tsx
```diff
@@ -3,81 +3,90 @@ import {
...
 
 interface DraftOutboundLineSeeds {
   invoiceId: string;
-  stockLine?: Partial<StockLine> | null;
-  invoiceLine?: Partial<InvoiceLine> | null;
+  invoiceLine: PartialOutboundLineFragment;
 }
 
-export const createDraftOutboundLine = ({
+export const createDraftOutboundLineFromStockLine = ({
```
joshxg, line 40: I attempted to simplify these functions.

- When creating a draft line, it's either from a stock line (when there is no existing invoice line), or an invoice line
- In the case of a stock line, we need to create an invoice line, which is relatively straight forward
- In the case of an existing invoice line, we just need to clone it (with the exception of available number of packs)

So I split this into two functions and it feels far easier to understand

packages/invoices/src/OutboundShipment/DetailView/modals/hooks/useNextItem.test.tsx
```diff
@@ -137,10 +137,10 @@ describe('useNextItem', () => {
   });
 
   it('eventually equals an object where the next item is equal to the next item in sorted order.', () => {
-    const invoice = getInvoice();
-    invoice.lines.nodes = getLines();
-    const handler = graphql.query('invoice', (_, res, ctx) => {
-      return res(ctx.data({ invoice }));
+    const invoiceByNumber = getInvoice();
```
joshxg, line 140: test updates from using `invoiceByNumber` rather than `invoice` query in a older PR

packages/invoices/src/OutboundShipment/DetailView/modals/hooks/usePackSizeController.test.tsx
```diff
@@ -3,153 +3,177 @@ import { createDraftOutboundLine } from './useDraftOutboundLines';
 import { DraftOutboundLine } from 'packages/invoices/src/types';
 import { usePackSizeController } from './usePackSizeController';
 import { act } from '@testing-library/react';
-
-const singlePackSizeLines: DraftOutboundLine[] = [
+import { InvoiceLineNodeType } from '@openmsupply-client/common';
```
joshxg, line 6: test updates from the types changing

packages/invoices/src/OutboundShipment/DetailView/modals/hooks/usePackSizeController.tsx
```diff
@@ -13,8 +13,8 @@ const distinctSortedPackSizes = (lines: DraftOutboundLine[]): number[] =>
     new Set(
       lines
         .filter(
-          ({ onHold, availableNumberOfPacks }) =>
-            availableNumberOfPacks > 0 && !onHold
+          ({ stockLine }) =>
```
joshxg, line 16: updates from fields being on the nested `stockLine` rather than the draft line itself

packages/invoices/src/OutboundShipment/DetailView/modals/utils.ts
```diff
@@ -40,8 +40,8 @@ export const sumAvailableQuantity = (
   draftOutboundLines: DraftOutboundLine[]
 ) => {
   const sum = draftOutboundLines.reduce(
-    (acc, { availableNumberOfPacks, packSize }) =>
-      acc + availableNumberOfPacks * packSize,
+    (acc, { stockLine, packSize }) =>
```
joshxg, line 43: updates from fields being on the nested `stockLine` rather than the draft line itself

packages/invoices/src/OutboundShipment/DetailView/modals/utils.test.ts
```diff
@@ -1,87 +1,106 @@
+import {
```
joshxg, line 1: Updates for the new types

packages/invoices/src/OutboundShipment/api/api.ts
```diff
@@ -47,7 +48,7 @@ const outboundParsers = {
       }
     }
   },
-  toStatus: (patch: RecordPatch<Invoice>) => {
```
joshxg, line 50: also removed the use of the `Invoice` type (previously had removed most uses) as it has the same problems as the `InvoiceLine` type

packages/invoices/src/OutboundShipment/api/operations.graphql
```diff
@@ -1,3 +1,55 @@
+fragment PartialLocation on LocationNode {
```
joshxg, line 1: I'm thinking I should remove these fragments and just inline them as we're not actually using the generated types

packages/invoices/src/OutboundShipment/api/operations.graphql
```diff
@@ -1,3 +1,55 @@
...
+}
+
+fragment PartialItem on ItemNode {
+  id
+  name
+  code
+  unitName
+}
+
+fragment PartialOutboundLine on InvoiceLineNode {
```
mark-prins, line 29: 🤔  We have `PartialOutboundLineFragment` and `OutboundShipmentLineFragment`
Could one replace the other? 
Should `OutboundShipmentLineFragment` be renamed to `OutboundLineFragment`?
Does `Partial...` indicate a subset or is it a new naming convention for fragments?

packages/invoices/src/types.ts
```diff
@@ -1,5 +1,10 @@
 import {
-  Column,
+  OutboundShipmentLineFragment,
+  PartialOutboundLineFragment,
+  // OutboundShipmentFragment,
```
mark-prins, line 4: is the `OutboundShipmentFragment` needed?

packages/system/src/Item/utils.ts
```diff
@@ -50,7 +50,5 @@ export const toItem = (line: ItemLike): Item => ({
     averageMonthlyConsumption: 0,
     availableStockOnHand: 0,
   },
-  unitName:
-    ('lines' in line ? line.lines[0].item?.unitName : line.item?.unitName) ??
-    '',
+  unitName: '',
```
mark-prins, line 53: did you have thoughts about how to bring the unit back?
The inconsistency of `itemName` `itemCode` etc and the `item` as an object  bothers me, and is made slightly more annoying because `itemUnitName` is missing. Wonder if we have to map the item on to the ItemLike object?

mark-prins (approved):
🎉  Looks great! much clearer now

packages/invoices/src/OutboundShipment/DetailView/modals/hooks/useDraftOutboundLines.tsx
mark-prins: yeah - I like it. much clearer

packages/invoices/src/OutboundShipment/DetailView/ContentArea.tsx
mark-prins: For some reason, didn't see the comments while reviewing. Might have opened the PR too early!
Agree with the rename of `OutboundShipmentLineFragment` to `OutboundLineFragment` 😁

packages/system/src/Item/utils.ts
joshxg: Ah nice, thanks. I should have added a TODO to come back before I PRd - I commented this out to make the typings easier to handle at one point and forgot 😁 

It's not too bad to fix now. Previously we omitted the `item` field on `InvoiceLine` because it was so annoying (having all the required fields like `availableBatches` etc) - now `OutboundLine` and `InboundLine` have an `item` field so we can go back to just doing `lines[0].item.unitName : line.item.unitName`

(I'll do this is in the next PR 😁 )

packages/invoices/src/OutboundShipment/api/operations.graphql
joshxg: My thinking was the `Partial` fragment was to make the `DraftOutboundLine` creation easier.

Buuut. I'm thinking that the `OutboundShipmentLine` has too much stuff and we can probably use the one type 😁 

i've almost finished off the inbound side of changes so I'm going to take a look at that after and do the renaming also

PR #934, opened 2022-03-02
#920 list view color mutation

### LINKED ISSUE #920: List views colour update mutations
Each of the list views which have a colour column do not have the mutation correctly hooked up

joshxg:
Fixes #920 

- Really the colour should be optimistically updated - on a very slow connection there's a bit of lag. Thought for the beta this is fine and there's plenty other more important things to do!

### DIFF
--- a/packages/inventory/src/Stocktake/api/hooks.ts
+++ b/packages/inventory/src/Stocktake/api/hooks.ts
@@ -76,19 +76,30 @@ export const useInsertStocktake = () => {
   });
 };
 
+export const useUpdateStocktake = () => {
+  const queryClient = useQueryClient();
+  const api = useStocktakeApi();
+  return useMutation(api.update, {
+    onSuccess: () => {
+      return queryClient.invalidateQueries(['stocktake']);
+    },
+  });
+};
+
 export const useStocktakeFields = <
   KeyOfStocktake extends keyof StocktakeFragment
 >(
   keys: KeyOfStocktake | KeyOfStocktake[]
 ): FieldSelectorControl<StocktakeFragment, KeyOfStocktake> => {
   const { stocktakeNumber = '' } = useParams();
+  const { mutateAsync } = useUpdateStocktake();
   const { data } = useStocktake();
   const api = useStocktakeApi();
   return useFieldsSelector(
     ['stocktake', stocktakeNumber],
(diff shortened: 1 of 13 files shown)


packages/inventory/src/Stocktake/api/hooks.ts
```diff
@@ -76,19 +76,30 @@ export const useInsertStocktake = () => {
...
   const { stocktakeNumber = '' } = useParams();
+  const { mutateAsync } = useUpdateStocktake();
   const { data } = useStocktake();
   const api = useStocktakeApi();
   return useFieldsSelector(
     ['stocktake', stocktakeNumber],
     () => api.get.byNumber(Number(stocktakeNumber)),
     (patch: Partial<StocktakeFragment>) =>
-      api.update({ ...patch, id: data?.id ?? '' }),
+      mutateAsync({ ...patch, id: data?.id ?? '' }),
```
joshxg, line 102: realised this fn (in each of the different features) wasn't invalidating.. so if you were really a big meany you could update a stocktake comment for example, navigate back to the list view and it wouldn't refetch as the list query wasn't invalidated.

packages/invoices/src/InboundShipment/ListView/AppBarButtons.tsx
```diff
@@ -8,13 +9,14 @@ import {
   ButtonWithIcon,
   Grid,
   useTranslation,
+  useToggle,
```
joshxg, line 12: I also moved this modal in a couple of places to the `AppBarButtons` component as that's the lowest it needs to be - it doesn't actually need to be in the `ListView` which is essentially the root - not ideally for triggering state updates (though it didn't actually cause any issues, I suppose)

packages/invoices/src/InboundShipment/ListView/ListView.tsx
```diff
@@ -35,7 +33,7 @@ export const InboundListView: FC = () => {
 
   const columns = useColumns<InboundRowFragment>(
     [
-      [getNameAndColorColumn(), { setter: () => {} }],
+      [getNameAndColorColumn(), { setter: onUpdate }],
```
joshxg, line 36: and this is essentially what I did for Request/Response/Inbound/Outbound

- Move modal to AppBarButtons
- Make `useUpdateX` hook
- Use hook as setter here

PR #938, opened 2022-03-02
#925 custom currency formatter

### LINKED ISSUE #925: Custom currency formatter?
Currently we're using a precision of 2 for currencies, which makes all money values look like 2.00 - we need to use higher to allow for smaller values.

However, if we use a higher precision, i.e. 5, then all money values look like: 2.00000 

We can use a custom formatter for `currency.js`, to remove trailing zeros

joshxg:
Fixes #925 

- Adds a custom formatter which is a rip from https://github.com/scurker/currency.js/blob/66b7a0c6860d5d30efe8edbf4f8ea016149eab55/src/currency.js#L96-L105 just adding some trimming to the cents 
- Think it's OK 🤷


packages/common/src/intl/currency.ts
```diff
@@ -0,0 +1,112 @@
...
+    separator = ',',
+    decimal = '.',
+    // Groups is in the options object, but the type is not right.
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    groups = /(\d)(?=(\d{3})+\b)/g,
+  } = opts ?? {};
+
+  // Split the currency string into cents and dollars.
+  const [dollars = '', cents = ''] = ('' + currency)
```
mark-prins, line 51: thanks currency.js.. the `toString` method returns `0.00` even when the settings say to use `,` as the decimal.
I thought to use `currency.dollar()` and `currency.cents()` to avoid this; but then you have issues with the dropping of trailing zeros that you are testing for.

With the 'fr' language you get this:

Screen Shot 2022-03-03 at 11 50 01 AM

root of the problem is in the `trimCents` which is converting to a number - but the `,` results in `NaN`

I've added a test case for it, have spent a little while fixing but will stop for lunch!

mark-prins (approved):
😃  In general - awesome! Though there is an issue with French currency values. and outbound isn't using the change yet?

packages/invoices/src/InboundShipment/DetailView/columns.ts
```diff
@@ -121,9 +123,11 @@ export const useInboundShipmentColumns = (): Column<
           accessor: ({ rowData }) => {
             if ('lines' in rowData) {
               const { lines } = rowData;
-              return ifTheSameElseDefault(lines, 'sellPricePerPack', '');
+              return c(
+                ifTheSameElseDefault(lines, 'sellPricePerPack', '')
+              ).format();
             } else {
-              return rowData.sellPricePerPack;
+              return c(rowData.sellPricePerPack).format();
```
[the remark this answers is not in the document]
joshxg: Yeah this has been on my mind! `useCurrencyFormat` is a hook used like

```
const currencyString = useCurrencyFormat(line.sellPricePerPack)
```

So it can't be used here

having `c` format straight away would be nice, like the other hooks so I can do that, but I also wanted a way to 'create' currency objects 🤔

packages/common/src/intl/currency.ts
joshxg: Intl is looking better and better 🤔 

Happy to take a look but unsure if you're looking also 😁  let me know!

mark-prins: I was.. but I've stopped. if you can pick it up that'd be great

joshxg: Sorry for any time you wasted on this. Testing french was an obvious thing to do.. 🙄 

I've updated the formatter, though I wouldn't describe it as sophisticated, it seems to do the job. I've updated the tests also, which has meant I needed to stop the mock for react-i18n which has had the side effect of causing 100's of tests to fail, which I'll work on next in a new PR..!

mark-prins: that's what code reviews are for 😁 
There's a couple of clever changes in the currency formatter; you've deftly sidestepped the problem with `toFixed` by splitting on `.` and then the changes in the `trimCents` is very nice!

jest.config.js
```diff
@@ -11,9 +11,8 @@ module.exports = {
   testEnvironment: 'jsdom',
   setupFilesAfterEnv: ['<rootDir>/jest-setup.ts'],
   moduleNameMapper: {
-    'react-i18next': '<rootDir>/__mocks__/react-i18next.ts',
-    ...pathsToModuleNameMapper(compilerOptions.paths, {
-      prefix: '<rootDir>',
+    ...pathsToModuleNameMapper(compilerOptions.path, {
```
mark-prins, line 14:
```suggestion
    ...pathsToModuleNameMapper(compilerOptions.paths, {
```
I had to do this for jest to run, with just `.path` it complains about "undefined" and stuff 🤷

mark-prins (approved):
niiiiiice!

jest.config.js
joshxg: heh thanks

PR #943, opened 2022-03-02
#929 system types common

### LINKED ISSUE #929: System types in common
There are types here: https://github.com/openmsupply/openmsupply-client/blob/main/packages/common/src/types/index.ts

Which shouldn't be used

- `StockLine`
- `Item`
- `Store`
- `User` (?)
- `Test` 
- `Name`

We can add fragments to the individual packages and remove these. They're generally either annoying to use (extending the base gql nodes) or don't interact with code gen at all (so can easily be incorrect and be annoying to use)

joshxg:
Fixes #929 

- Removes the types in common for some system data
- `Name` -> `NameRowFragment` in the name package ✅ 
- `StockLine` wasn't used anywhere, think I removed the last use of it in a previous PR 🤔 
- `Test` wasn't used anywhere 🎉 
- `Store` and `User`- These are a little funny. `Store` -> `StoreRowFragment` in store package ✅  but we also use these types in common in the `localStorage` and `authentication` packages. I think this coupling isn't ideal - for now I've duplicated the types 😕 
- Am doing a separate PR to deal with the `DomainObject` / `RecordWithId` weirdness 👍


packages/requisitions/src/types.ts
```diff
@@ -1,32 +1 @@
...
-    | 'otherPartyName'
-    | 'otherPartyId'
-    | 'theirReference'
-    | 'status'
-  > {
-  color: string;
-  orderDate: Date | null;
-}
-
-export type RequisitionLine = Omit<RequisitionLineNode, '__typename'>;
```
joshxg, line 32: None of these were in use

packages/common/src/authentication/AuthContext.tsx
```diff
@@ -10,6 +9,16 @@ import { AuthenticationResponse } from './api';
...
+type User = {
+  id: string;
+  name: string;
+};
+
+interface Store {
+  id: string;
+  code: string;
+}
+
```
joshxg, line 21: Not ideal. Though, if the types changed, it would likely be for different reasons between here and somewhere else?

mark-prins: yeah, not ideal. Ok for now though - I expect the shape of these will change, and will be useful elsewhere maybe.

mark-prins (approved):
🎉  that all looked good - thanks! another big tidy 😁

PR #944, opened 2022-03-02
Add find and replace for DomainObject -> RecordWithId

### LINKED ISSUE #929: System types in common

joshxg:
Fixes #929 

- We can go with either `DomainObject` or `RecordWithId` whichever you prefer, but I thought it would be better to stick to one instead of having `DomainObject = RecordWithId` and using `DomainObject` everywhere 🤷

### DIFF
--- a/packages/common/src/hooks/useQueryParams/useQueryParams.ts
+++ b/packages/common/src/hooks/useQueryParams/useQueryParams.ts
@@ -1,14 +1,14 @@
-import { DomainObject } from './../../types/index';
+import { RecordWithId } from './../../types/index';
 import { usePagination, PaginationState } from '../usePagination';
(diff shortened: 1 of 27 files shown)


mark-prins (approved):
`RecordWithId` is good for me! I'm just gonna close my eyes and click approve on this one!

🙈  😁

Issue #302, opened 2021-10-20
Detail panel: Add pricing section


josh-griffin:
- Need to add a pricing section to the detail panel with totals etc

also:
```
RM: should include
- possibility to add "Other Charges"
- Subtotal, % tax & Total
```

joshxg:
semi blocked as I'm not sure what to do about tax - maybe it's just already on the invoice? Is it editable? Also heard something about tax being on invoice lines instead

mark-prins:
in desktop, there is `tax` on the `transact` level and not on invoice lines - though it sounds like that is coming for the OMSAPI. I was thinking just to have what is currently on invoices in desktop:

- subtotal
- tax percentage
- total

tax is editable ( if preference is enabled ) on desktop

mark-prins:
Mostly done

PR #945, opened 2022-03-02
#571 Outbound service lines

### LINKED ISSUE #571: Invoice service charges
Possibly related to:

https://github.com/openmsupply/openmsupply-client/issues/302

- Able to create multiple arbitrary 'service' charge key/value pairs to an invoice
- I am thinking this could just be rows in the pricing section 🤷

joshxg:
Fixes #571  / #302 

- Very basic setup for the service line modal, adding the pricing section which is maybe really #302 
- Added a 'fix' for sending total costs per line - any previous lines added won't have updated the totals 

image


packages/invoices/src/OutboundShipment/api/api.ts
```diff
@@ -94,6 +94,11 @@ const outboundParsers = {
       invoiceId: line.invoiceId,
       numberOfPacks: line.numberOfPacks,
       stockLineId: line.stockLine?.id ?? '',
+      tax: { percentage: 10 },
```
joshxg: Added this for now as we're not able to query for tax values - pretending they don't exist. Also the totals need to be sent rather than calculated on the backend, which I think is not ideal. Ideally would just send the `sellPrice` for outbounds (and the tax should generally always be taken from somewhere centrally, with editable tax on the line being the edge case  🤷

packages/invoices/src/OutboundShipment/DetailView/SidePanel.tsx
```diff
@@ -106,6 +111,69 @@ const RelatedDocumentsSection: FC = () => {
   );
 };
 
+export const PricingSection = () => {
+  const t = useTranslation('distribution');
+  const serviceLineModal = useToggle(false);
+  const { c } = useCurrency();
+  const { pricing } = useOutboundFields(['pricing']);
+
+  return (
```
joshxg, line 120: This looks OK. Not great - maybe OK for now?

packages/invoices/src/OutboundShipment/DetailView/SidePanel.tsx
```diff
@@ -106,6 +111,69 @@ const RelatedDocumentsSection: FC = () => {
...
+        <PanelRow>
+          <PanelLabel>{t('heading.sub-total')}</PanelLabel>
+          <PanelField>{c(pricing.totalBeforeTax).format()}</PanelField>
+        </PanelRow>
+        <PanelRow>
+          <PanelLabel>{t('heading.total')}</PanelLabel>
+          <PanelField>{c(pricing.totalAfterTax).format()}</PanelField>
+        </PanelRow>
+      </Grid>
+    </DetailPanelSection>
```
joshxg, line 173: I should probably add `tax` here somewhere also? For each section? 🤷  Would need to calculate from the subtotals and totals 🤔

packages/invoices/src/OutboundShipment/api/api.ts
mark-prins: I thought so too - wasn't following the reasoning on why this couldn't be calculated in the api

packages/invoices/src/OutboundShipment/DetailView/SidePanel.tsx
mark-prins: My helpful comment is that I think the pencil is too big 😄 

I've played around a bit and put a suggestion in 529449ae
I know that using margin is well dodgy.. looks nicer with a bit of spacing is all.

mark-prins (approved):
👍  Looks good!

packages/invoices/src/OutboundShipment/DetailView/SidePanel.tsx
joshxg: Thanks 😁

packages/invoices/src/OutboundShipment/DetailView/SidePanel.tsx
mark-prins: yes, probably should. maybe not right now though?

PR #946, opened 2022-03-03
Inbound item editing validation

### LINKED ISSUE #503: Inbound item editing validation
Need to add some validation: non-negative integers for pack size and # packs

Non negative decimals for prices

mark-prins:
Fixes #503 
Prevent inbound pack size and quantity being negative.
The currency input is already preventing me from too much negativity.

### DIFF
--- a/packages/common/src/ui/layout/tables/components/Cells/NumberInputCell/index.ts
+++ b/packages/common/src/ui/layout/tables/components/Cells/NumberInputCell/index.ts
@@ -1,1 +1,2 @@
 export * from './NumberInputCell';
+export * from './PositiveNumberInputCell';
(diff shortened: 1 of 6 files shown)


packages/common/src/ui/layout/tables/components/Cells/NumberInputCell/index.ts
```diff
@@ -1 +1,2 @@
 export * from './NumberInputCell';
+export * from './PositiveNumberInputCell';
```
mark-prins, line 2: should be in its own folder?

joshxg (approved):
Thinking about a couple others:

- Upper limit for number of packs? Eventually we do overflow which then gets parsed as zero, but it seems like not a good look?
- pack size should be > 1 (Does that mean `NonNegativeInput` and `PositiveInput`? 😁) and likely have some sort of upper limit also?
-  batch: Limit the number of chars? At 1.2mb the field updated fine but it was making input a little troublesome 😁

mark-prins:
thanks - I should have stopped and thought about a pack size of zero 🙄

PR #952, opened 2022-03-03
#950 broken tests bonanza

### LINKED ISSUE #950: Broken tests bonanza
I removed the `react-i18n` mock and it broke basically all of the tests. I am pretty sure that even some tests in other repos broke because of this disaster!!!! 💥 💣 

lets fix it!

joshxg:
Fixes #950 

- Woah! 🎉 
- Basically a bunch of tests which previously were using the locale strings to match things on (i.e. `button.close-the-menu` rather than `close the menu`) or others which were calling translation hooks but did not have a provider after the mock was removed


jest.config.js
```diff
@@ -11,6 +11,10 @@ module.exports = {
   testEnvironment: 'jsdom',
   setupFilesAfterEnv: ['<rootDir>/jest-setup.ts'],
   moduleNameMapper: {
+    // ref: https://react-hooks-testing-library.com/installation#being-specific
+    // This mapping will ensure the correct renderHook is used in the test environment
+    // as the import within `utils/testing` is for the specific dom version.
+    '@testing-library/react-hooks/dom': '@testing-library/react-hooks',
```
joshxg, line 17: Without this, the wrong import is used in tests. In a build, if just using the import `'@testing-library/react-hooks'` - the wrong import is used. 

So.. use  the import `'@testing-library/react-hooks/dom';` and then remap when in a test env

Technically this is because `utils/testing.ts` is being bundled with the app in the dom, when it shouldn't be (should be tree shaken and only used in tests).

packages/common/package.json
```diff
@@ -31,7 +31,9 @@
     "kbar": "^0.1.0-beta.27",
     "lodash": "^4.17.21",
     "notistack": "^2.0.3",
+    "react": "^17.0.2",
     "react-currency-input-field": "^3.6.0",
+    "react-dom": "^17.0.2",
```
joshxg, line 36: Seemed look a good idea..?

mark-prins: 😬  sure..

packages/common/src/ui/components/menus/ColorMenu/ColorMenu.test.tsx
```diff
@@ -22,7 +22,7 @@ describe('ColorMenu', () => {
     };
 
     return (
-      <TestingProvider>
+      <>
```
mark-prins, line 25: Is there a reason for the move?

mark-prins (approved):
🎉  woo! un-broken!

packages/common/src/ui/components/menus/ColorMenu/ColorMenu.test.tsx
joshxg: There's a few like this, where `useTranslation` was used outside of the provider.. i.e.

```
const SomeComponent = () => {
    const t = useTranslation();
    return <TestingProvider> ... </...>
}
```

Can't see the `useTranslation` in the diff on this one

mark-prins: ah. sneaky

Issue #931, opened 2022-03-02
query key factories for maintainability


joshxg:
Currently we're *generally* inlining query keys.

I've already come up against a couple of bugs where I've miss copy pasted a query key and/or mis-typed.

It would be good to have a strong pattern for query keys to:

- Avoid mistyping / using the wrong key
- Easily able to determine at what level in the hierarchy we can invalidate, at
- Easily use the query keys

One way is to use a global query key file/folder, with all the query keys, so that we can easily view all query keys. This would also stop accidentally using a query key the same as another feature.

However, I think colocation next to where the queries are used is generally better all round.

I am thinking to have something like this in the `useXXApi` hook:

```
const keys = {
  base: () => [ 'requisition', storeId ],
  detail: (id) => [ ...keys.base(), id ],
  list: () => [ ...keys.base(), 'list' ],
  listWithParam: (params) => [...keys.list(),  params ],
}
```

Then when we use the api..

`useQuery(api.keys.detail(id), () => api.get.byId(id))`

mark-prins:
that is a neat idea. Agree with colocation - the operations and api are colocated too, so that makes sense.

Issue #936, opened 2022-03-02
Put the query parameters for items/lists in a more accessible place


josh-griffin:
The hooks `useInboundItems` / `useInboundLines` / `useInbounds` etc have a `useQueryParams` per instance.

So, if you're to call `useInbounds` in two places, the underlying data (the queried inbound shipments) can be the same (if the query params are the same, as that is in the query key), however if you have the query params for one instance of the hook, this does not effect the other.

i.e.

```
const { data: d1, onSortBy: sortD1 } = useInbounds()
const { data: d2, onSortBy: sortD2 } = useInbounds()

console.log(d1 === d2) // This is true

sortD2('key') // After this call, the above will log false, as d2 is a new set of data and d1 is not updated.
```

We can fix this by having the query parameters and controllers live in for example, zustand state, the URL or context.

This isn't so much of a problem right now and could potentially be left as these hooks aren't called all that much.

My preference would be to have a single source of truth for the query parameters in the URL as that also solves the URL issue

mark-prins:
Fixed with #1204

PR #956, opened 2022-03-03
#931 Query key factory for inbound

joshxg:
Starts #931 

- Before I go and add this to all the api files, I wanted to do one and get feedback / see if you had any suggestions / improvements / whatever

### DIFF
--- a/packages/invoices/src/InboundShipment/api/hooks.ts
+++ b/packages/invoices/src/InboundShipment/api/hooks.ts
@@ -20,7 +20,7 @@ import {
 import { ItemRowFragment } from '@openmsupply-client/system';
 import { inboundLinesToSummaryItems } from './../../utils';
 import { InboundItem } from './../../types';
-import { getInboundQueries } from './api';
+import { getInboundQueries, ListParams } from './api';
 import {
   getSdk,
   InboundFragment,
@@ -30,9 +30,16 @@ import {
 
 export const useInboundApi = () => {
   const { storeId } = useAuthContext();
+  const keys = {
+    base: () => ['inbound'] as const,
+    detail: (id: string) => [...keys.base(), storeId, id] as const,
+    list: (params: ListParams) =>
+      [...keys.base(), storeId, 'list', params] as const,
+  };
+
   const { client } = useOmSupplyApi();
   const queries = getInboundQueries(getSdk(client), storeId);
-  return { ...queries, storeId };
+  return { ...queries, storeId, keys };
 };
 
 const useInvoiceNumber = () => {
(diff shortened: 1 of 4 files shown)


packages/invoices/src/InboundShipment/api/hooks.ts
```diff
@@ -30,9 +30,16 @@ import {
 
 export const useInboundApi = () => {
   const { storeId } = useAuthContext();
+  const keys = {
+    base: () => ['inbound'] as const,
+    detail: (id: string) => [...keys.base(), storeId, id] as const,
+    list: (params: ListParams) =>
+      [...keys.base(), storeId, 'list', params] as const,
+  };
```
joshxg, line 38: Maybe would be better to have a `stores` key which was `[ ...base(), storeId ]` to build upon.. then you could also invalidate at that level - 'all the inbounds for this store'

Also wonder if there needs to be a list v list with params.. 'invalidate all the inbound lists' .. v 'invalidate all the inbounds for this store' .. 🤷

packages/invoices/src/InboundShipment/api/hooks.ts
```diff
@@ -191,7 +196,7 @@ export const useInbounds = () => {
   const api = useInboundApi();
 
   return {
-    ...useQuery(['invoice', 'list', api.storeId, queryParams], () =>
+    ...useQuery(api.keys.list(queryParams), () =>
```
joshxg, line 199: use `queryParams` actually seems not ideal.. it's a massive object with a bunch of duplicate things which gets diff'd all the time..

I think it would be a better idea to try and align this better with the generated gql code - i.e. `PaginationInput` etc..

I had a quick look but it'd take a bit more work to untangle all the filtering/sorting/pagination so have left that - likely something to tackle with #936 , I think

mark-prins: Something that would be good to look at with #936 I agree. 
The `queryParams` is ok for now but could be simplified a bit for use in the key.

Was wondering about alignment with the gql types and if that helps anything

packages/invoices/src/InboundShipment/api/hooks.ts
mark-prins: I'm not sure of the need for a key of 'all things in this store' though that is easier to call than invalidating a specific id and all lists. 

I think a list key would be helpful, and then extend that for the list with params version.

mark-prins (approved):
👍  That looks great - super helpful I think!

packages/invoices/src/InboundShipment/api/hooks.ts
joshxg: I wouldn't mind a query params 'system' that would allow for better typing of sort keys etc which would likely require using the gql types

PR #963, opened 2022-03-04
PaperClickPopover

### LINKED ISSUE #597: ClickPopover component
Want a popover which opens on a click rather than a hover!

Unfortunately the paperpopover uses the hoverpopover in https://github.com/openmsupply/openmsupply-client/tree/main/packages/common/src/ui/components/popover so won't do the job. 

Make `PaperPopover` .. `PaperHoverPopover` and make a `PaperClickPopover` 🤷 ? Or.. give `PaperPopover` a prop `opensOnHover`. My gut feeling is that prop alone basically changes most of the implementation of the component so just make the second one. But my other gut says that having two components is not a great api for a consumer. Life is so hard isn't it?

mark-prins:
Fixes #597 

Slightly overboard perhaps.. using inheritance and extending the current poppers etc. 
Renamed existing paper popover and added a click only version, and after than, wondering why I didn't just rename the hover and add an option 🙄


packages/common/src/ui/components/popover/ClickPopover/ClickPopover.tsx
```diff
@@ -0,0 +1,40 @@
...
+  hide,
+  Popover,
+}) => (
+  <>
+    <div style={{ cursor: 'pointer' }} onClick={show}>
+      {children}
+    </div>
+    <Popover placement={placement}>
+      <Box>
+        <IconButton
```
mark-prins, line 28: ok, so maybe worth it. added a close button to the click popover, which isn't required on the hover version.

packages/common/src/ui/components/popover/PaperPopover/PaperClickPopover/PaperClickPopover.tsx
```diff
@@ -39,20 +41,6 @@ export const PaperPopover: FC<PaperPopoverProps> = ({
       }
     >
       {children}
-    </HoverPopover>
+    </ClickPopover>
   );
 };
-
-export interface PaperPopoverSectionProps {
```
mark-prins, line 46: moved these out so I could reuse them

packages/common/src/ui/components/popover/PaperPopover/PaperClickPopover/usePaperClickPopover.tsx
```diff
@@ -0,0 +1,17 @@
+import React, { FC } from 'react';
+import { usePopover } from '../../BasePopover';
+import {
+  PaperClickPopover as ClickPopover,
+  PaperClickPopoverProps,
+} from './PaperClickPopover';
+
+export const usePaperClickPopover = () => {
```
mark-prins, line 8: made the hook to make usage easier. My problem was wanting to use `hide` in both the content and in the popover - so I lifted that out to the consumer, which meant passing it back in, which is annoying.

joshxg: Ah yes I see the problem.. annoying! Good work around though, makes it nice and easy

packages/common/src/ui/components/popover/PaperPopover/PaperClickPopover/PaperClickPopover.tsx
joshxg: very nice

packages/common/src/ui/components/popover/ClickPopover/ClickPopover.tsx
joshxg: Ah yes.

I like the close button but I also wonder if clicking away should close also.. that was my expectation.. or, at least when navigating away, I suppose.

I had a play and adding `<ClickAwayListener onClickAway={hide}>` around the `Box` seems to work fine if you think so also

packages/common/src/ui/components/popover/PaperPopover/PaperClickPopover/usePaperClickPopover.tsx
```diff
@@ -0,0 +1,17 @@
...
+  const { show, hide, Popover } = usePopover();
+  const PaperClickPopover: FC<
+    Omit<PaperClickPopoverProps, 'show' | 'hide' | 'Popover'>
+  > = props => (
+    <ClickPopover show={show} hide={hide} Popover={Popover} {...props} />
+  );
+
+  return { show, hide, PaperClickPopover };
```
joshxg, line 16: very nice!

joshxg (approved):
Love it!

joshxg (approved):
Ah .. I noticed there was a slight delay in opening / closing which was irking me .. I thought it might have been a render issue, but it's actually that the `Popover` is using `useDebounceCallback` (was for the hovering 🙄 ) with a delay of 250ms on opening and 500ms on closing..

Adding some optional parameters for the timeout to `usePopover` should be a straight forward change to get rid of the delay

joshxg:
My suggestions are minor and could be ignored.. since I'm shellfish and want to utilise the store picker.. lets get it in 😁

packages/common/src/ui/components/popover/ClickPopover/ClickPopover.tsx
mark-prins: yeah - was undecided, my initial reaction was that click away should work. thanks for adding it in

PR #968, opened 2022-03-04
#962 yarn generate api fixes - The PR that all of NZ is talking about -- [1]

### LINKED ISSUE #962: Add updated schema from a lot of backend changes
There are a lot of backend changes if you were to do `yarn generate`!

Want to do a PR with just the result of running `yarn generate`

joshxg:
Fixes #962 

- Basically I think there was some error shuffling around - renaming, removing etc.
- Getting errors in almost all mutations like: "Cant spread this type it doesn't exist" etc
- So.. I went through all the operation files with mutations (queries all seemed fine) and updated the error returns.
- side note: There's an argument, I think, to just use a single batch gql mutation for invoices/requisitions/stocktakes and then have the `api` file manage that.. at the moment it's bit of a pick n mix  🤷 

Update: Seems I didn't fix requisition updates. I did that here: https://github.com/openmsupply/openmsupply-client/pull/971#pullrequestreview-900913468


joshxg:
Just noticed the outbound delete lines mutation is still broken.. will fix soon!

mark-prins (approved):
🎉 
might have changed again? I had to remove deleteOutboundShipmentLines.

I also have a problem when adding lines to an outbound - there is always a delete of unallocated and it sends an insert for quantity of 0 which also is 'bad user input'. will raise an issue, though likely covered in one of the other PRs

joshxg:
Thanks - my mistakes though.

Delete outbound I missed. Adding outbound rows is fixed in a later PR https://github.com/openmsupply/openmsupply-client/pull/980

Issue #961, opened 2022-03-04
Requisition changes from API Changes


joshxg:
- https://github.com/openmsupply/remote-server/issues/803
- https://github.com/openmsupply/remote-server/issues/870

PR #971, opened 2022-03-05
#961 Requisition changes from api changes - But would THEY really admit to it? -- [5]

joshxg:
#961 

- Replaces the many promises implementation for bulk deleting requisitions to using the batch mutation
- Adds requisition line mutations for comments


packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEditForm.tsx
```diff
@@ -119,6 +119,8 @@ export const RequestLineEditForm = ({
             {t('heading.comment')}
           </Typography>
           <TextArea
+            value={draftLine?.comment}
+            onChange={e => update({ comment: e.target.value })}
```
joshxg, line 123: The comment component was already in place.. just needed to add it to the query and pass through the values/trigger an update

mark-prins (approved):
👍  LGTM!

PR #972, opened 2022-03-05
#948 Master list search modal - Are the rumours about JACINDA true?  -- [6]

### LINKED ISSUE #948: Master list search modal
For #581 we need a way to select a master list.

A master list search modal similar to name, item etc would do the trick!

Though, I wonder if we want a confirmation modal 🤔

joshxg:
Fixes #948 

- Adds a pretty basic master list search modal component
- I wonder if it would be a good idea to add some button components in the system packages - `MasterListSearchModalButton`
- I also wondered about using some sort of context to have system modals like this.. so instead you can just use a hook `useMasterListSearch({ onSelect })` 🤔


packages/host/src/CommandK.tsx
```diff
@@ -81,22 +81,6 @@ export const CommandK: FC = ({ children }) => {
   const drawer = useDrawer();
   const t = useTranslation('app');
   const actions = [
-    // {
```
joshxg, line 84: 🤷

packages/host/src/CommandK.tsx
```diff
@@ -254,7 +226,7 @@ export const CommandK: FC = ({ children }) => {
       keywords: 'master lists',
       perform: () =>
         navigate(
-          RouteBuilder.create(AppRoute.Inventory)
```
joshxg, line 257: offroad

packages/requisitions/src/RequestRequisition/DetailView/AppBarButtons.tsx
```diff
@@ -18,11 +20,22 @@ export const AppBarButtonsComponent: FC<AppBarButtonProps> = ({
...
+        <ButtonWithIcon
+          Icon={<PlusCircleIcon />}
+          label={t('button.add-from-master-list')}
+          onClick={modalController.toggleOn}
+        />
+        <MasterListSearchModal
+          open={modalController.isOn}
+          onClose={modalController.toggleOff}
+          onChange={masterList => console.log(masterList)}
+        />
```
joshxg, line 38: as mentioned in the top level comment.. it's not a lot of code to add where you want it.. but it could be less 😁 

Though.. I suppose these should maybe lazily loaded 🤔

packages/system/src/MasterList/Components/MasterListSearchModal/MasterListSearchModal.tsx
```diff
@@ -0,0 +1,33 @@
...
+      options={data?.nodes ?? []}
+      onClose={onClose}
+      title={t('master-lists')}
+      optionKey="name"
+      onChange={(_, masterList: MasterListRowFragment | null) => {
+        if (masterList) onChange(masterList);
+      }}
+    />
+  );
+};
```
joshxg, line 33: No pagination etc.. but don't think it's needed

packages/system/src/MasterList/api/api.ts
```diff
@@ -3,36 +3,27 @@ import {
   FilterBy,
   MasterListSortFieldInput,
 } from '@openmsupply-client/common';
-import { Sdk } from './operations.generated';
-import { MasterList, MasterListRow } from './../types';
+import { Sdk, MasterListRowFragment } from './operations.generated';
```
joshxg, line 6: general updates to master list api for consistency

packages/requisitions/src/RequestRequisition/DetailView/AppBarButtons.tsx
```diff
@@ -18,11 +20,22 @@ export const AppBarButtonsComponent: FC<AppBarButtonProps> = ({
...
+  const t = useTranslation('distribution');
+  const modalController = useToggle();
 
   return (
     <AppBarButtonsPortal>
       <Grid container gap={1}>
+        <ButtonWithIcon
+          Icon={<PlusCircleIcon />}
+          label={t('button.add-from-master-list')}
+          onClick={modalController.toggleOn}
```
mark-prins, line 32: and `disabled={isDisabled}` maybe?

packages/requisitions/src/RequestRequisition/DetailView/AppBarButtons.tsx
mark-prins: yes, was wondering about generalising this - could just be one component which manages the state internally.
like you say though, this isn't much code - would be good to get this in place and then think about how it could be simplified for the consumer.

mark-prins (approved):
👍 very nice!

packages/requisitions/src/RequestRequisition/DetailView/AppBarButtons.tsx
joshxg: nice

PR #973, opened 2022-03-05
#581 Add from master list - A kind stranger made this PR and got an UNFORGETTABLE SURPRISE -- [7]

### LINKED ISSUE #581: Populate a supplier requisition by selecting a master list
- Add from list button
image

1. Adding `masterlist` sub module to `system` with an exposed list modal
2. will probably need a blocking modal to while we wait for a request to the server for the items of the master list and adding them to the requisition

joshxg:
Fixes #581 

- Minor change to `useConfirmationModal` which I hope is a good change, but made diffs in a fair amount of files

Essentially the issue being:

```
const { mutate } = useMutation(api.addFromMasterList)

const addFromMasterList = async (masterList) => {
    await  mutate(masterList)
}

const getConfirmation = useConfirmation({text: "are you sure", message: "yes I'm sure!", onConfirm: () => { ... } })

return <MasterListSearchModalList onChange={masterList => { 
    toggleOff(); 
    // ---- Need to open the confirm here and also pass the master list into the onConfirm callback ----
    // ---- would need to set some state with the master list and then trigger the getConfirmation   ----
   // ----  with a useEffect.. Then, remember to clear the state! 
   // ---- so, instead:
   getConfirmation({ onChange: addFromMasterList })
}
/>
```

Though, I've made it so you can pass the `onConfirm` in either with the initial call or the extra call.. which means it's optional in both, ending with a situation where you can have no onConfirm 🤔

### DIFF
--- a/packages/common/src/intl/locales/en/distribution.json
+++ b/packages/common/src/intl/locales/en/distribution.json
@@ -23,6 +23,8 @@
   "heading.sub-total": "Sub total",
   "heading.total": "Total",
   "heading.stock-details": "Stock details",
+  "heading.are-you-sure": "Are you sure?",
+  "message.confirm-add-from-master-list": "Are you sure you want to add all of the items from this master list?",
   "heading.create-outbound-shipment": "Create Outbound Shipment",
   "label.add-batch": "Add batch",
   "label.allocated": "Allocated",
--- a/packages/common/src/intl/locales/en/distribution.json
+++ b/packages/common/src/intl/locales/en/distribution.json
@@ -24,7 +24,7 @@
   "heading.total": "Total",
   "heading.stock-details": "Stock details",
   "heading.are-you-sure": "Are you sure?",
-  "message.confirm-add-from-master-list": "Are you sure you want to add all of the items from this master list?",
+  "message.confirm-add-from-master-list": "Do you want to add all of the items from this master list?",
   "heading.create-outbound-shipment": "Create Outbound Shipment",
   "label.add-batch": "Add batch",
   "label.allocated": "Allocated",
--- a/packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModal.stories.tsx
+++ b/packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModal.stories.tsx
@@ -59,7 +59,7 @@ const UseConfirmationModalHook: Story = () => {
     onConfirm: () => alert('confirmed!'),
   });
 
-  return <BaseButton onClick={getConfirmation}>Open Modal</BaseButton>;
+  return <BaseButton onClick={() => getConfirmation()}>Open Modal</BaseButton>;
 };
 
 export const Primary = Template.bind({});
--- a/packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModalProvider.tsx
+++ b/packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModalProvider.tsx
@@ -82,7 +82,6 @@ export const ConfirmationModalProvider: FC = ({ children }) => {
         onConfirm={onConfirm}
         onCancel={() => {
           setState(state => ({ ...state, open: false }));
-          console.log('cancel', onCancel);
           onCancel && onCancel();
         }}
         iconType={iconType}
--- a/packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModalProvider.tsx
+++ b/packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModalProvider.tsx
@@ -13,14 +13,20 @@ interface ConfirmationModalState {
   message: string;
   title: string;
   iconType?: 'alert' | 'info';
-  onConfirm: (() => void) | (() => Promise<void>);
+  onConfirm?: (() => void) | (() => Promise<void>);
+  onCancel?: (() => void) | (() => Promise<void>);
 }
 
 interface ConfirmationModalControllerState extends ConfirmationModalState {
   setState: (state: ConfirmationModalState) => void;
   setMessage: (message: string) => void;
   setTitle: (title: string) => void;
-  setOnConfirm: (onConfirm: (() => Promise<void>) | (() => void)) => void;
+  setOnConfirm: (
+    onConfirm: (() => Promise<void>) | (() => void) | undefined
+  ) => void;
+  setOnCancel: (
+    onCancel: (() => Promise<void>) | (() => void) | undefined
+  ) => void;
   setOpen: (open: boolean) => void;
 }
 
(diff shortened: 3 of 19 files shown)


packages/common/src/intl/locales/en/distribution.json
```diff
@@ -23,6 +23,8 @@
   "heading.sub-total": "Sub total",
   "heading.total": "Total",
   "heading.stock-details": "Stock details",
+  "heading.are-you-sure": "Are you sure?",
+  "message.confirm-add-from-master-list": "Are you sure you want to add all of the items from this master list?",
```
joshxg, line 27: 🤷

packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModal.stories.tsx
```diff
@@ -59,7 +59,7 @@ const UseConfirmationModalHook: Story = () => {
     onConfirm: () => alert('confirmed!'),
   });
 
-  return <BaseButton onClick={getConfirmation}>Open Modal</BaseButton>;
+  return <BaseButton onClick={() => getConfirmation()}>Open Modal</BaseButton>;
```
joshxg, line 62: Bunch of these changes because I added a param to `getConfirmation`

packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModalProvider.tsx
```diff
@@ -82,13 +97,18 @@ export const useConfirmationModal = ({
...
+  const { setOpen, setMessage, setOnConfirm, setOnCancel, setTitle } =
+    useContext(Context);
 
-  const trigger = () => {
-    setMessage(message);
-    setOnConfirm(onConfirm);
-    setTitle(title);
+  const trigger = (
+    paramPatch?: Partial<PartialBy<ConfirmationModalState, 'open'>>
+  ) => {
```
joshxg, line 107: It might be better to force the onChange to be required here 🤔

packages/requisitions/src/RequestRequisition/api/hooks.ts
```diff
@@ -202,3 +204,33 @@ export const useDeleteSelectedRequisitions = () => {
...
+  const addFromMasterList = async ({
+    id: masterListId,
+  }: MasterListRowFragment) => {
+    getConfirmation({
+      onConfirm: () => mutationState.mutateAsync({ masterListId, requestId }),
+    });
+  };
+
+  return { ...mutationState, addFromMasterList };
+};
```
joshxg, line 259: This is *usually* quite an annoying flow to do 

- open a search modal 
- close it
- show a confirmation
- trigger ab=n action with the original selected value

pretty happy how it's turned out.. a part from the optional `onChange`

packages/requisitions/src/RequestRequisition/api/hooks.ts
```diff
@@ -202,3 +204,56 @@ export const useDeleteSelectedRequisitions = () => {
...
+                  return error('This master list has been deleted!')();
+                }
+                case 'MasterListNotFoundForThisStore': {
+                  return error(
+                    "Uh oh this is not the master list you're looking for"
+                  )();
+                }
+                default:
+                  return error('Could not add items to requisition')();
+              }
```
joshxg, line 251: I didn't spend time adding translations for this yet. I am not sure if it'll stay?

I *think* it's a decent idea.. essentially throw the error 'code' from the api fn and then switch on it.. likely a better implementation and maybe worth some iterations.

packages/requisitions/src/RequestRequisition/api/hooks.ts
```diff
@@ -202,3 +204,56 @@ export const useDeleteSelectedRequisitions = () => {
...
+          { masterListId, requestId },
+          {
+            onError: e => {
+              const { message } = e as Error;
+              switch (message) {
+                case 'CannotEditRequisition': {
+                  return error('Cannot edit requisition')();
+                }
+                case 'RecordNotFound': {
+                  return error('This master list has been deleted!')();
```
joshxg, line 242: could navigate away here 🤷

packages/requisitions/src/RequestRequisition/api/hooks.ts
```diff
@@ -202,3 +204,56 @@ export const useDeleteSelectedRequisitions = () => {
...
+
+  const addFromMasterList = async ({
+    id: masterListId,
+  }: MasterListRowFragment) => {
+    getConfirmation({
+      onConfirm: () =>
+        mutationState.mutate(
+          { masterListId, requestId },
+          {
+            onError: e => {
```
joshxg, line 235: The reason to use the `onError` here on the `mutate` fn, rather than on the hook `useMutation`, is because this won't be triggered if the hook is still mounted. If the user navigates away, the toast won't show (is this desired? I think so, but unsure)

packages/requisitions/src/RequestRequisition/api/hooks.ts
```diff
@@ -202,3 +204,56 @@ export const useDeleteSelectedRequisitions = () => {
...
+  const queryClient = useQueryClient();
+  const { id: requestId, requisitionNumber } = useRequestFields([
+    'id',
+    'requisitionNumber',
+  ]);
+  const api = useRequestApi();
+  const mutationState = useMutation(api.addFromMasterList, {
+    onSettled: () =>
+      queryClient.invalidateQueries(api.keys.detail(String(requisitionNumber))),
+  });
```
joshxg, line 219: I'm thinking that all invalidates can pretty much be on the `onSettled` rather than `onSuccess`. Reasoning is because the rejection may be because something has happened to the record on the server since queried.. could invalidate on a specific error which might be caused by this use case, but I think it's simpler to just *always* use `onSettled` to avoid forgetting

packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModalProvider.tsx
mark-prins: but that makes it hard to use when not passing in a trigger-time version of `onConfirm`. tricky either way..
I prefer what you've done

packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModalProvider.tsx
```diff
@@ -13,14 +13,20 @@ interface ConfirmationModalState {
   message: string;
   title: string;
   iconType?: 'alert' | 'info';
-  onConfirm: (() => void) | (() => Promise<void>);
+  onConfirm?: (() => void) | (() => Promise<void>);
```
mark-prins, line 16: it troubles me, having the `onConfirm` as optional, but I don't see a good alternative.
could require passing in `onConfirm` when triggering I suppose - wondering if that would be clearer overall

mark-prins (approved):
👍 that's great! - and the confirmation good to have

packages/common/src/intl/locales/en/distribution.json
joshxg: Yes! much better

PR #974, opened 2022-03-05
#954 Delete request requisition lines - A week after this PR was made, the TRUTH was revealed -- [8]

### LINKED ISSUE #954: Implement request line deletion
Request requisition line deletion is not implemented.

It should be!

joshxg:
Fixes #954 

- Straight forward deletion of rows
- Removed some commented out code which was done because of waiting for the API to implement a requisition date field. However apparently it's not needed

### DIFF
--- a/packages/requisitions/src/RequestRequisition/api/api.ts
+++ b/packages/requisitions/src/RequestRequisition/api/api.ts
@@ -13,6 +13,7 @@ import { DraftRequestLine } from './../DetailView/RequestLineEdit/hooks';
 import {
   RequestRowFragment,
   RequestFragment,
+  RequestLineFragment,
   Sdk,
 } from './operations.generated';
 
@@ -74,6 +75,7 @@ const requestParser = {
       status: requestParser.toStatus(requisition),
     };
   },
+  toDeleteLine: (line: RequestLineFragment) => ({ id: line.id }),
   toInsertLine: (
     line: DraftRequestLine
   ): InsertRequestRequisitionLineInput => ({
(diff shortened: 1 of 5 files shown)


packages/requisitions/src/RequestRequisition/api/api.ts
```diff
@@ -123,6 +125,16 @@ export const getRequestQueries = (sdk: Sdk, storeId: string) => ({
       throw new Error('Record not found');
     },
   },
+  deleteLines: async (requestLines: RequestLineFragment[]) => {
+    const ids = requestLines.map(requestParser.toDeleteLine);
+    const result = await sdk.deleteRequestLines({ ids, storeId });
+
+    if (result.batchRequestRequisition.deleteRequestRequisitionLines) {
+      return result.batchRequestRequisition.deleteRequestRequisitionLines;
+    }
```
mark-prins, line 134: too much? was wondering about adding in error handling into the hook and realised that the batch return needed some checking before it would be helpful

```suggestion
    if (result.batchRequestRequisition.deleteRequestRequisitionLines) {
      const failedLines =
        result.batchRequestRequisition.deleteRequestRequisitionLines.filter(
          line =>
            line.response.__typename === 'DeleteRequestRequisitionLineError'
        );
      if (failedLines.length === 0) {
        return result.batchRequestRequisition.deleteRequestRequisitionLines;
      }
    }
```

mark-prins (approved):
👍  lookin' good!

packages/requisitions/src/RequestRequisition/api/api.ts
joshxg: Nice.

Yeah.. so if we went with throwing errors and utilising `onError` to catch and action, then we would likely want to throw whatever the errors were in the response array..

So maybe we could create a custom error object to wrap either a single response error or array of errors and catch that in `onError` 🤔

PR #975, opened 2022-03-05
#942 Pack size controller when only one placeholder - We were SHOCKED at what this bug was doing  -- [9]

### LINKED ISSUE #942: Pack size controller when only placeholder lines
As per title - the packSize controller isn't coping well when there is only a placeholder row:

Screen Shot 2022-03-03 at 11 05 35 AM

joshxg:
Fixes #942 

- Just hides the control.. the actual logic seems fine.
- Tried another fix to show 'any' when there was a placeholder, but I thought this was better?


mark-prins (approved):
🎉  LGTM!
I suppose, technically, it should show pack size of `1` as the only option, since that's what the placeholder row gives you. this is neater and requires much less awkward logic

mark-prins:
Hm. Just noticed something else.. which I'm calling #985

PR #977, opened 2022-03-05
#949 Prevent status changes until all lines allocated - This PR did the most incredible thing -- [10]

### LINKED ISSUE #949: Prevent outbound shipment status changes when there are placeholders
When there are unallocated lines with quantity in a outbound shipment, the user should be prevented from changing the status to allocated.

Also, placeholder rows are disabled when the status is allocated+, but maybe it should be hidden instead? 🤷

joshxg:
Fixes #949 

- Off road created an alert modal context - it makes it so easy 🤷 
- Also offroad factored confirmation modal file
- the actual issue was pretty trivial and could have been a couple of lines.. soz about it!!


packages/common/src/intl/locales/en/distribution.json
```diff
@@ -13,6 +13,7 @@
   "warning.nothing-to-supply": "Nothing left to supply!",
   "error.failed-to-create-outbound": "Failed to create Outbound Shipment",
   "heading.comment": "Comment",
+  "heading.cannot-do-that": "Can't do that!",
```
joshxg, line 16: 🤷

packages/common/src/types/utility.ts
```diff
@@ -3,3 +3,5 @@ import { RecordWithId } from './index';
 export type ObjectWithStringKeys = Record<string, unknown>;
 
 export type RecordPatch<T extends RecordWithId> = Partial<T> & { id: string };
+
+export type PartialBy<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>;
```
joshxg, line 7: Better name? Making specific properties optional 🤔

packages/common/src/ui/components/modals/AlertModal/AlertModalContext.tsx
```diff
@@ -0,0 +1,28 @@
...
+export const AlertModalContext = createContext<AlertModalControllerState>({
+  open: false,
+  message: '',
+  title: '',
+  setOpen: () => {},
+  setState: () => {},
+  setMessage: () => {},
+  setTitle: () => {},
+  setOnOk: () => {},
+});
```
joshxg, line 28: 😁  

I find the `useConfirmationModal` hook incredibly useful, versatile and easy - thought to do the same thing with the alert modal

packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModalContext.ts
```diff
@@ -0,0 +1,39 @@
+import { createContext } from 'react';
```
joshxg, line 1: When I did the alert modal I found that separating the different pieces made it clearer. Hope it does for you also!

packages/common/src/ui/index.ts
```diff
@@ -6,5 +6,4 @@ export * from './components';
 export * from './forms';
 export * from './icons';
 export * from './layout';
-export * from './notifications';
```
joshxg, line 9: The naming of Alert*Modal* made me move it to the modal folder... though the terminology between modals, dialogs and alerts isn't quite clear and distinct!  🤔

packages/host/src/Host.tsx
```diff
@@ -17,6 +17,7 @@ import {
   RandomLoader,
   ConfirmationModalProvider,
   AuthProvider,
+  AlertModalProvider,
```
joshxg, line 20: I think soon we need to start grouping some of this? Though.. then we would need to start naming the groups. Uh oh..

packages/invoices/src/OutboundShipment/DetailView/Footer/StatusChangeButton.tsx
```diff
@@ -133,24 +139,49 @@ const useStatusChangeButton = () => {
     setSelectedOption(() => getNextStatusOption(status, options));
   }, [status, options]);
 
-  return { options, selectedOption, setSelectedOption, onGetConfirmation };
+  return { options, selectedOption, setSelectedOption, getConfirmation };
+};
+
+const useStatusChangePlaceholderCheck = () => {
```
joshxg, line 145: Aaaand the actual change for the issue. Ended up being so simple and I think is informative for the user rather than just disabling the status? 🤔

packages/common/src/types/utility.ts
mark-prins: first thought was `PartialOn` which says the same thing. `PartialBy` is fine

packages/common/src/ui/components/modals/ConfirmationModal/ConfirmationModalContext.ts
mark-prins: yeah, it did. I wondered about splitting this one as well when I read through the alert modal

packages/common/src/ui/index.ts
mark-prins: I'm thinking about renaming all the dialog bits to 'modal' and be done with it. I think of dialog as a popup thingy and a modal as a blocking popup thingy - you can't continue until you interact with it.

Since we are only using blocking alert windows, they are really all _modals_ and we might as well just call them that!

packages/host/src/Host.tsx
mark-prins: 😱

packages/invoices/src/OutboundShipment/DetailView/Footer/StatusChangeButton.tsx
mark-prins: much better than disabling - which could be confused with  being disabled because on hold, or finalised.
🤔  on hold doesn't disable anything.. should it?

mark-prins (approved):
👍  awesome!

joshxg:
> on hold doesn't disable anything.. should it?

Whoops! Yes, I've added disabling of the status button when on hold. Actually, I wonder if setting on hold should also get a confirmation..

mark-prins:
> Actually, I wonder if setting on hold should also get a confirmation..

to be consistent with stocktakes, yes.  to be consistent with mSupply, no 😄 
I'm thinking yes, will pop an issue in.

PR #980, opened 2022-03-06
#979 Update Outbound line mutations - This changes EVERYTHING -- [11]

### LINKED ISSUE #979: Sending the right line edit mutations
For each of the line edit saving mutations the logic needs to be adjusted:

Outbound:
- Only send mutations for packs with zero or more packs. Delete if updating to zero.

Inbound: 
- Only send mutations for packs with zero or more packs. Delete if updating to zero.

Stocktake:
- Should only insert or update if `countedNumberOfPacks` is not `null`
- Should delete if `countedNumberOfPacks` is `null`

Response
- I think nothing. Supplying zero quantity is legitimate

Request
- I think nothing. We can send updates for zero quantity with this. Also when we add from master list we have lines with requested quantity of zero

joshxg:
Fixes #979 Part1 😁 

- This updates the outbound line mutation so that when setting a lines number of packs to zero, the line is deleted. Otherwise, the update is sent and the API throws and error
- I was also working on inbound and refactored a little bit too much, so thought to do another PR with a different diff

### DIFF
--- a/packages/invoices/src/OutboundShipment/api/api.ts
+++ b/packages/invoices/src/OutboundShipment/api/api.ts
@@ -108,6 +108,13 @@ const outboundParsers = {
         line.numberOfPacks * (line.stockLine?.sellPricePerPack ?? 0),
     };
   },
+  toDeleteLine: (line: {
+    id: string;
+    invoiceId: string;
+  }): DeleteOutboundShipmentLineInput => ({
+    id: line.id,
+    invoiceId: line.invoiceId,
+  }),
   toInsertPlaceholder: (
     line: DraftOutboundLine
   ): InsertOutboundShipmentUnallocatedLineInput => ({
@@ -124,13 +131,6 @@ const outboundParsers = {
   }),
   toDeletePlaceholder: (line: DraftOutboundLine) => ({
     id: line.id,
-  }),
-  toDeleteLine: (
-    invoiceId: string,
-    id: string
-  ): DeleteOutboundShipmentLineInput => ({
-    invoiceId,
-    id,
   }),
 };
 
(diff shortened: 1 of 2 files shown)


packages/invoices/src/OutboundShipment/api/api.ts
```diff
@@ -237,16 +237,30 @@ export const getOutboundQueries = (sdk: Sdk, storeId: string) => ({
     const input = {
       insertOutboundShipmentLines: draftStocktakeLines
         .filter(
-          ({ type, isCreated }) =>
-            isCreated && type === InvoiceLineNodeType.StockOut
+          ({ type, isCreated, numberOfPacks }) =>
+            isCreated &&
+            type === InvoiceLineNodeType.StockOut &&
+            numberOfPacks > 0
```
joshxg, line 243: Not the most elegant code! suggestions welcome 🤔

packages/invoices/src/OutboundShipment/api/api.ts
```diff
@@ -276,12 +293,10 @@ export const getOutboundQueries = (sdk: Sdk, storeId: string) => ({
 
     return result;
   },
-  deleteLines: (invoiceId: string) => async (ids: string[]) => {
+  deleteLines: async (lines: { id: string; invoiceId: string }[]) => {
```
joshxg, line 296: I wanted to consolidate upsert/delete to use an array of line objects so changed the param on this

packages/invoices/src/OutboundShipment/api/hooks.ts
```diff
@@ -269,27 +267,19 @@ export const useSaveOutboundLines = () => {
...
-export const useDeleteInboundLine = (): UseMutationResult<
-  DeleteOutboundShipmentLinesMutation,
-  unknown,
-  string[],
-  { previous?: OutboundFragment; ids: string[] }
-> => {
-  // TODO: Shouldn't need to get the invoice ID here from the params as the mutation
-  // input object should not require the invoice ID. Waiting for an API change.
-  const { data } = useOutbound();
+export const useDeleteInboundLines = () => {
```
joshxg, line 270: which made me look at this and then I started.. tidying?

packages/invoices/src/OutboundShipment/api/hooks.ts
```diff
@@ -317,33 +311,19 @@ export const useDeleteSelectedLines = (): {
...
     } else {
-      return {
-        selectedRows: (
-          Object.keys(state.rowState)
-            .filter(id => state.rowState[id]?.isSelected)
-            .map(selectedId => lines.find(({ id }) => selectedId === id))
-            .filter(Boolean) as OutboundLineFragment[]
-        ).map(({ id }) => id),
-      };
+      return lines.filter(({ id }) => state.rowState[id]?.isSelected);
```
joshxg, line 326: this seems a little nicer!

packages/invoices/src/OutboundShipment/api/api.ts
mark-prins: uh, when testing an earlier PR and having problems.. I did exactly this! 😆

packages/invoices/src/OutboundShipment/api/hooks.ts
mark-prins: much 😄 
Might almost be small enough to 

```
    return isGrouped
      ? items
          ?.filter(({ id }) => state.rowState[id]?.isSelected)
          .map(({ lines }) => lines.flat())
          .flat()
      : lines.filter(({ id }) => state.rowState[id]?.isSelected);
```

mark-prins (approved):
🎉  it does change everything!!

packages/invoices/src/OutboundShipment/api/hooks.ts
joshxg: Nice! I see the suggestion in another PR also, will do both at once

PR #991, opened 2022-03-08
Add max MOS input and toggle for hiding lines over threshold

### LINKED ISSUE #576: Filter lines by being over the threshold MOS of a requisition
- Filter lines by being over the threshold MOS of a requisition
- Unsure where the filter is... maybe the `show lines with 0 stock` thing?
- https://docs.google.com/spreadsheets/d/1H0Rv1DEOgW0cE07wpfc_KivLf8TcNhYDqlwRp92x-To/edit#gid=0&range=A228
image

### LINKED ISSUE #579: Change the stock target of a supplier requisition
- Add a feature to change the stock target of a supplier requisition.
- This should also adjust all of the suggested quantities for each row
- https://docs.google.com/spreadsheets/d/1H0Rv1DEOgW0cE07wpfc_KivLf8TcNhYDqlwRp92x-To/edit#gid=0&range=A237
- Not sure about the pencil, but seems to be the top right. Maybe clicking is a select?
image

joshxg:
Fixes #579 

Should also close #576

- Went a little offroad on the labels 🤔 
- Pretty sure that's what the toggle is meant to do..


packages/requisitions/src/RequestRequisition/DetailView/Toolbar/Toolbar.tsx
```diff
@@ -23,55 +22,44 @@ export const Toolbar: FC = () => {
...
-                    <NameSearchInput
-                      type="supplier"
-                      disabled={isDisabled}
-                      value={otherParty ?? null}
-                      onChange={otherParty => {
-                        update({ otherParty });
-                      }}
-                    />
-                  }
+      <Grid container>
```
joshxg, line 25: removed some unneeded styles..

packages/requisitions/src/RequestRequisition/DetailView/Toolbar/ToolbarActions.tsx
```diff
@@ -1,49 +1,106 @@
 import React from 'react';
 import {
+  Switch,
```
joshxg, line 3: Maybe this would be better in the hoverclick popover thingy? Though, there's also an issue to add a filter on item name, so still need to add a search input on the left unless that's also hidden away

packages/requisitions/src/RequestRequisition/DetailView/Toolbar/ToolbarActions.tsx
```diff
@@ -1,49 +1,106 @@
...
+import {
+  useRequestFields,
+  useIsRequestDisabled,
+  useHideOverStocked,
+} from '../../api';
 
 const months = [1, 2, 3, 4, 5, 6];
 
 export const ToolbarActions = () => {
+  const { on, toggle } = useHideOverStocked();
```
joshxg, line 19: used zustand because I need this toggle to be the same as the one used in the hook. 

related https://github.com/openmsupply/openmsupply-client/issues/936

packages/common/src/intl/locales/en/replenishment.json
```diff
@@ -22,11 +22,14 @@
   "label.supplier-name": "Supplier Name",
   "label.supplier-ref": "Supplier Ref",
   "label.min-months-of-stock": "Minimum MOS",
+  "label.hide-stock-over-threshold": "Hide stock over threshold",
```
mark-prins, line 25: I'm wondering if this would be more helpful if it said `Hide stock over minimum`? That relates it to the 'Minimum MOS' value better than `threshold` maybe

mark-prins (approved):
well, that's awesome. 😁
It's like a real requisition form now!!
I think we'll need to get the in-house design team to think about the top area, there's a bunch of things on the RH side and maybe we could spread things out a bit :shrug:

packages/common/src/intl/locales/en/replenishment.json
joshxg: yes much better

joshxg:
Using a popover should fix it. Though, I think some chips or some other indication/feedback would be needed if the inputs are hidden away.

PR #997, opened 2022-03-08
Add disabling of edit modals in disabled documents

### LINKED ISSUE #984: Outbound / Stocktake edit modals disabled state.
Both outbound and stocktake edit modals show all stock lines for these edit modals, allowing the user to pick to issue or count from any.

When these modals are in a finalised state, this doesn't make sense.

So:
1. Disable opening the edit modals when a document is disabled
2. Show only the lines in the document in the edit modal

joshxg:
Fixes #984 

- Disables all the row clicks and 'add item' buttons when a document is disabled
- Offroad inbound disabling fix


packages/invoices/src/utils.ts
```diff
@@ -102,7 +102,6 @@ export const isOutboundDisabled = (outbound: OutboundRowFragment): boolean => {
 export const isInboundDisabled = (inbound: InboundRowFragment): boolean => {
   return (
     inbound.status === InvoiceNodeStatus.Verified ||
-    inbound.status === InvoiceNodeStatus.Shipped ||
```
joshxg, line 105: This was wrong. Shipped is the status of the inbound when it is shown in the receiving store. If it's disabled here, you can't move to 'delivered'

mark-prins: oh. oops, I missed that

mark-prins (approved):
👍  sweet!

PR #1017, opened 2022-03-10
Enable sort on internal orders

### LINKED ISSUE #1016: Internal order: sorting
The internal order detail view isn't sorting.

mark-prins:
Fixes #1016 
Ran into some bother with the `OK & Next` which was of course using the default sort state. Which means that if you sort the table on a different column, then the order of items in the useNext is incorrect.

I've fixed that by lifting the array of lines higher so the table and next hook use the same collection.
Outbound shipments has the same problem - thought I'd run this past you before fixing that one.
Response requisitions isn't sortable - so isn't a problem 😁

### DIFF
--- a/packages/requisitions/src/RequestRequisition/DetailView/ContentArea.tsx
+++ b/packages/requisitions/src/RequestRequisition/DetailView/ContentArea.tsx
@@ -1,19 +1,14 @@
-import React, { FC } from 'react';
-import { Column, DataTable, useTranslation } from '@openmsupply-client/common';
-import { RequestLineFragment } from '../api';
+import React from 'react';
+import { DataTable, useTranslation } from '@openmsupply-client/common';
+import { RequestLineFragment, useRequestLines } from '../api';
 
 interface ContentAreaProps {
-  columns: Column<RequestLineFragment>[];
-  lines: RequestLineFragment[];
   onRowClick: null | ((line: RequestLineFragment) => void);
 }
 
-export const ContentArea: FC<ContentAreaProps> = ({
-  columns,
-  lines,
-  onRowClick,
-}) => {
+export const ContentArea = ({ onRowClick }: ContentAreaProps) => {
   const t = useTranslation('common');
+  const { lines, columns } = useRequestLines();
   return (
     <DataTable
       onRowClick={onRowClick}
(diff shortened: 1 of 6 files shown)


joshxg (approved):
Nice! working great.

Yeah.. the sharing of that sort state is a bit of an oversight 😁  I think sorting/filtering is in a bit of a mess at the moment - would probably be one of the top of my list to refactor - I'm not sure if that would include using the URL.

I think what you have is fine, but I have been thinking about this a bit so I'm biased and have some thoughts 😀 

My personal preference is having the hook close to usage as possible and not passing props as much. I also think that the hook will likely be misused again in the future if just left alone.

Some possible alternatives:
- Use a shared state hook. There's plenty on npm and pretty easy to write
- Create a zustand store which can do all the sorting instead of `useSortBy`
- local storage: just use local storage to sync up the hooks state, using listeners on the LocalStorage interface we have.
- Use zustand to share state: Similar to local storage, but zustand instead. I had a play with this and it is working for sorting and the ok/next and is pretty simple and quick

```
type Store = {
  sortBy: SortBy<RequestLineFragment>;
  onChangeSortBy: (newSortBy: SortBy<RequestLineFragment>) => void;
};

const useStore = zustand<Store>(() => ({
  onChangeSortBy: () => {},
  sortBy: { key: 'name', isDesc: false, direction: 'asc' },
}));

const useSharedSortBy = () => {
  const { sortBy, onChangeSortBy } = useSortBy<RequestLineFragment>({
    key: 'itemName',
    isDesc: false,
  });
  const sharedSortBy = useStore();

  useEffect(() => {
    sharedSortBy.onChangeSortBy(sortBy);
    const unsub = useStore.subscribe(state => {
    // Might want to compare literals - isDesc and key instead.
      if (state.sortBy !== sharedSortBy.sortBy ) {
        sharedSortBy.onChangeSortBy(state.sortBy);
      }
    });
    return unsub;
  }, [sortBy]);

  return { sortBy, onChangeSortBy };
};

export const useRequestColumns = () => {
  const { sortBy, onChangeSortBy } = useSharedSortBy();
  const columns = useColumns<RequestLineFragment>(.
```

Though, might also want to put the columns in the store. I think there's a subtle bug where if you use `useColumns` twice, then the columns sort state will be out of wack, and if you try to sort on the same column, it won't know if it's ascending or descending.

mark-prins:
😆  I knew you'd hate the props approach would want a hook!
Was thinking to use a shared state, will put that in. I think I'll have to add columns, but will check that first.

PR #1022, opened 2022-03-10
#564 response requisition columns

### LINKED ISSUE #564: Add correct customer requisition columns
As per: https://docs.google.com/spreadsheets/d/1H0Rv1DEOgW0cE07wpfc_KivLf8TcNhYDqlwRp92x-To/edit#gid=320810900&range=222:222

At the moment the columns on a customer requisition aren't right.

I'm not 100% sure what the columns should be. For an ad hoc customer requisition, the columns might be different to those listed (from my experience with adding ad hoc customer reqs to mobile, we had to add many more columns).

So, would likely need to confirm the exact requirements for customer requisitions and whether ad hoc ones differ

mark-prins:
Fixes #564 
Adds the columns anyway. I'll do sorting separately. 
Battled with trying to get the column headers looking good, and almost managed to get them on one line - the `SOH` roll on to two lines on my laptop screen. maybe if the comment column was the speech bubble it'd be fine! 😆


packages/requisitions/src/ResponseRequisition/DetailView/columns.ts
```diff
@@ -21,6 +22,43 @@ export const useResponseColumns = ({
...
+        label: 'label.already-issued',
+        description: 'description.already-issued',
+        key: 'alreadyIssued',
+        width: 100,
+        align: ColumnAlign.Right,
+        accessor: ({ rowData }) =>
+          rowData.supplyQuantity - rowData.remainingQuantityToSupply,
+      },
+      {
+        label: 'label.remaining-to-supply',
```
mark-prins, line 54: some of these I thought to be too specific to add to the presets 🤷‍♂️

joshxg (approved):
Looking super slick!

PR #1024, opened 2022-03-10
Response sorting

### LINKED ISSUE #1023: Response Requisition: Detail sort
Be nice to have sorting on the columns of the response req

mark-prins:
Fixes #1023 
Adds sorting using the same approach as requests

### DIFF
--- a/packages/requisitions/src/ResponseRequisition/DetailView/ContentArea.tsx
+++ b/packages/requisitions/src/ResponseRequisition/DetailView/ContentArea.tsx
@@ -1,6 +1,5 @@
 import React from 'react';
 import { DataTable, useTranslation } from '@openmsupply-client/common';
-import { useResponseColumns } from './columns';
 import { useResponseLines, ResponseLineFragment } from '../api';
 
 interface ContentAreaProps {
@@ -8,8 +7,7 @@ interface ContentAreaProps {
 }
 
 export const ContentArea = ({ onRowClick }: ContentAreaProps) => {
-  const { lines, onChangeSortBy, sortBy } = useResponseLines();
-  const columns = useResponseColumns({ sortBy, onChangeSortBy });
+  const { columns, lines } = useResponseLines();
   const t = useTranslation('common');
 
   return (
--- a/packages/requisitions/src/ResponseRequisition/DetailView/columns.ts
+++ b/packages/requisitions/src/ResponseRequisition/DetailView/columns.ts
@@ -1,45 +1,83 @@
+import { useEffect } from 'react';
 import {
   useColumns,
   GenericColumnKey,
   SortBy,
-  Column,
   ColumnAlign,
+  zustand,
+  useSortBy,
 } from '@openmsupply-client/common';
 import { ResponseLineFragment } from './../api';
 
-interface UseResponseColumnOptions {
+type Store = {
   sortBy: SortBy<ResponseLineFragment>;
-  onChangeSortBy: (
-    column: Column<ResponseLineFragment>
-  ) => SortBy<ResponseLineFragment>;
-}
+  setSortBy: (sortBy: SortBy<ResponseLineFragment>) => void;
+};
 
-export const useResponseColumns = ({
-  onChangeSortBy,
-  sortBy,
-}: UseResponseColumnOptions): Column<ResponseLineFragment>[] =>
-  useColumns<ResponseLineFragment>(
+const useStore = zustand<Store>(set => ({
+  sortBy: { key: 'itemName', isDesc: false, direction: 'asc' },
+  setSortBy: (sortBy: SortBy<ResponseLineFragment>) =>
+    set(state => ({ ...state, sortBy })),
+}));
+
+const useSharedSortBy = () => {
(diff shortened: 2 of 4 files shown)


packages/requisitions/src/ResponseRequisition/DetailView/ContentArea.tsx
```diff
@@ -1,15 +1,13 @@
...
 import { useResponseLines, ResponseLineFragment } from '../api';
 
 interface ContentAreaProps {
   onRowClick: null | ((line: ResponseLineFragment) => void);
 }
 
 export const ContentArea = ({ onRowClick }: ContentAreaProps) => {
-  const { lines, onChangeSortBy, sortBy } = useResponseLines();
-  const columns = useResponseColumns({ sortBy, onChangeSortBy });
+  const { columns, lines } = useResponseLines();
```
mark-prins, line 10: not great - using columns from `useResponseColumns` won't work, after the column sort is changed the `useResponseLines` hook isn't rerendered and has the wrong sort value.
Just having another look at it..

mark-prins: I give up. fewer renders like this and it works, so I'll stick with it.

packages/requisitions/src/ResponseRequisition/DetailView/columns.ts
```diff
@@ -1,45 +1,83 @@
...
+          rowData.linkedRequisitionLine?.itemStats?.availableStockOnHand,
+        getSortValue: rowData =>
+          rowData.linkedRequisitionLine?.itemStats?.availableStockOnHand ?? '',
+
+        label: 'label.customer-soh',
+        description: 'description.customer-soh',
+        width: 100,
+        align: ColumnAlign.Right,
+      },
       'requestedQuantity',
```
joshxg, line 81: this guy needs a `getSortValue` (I've looked into defaulting `getSortValue` to the `accessor`, but the return types are different)

packages/requisitions/src/ResponseRequisition/DetailView/ContentArea.tsx
joshxg: Though I am kind of diggin the columns being all wrapped up nicely with the lines 🤔

packages/requisitions/src/ResponseRequisition/DetailView/columns.ts
```diff
@@ -1,45 +1,83 @@
...
+const useSharedSortBy = () => {
+  const sharedSortBy = useStore();
+  const { sortBy, onChangeSortBy } = useSortBy<ResponseLineFragment>(
+    sharedSortBy.sortBy
+  );
+
+  useEffect(() => {
+    sharedSortBy.setSortBy(sortBy);
+  }, [sortBy]);
+  return { sortBy, onChangeSortBy };
```
joshxg, line 32: Tricky one!

1. Render data with columns [from content area hook]
1. Trigger sort by change [from lines hook]
2. Changes hook specific state [lines hook state]
3. Effect changes zustand state [global]
4. returns hook specific state [lines hook state]
5. Lines hook reruns as zustand state changed 
6. Lines hook see the line hook specific state which hasn't changed
7. data doesn't re-render as the columns and data haven't changed

Can return zustand state instead
`{ sortBy: sharedSortBy.sortBy }`

which seems to be working (but I like what you've done moving the hook, seems nice)

packages/requisitions/src/ResponseRequisition/api/hooks.ts
```diff
@@ -105,22 +106,22 @@ export const useResponseFields = <
...
-    );
-  }, [sortBy, lines]);
-
-  return { lines: sorted, sortBy, onChangeSortBy };
+    const currentColumn = columns.find(({ key }) => key === sortBy.key);
+    const { getSortValue } = currentColumn ?? {};
+    return getSortValue
+      ? lines?.nodes.sort(getColumnSorter(getSortValue, !!sortBy.isDesc))
+      : lines?.nodes;
+  }, [sortBy.key, sortBy.isDesc, lines]);
```
joshxg, line 122: With this method of sorting, we re-order whenever the lines change, so when we open the modal and change a quantity, and we have sorted on quantity, the order changes. Is that good or bad?

joshxg (approved):
awesome, it's starting to look so real!

packages/requisitions/src/ResponseRequisition/api/hooks.ts
mark-prins: In some ways good - you finish editing and the table is sorted according to the values shown.
In other ways bad - if you OK & Next then you'll get a fairly surprising next line and the nexting will stop before or after you expect it to.
Overall.. ok for now

packages/requisitions/src/ResponseRequisition/DetailView/columns.ts
mark-prins: 🎉  thanks! I've shifted to use `sharedSortBy.sortBy` and that works nicely. now the `useColumns` hook can be used safely 😊

PR #1027, opened 2022-03-11
Master list detail view

### LINKED ISSUE #802: Master lists: Detail page
Name should be editable. more than that 🤷 
Should show a list of items. Unable to add / remove items at this stage, so will leave that off.

**Details**
Name ( read only atm, no mutation )
Description ( show in sidebar as 'comment'? )

**Columns**
- Item code
- Item name
- Item unit

mSupply has department, category, impreset quantity and price - these aren't available in API

mark-prins:
Fixes #802 

It might not be great, but it is something!
There's no mutations, so you cannot update the name or comment, or add/remove items.
You can however, copy to the clipboard! and sort columns! so it was totally worth it.

### DIFF
--- a/packages/system/src/MasterList/api/operations.graphql
+++ b/packages/system/src/MasterList/api/operations.graphql
@@ -1,7 +1,16 @@
+fragment MasterListItem on ItemNode {
+  __typename
+  id
+  code
+  name
+  unitName
+}
+
 fragment MasterListLine on MasterListLineNode {
+  __typename
   id
   item {
-    ...ItemRow
+    ...MasterListItem
   }
 }
 
--- a/packages/system/src/MasterList/api/operations.graphql
+++ b/packages/system/src/MasterList/api/operations.graphql
@@ -1,3 +1,23 @@
+fragment MasterListLine on MasterListLineNode {
+  id
+  item {
+    ...ItemRow
+  }
+}
+
+fragment MasterList on MasterListNode {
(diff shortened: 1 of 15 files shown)


packages/system/src/MasterList/api/operations.graphql
```diff
@@ -1,3 +1,23 @@
+fragment MasterListLine on MasterListLineNode {
+  id
+  item {
+    ...ItemRow
```
joshxg, line 4: This causes a bit of a funky import from the `Item` package. Since these aren't tree shaking properly right now, it doesn't make a difference but would cause the Item bundle to be bundled with this package even if shaken? Probably a later concern

joshxg (approved):
Lovely ❤️ 

Sucks about the URL. Basically means we need a unique human readable identifier on everything. Generic `code` is pretty easy to add and should be brought up soon. 

Even if the backend does a basic auto generate algo from the description/name when coming through sync 🤷

packages/system/src/MasterList/api/operations.graphql
mark-prins: ah. yes - sorry. forgot about that - I edited the generated operations file and meant to go back and just duplicate the `ItemRow` instead.

mark-prins:
yeah - the URL bugs me. Will make an issue for the API for the mutations and ask about a 'master list number'

PR #1034, opened 2022-03-13
#574 indicate internal suppliers

### LINKED ISSUE #574: When creating - have an indication if the selected supplier is internal or external
For supplier requisitions

requirement: https://docs.google.com/spreadsheets/d/1H0Rv1DEOgW0cE07wpfc_KivLf8TcNhYDqlwRp92x-To/edit#gid=0&range=A217

Since the requirement for selecting a customer for an outbound is not clear whether we should indicator transfer stores or on hold customers, I suppose this is unclear also?

joshxg:
Fixes #574 

- Not entirely happy but it's something. Just copying mSupply?

### DIFF
--- a/packages/common/src/ui/components/inputs/Autocomplete/AutocompleteList.tsx
+++ b/packages/common/src/ui/components/inputs/Autocomplete/AutocompleteList.tsx
@@ -11,9 +11,8 @@ import { AutocompleteOnChange, AutocompleteOptionRenderer } from './types';
 import { BasicTextInput } from '../TextInput';
 import { defaultOptionMapper, getDefaultOptionRenderer } from './utils';
 
-export type BaseAutocompleteListProps<T> = {
+export type AutocompleteListProps<T> = {
   options: T[];
-
   filterOptionConfig?: CreateFilterOptionsConfig<T>;
   loading?: boolean;
   loadingText?: React.ReactNode;
@@ -21,19 +20,10 @@ export type BaseAutocompleteListProps<T> = {
   onChange?: AutocompleteOnChange<T>;
   width?: number;
   height?: number;
-
+  renderOption?: AutocompleteOptionRenderer<T>;
+  optionKey?: keyof T;
   renderInput?: (params: AutocompleteRenderInputParams) => React.ReactNode;
 };
-
-export type AutocompleteListPropsWithOptionsRenderer<T> =
-  | {
-      renderOption: AutocompleteOptionRenderer<T>;
-      optionKey?: never;
-    }
-  | { renderOption?: never; optionKey: keyof T };
-
-export type AutocompleteListProps<T> = BaseAutocompleteListProps<T> &
-  AutocompleteListPropsWithOptionsRenderer<T>;
 
 export const AutocompleteList = <T,>({
   options,
--- a/packages/system/src/Name/Components/SupplierSearchInput/SupplierSearchInput.tsx
+++ b/packages/system/src/Name/Components/SupplierSearchInput/SupplierSearchInput.tsx
@@ -5,6 +5,8 @@ import {
   Typography,
   Autocomplete,
   useBufferState,
+  HomeIcon,
+  Box,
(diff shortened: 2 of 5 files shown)


packages/common/src/ui/components/inputs/Autocomplete/AutocompleteList.tsx
```diff
@@ -11,30 +11,20 @@ import { AutocompleteOnChange, AutocompleteOptionRenderer } from './types';
...
   filterOptionConfig?: CreateFilterOptionsConfig<T>;
   loading?: boolean;
   loadingText?: React.ReactNode;
   noOptionsText?: React.ReactNode;
   onChange?: AutocompleteOnChange<T>;
   width?: number;
   height?: number;
-
+  renderOption?: AutocompleteOptionRenderer<T>;
+  optionKey?: keyof T;
```
joshxg, line 24: I couldn't get these props to work well so I just made it easier 😁

packages/system/src/Name/Components/SupplierSearchInput/SupplierSearchInput.tsx
```diff
@@ -1,11 +1,33 @@
...
+  props,
+  item
+) => (
+  <DefaultAutocompleteItemOption {...props}>
+    <Typography
+      sx={{
+        marginInlineEnd: '10px',
+        fontWeight: 'bold',
+        width: 75,
+        color: item.store ? 'red' : 'inherit',
```
joshxg, line 23: Made the code red for an internal supplier / store like msupply does. Open to suggestions

mark-prins: I still see red as a warning indication and it's kind the reverse here.
How about a little store icon?

Screen Shot 2022-03-14 at 9 16 29 AM

or in the attractive colour known as 'secondary'?

Screen Shot 2022-03-14 at 9 15 37 AM

joshxg: Oo  nice idea. For the supplier list with both internal and external - no icon or an alternative?

mark-prins (approved):
👍  very nice!

packages/system/src/Name/Components/SupplierSearchInput/SupplierSearchInput.tsx
mark-prins: I was thinking no icon - highlights the internal more that way. Just a blank space the right width
Started with an 'external supplier' icon ( like the 'external link' one ) but thought it was better to highlight internal.

Side question - is there a way to get a list of all suppliers? The API is only giving me internal and I don't remember where that discussion ended up

mark-prins: Committed the suggestion to 09e71ba7 in case it is helpful.

joshxg: > Side question - is there a way to get a list of all suppliers? The API is only giving me internal and I don't remember where that discussion ended up


Just adding the filter `isSupplier: true` should be enough. 

Were you checking by using the list of suppliers when making a request requisition? If so, that is using internal suppliers only. Inbound should be using both

This query is giving me all suppliers on the demo server, internal and external

```
query MyQuery {
  names(storeId: "D77F67339BF8400886D009178F4962E1", filter: {isSupplier: true}) {
    ... on NameConnector {
      nodes {
        isSupplier
        isCustomer
        id
        store {
          id
        }
      }
    }
  }
}

```


> Committed the suggestion to https://github.com/openmsupply/openmsupply-client/commit/09e71ba78be5861ef34c0f12ce84cdd53cb7ed04 in case it is helpful.

joshxg: Nice.. much better than the red.

image

image

Using `Tamaki` on the demo server

PR #1042, opened 2022-03-14
#1038 organise utils

### LINKED ISSUE #1038: organise utils folder
- Add a `PricingUtils`
- Put in own folders

joshxg:
Fixes #1038 

- Sorry? :)
- This makes me feel better
- I like namespaces
- I made the utils folder have a bunch of folders each exposing an `XUtils` object (mainly)). Hopefully you like it heh. I always find when working in a code base it's often difficult to know where to put common utils. Hopefully this makes it clear where things should go
- Maybe better would be each `Util` or `Formatter` etc object should probably be a function `get...` and then also have a hook `useFormatter` for example. Then you could pass in a pref for tax decimals for example.. or `QuantityUtils.suggestedQuantity` could change depending on a pref


packages/common/src/utils/index.ts
```diff
@@ -1,14 +1,16 @@
...
-export * from './debounce';
-export * from './dateFunctions';
+export * from './dates';
 export * from './arrays';
 export * from './regex';
+export * from './pricing';
+export * from './functions';
+export * from './navigation';
 
 export type UUID = string;
```
joshxg, line 15: There's some extras in here that I wanted to organise but I ran out of ideas. Got any?

mark-prins: do you mean the rest of the bits in the index?

- `UUID` could be in types
- `isTypeOf` could be in types as well. it is a util fn too though, so 🤷 
- `isEqual` in functions
- `generateUUID` in functions or a `identifiers` along with `UUID` as a type
- `isProduction` ?? `runtime` / `environment` or just stuff it into functions?

mark-prins (approved):
👍  wow - that's some tidying! 
Looking very organised 😁

joshxg:
I went with even more objects.. `type`, `object` and `environment`.....

🙄

PR #1043, opened 2022-03-14
Add adjustments to react-query props

### LINKED ISSUE #1001: I have made an assumption and it's turned out to be..
FALSE!

I assumed the default for `notifyOnChangeProps` https://react-query.tanstack.com/reference/useQuery
```
Optional
If set, the component will only re-render if any of the listed properties change.
If set to ['data', 'error'] for example, the component will only re-render when the data or error properties change.
If set to "tracked", access to properties will be tracked, and the component will only re-render when one of the tracked properties change
```

was `"tracked"`.

It's not! it's undefined.

There's a PR to make it the default. But that was understandably, made for their v4 release as it is a breaking change.

What this means is when doing:

```
const { onHold } = useInboundFields("onHold")
```

I had assumed only a re-render when `onHold` changes. But this is not right. The selector *does* prevent re-renders from *other* data changing, however, since it is not using 'tracked' for `notifyOnChangeProps`, `isLoading` triggers re-renders.

Oh no!

So, because of this CARELESS assumption, will need to go through and check uses and add `tracked` in places. Starting with `useQuerySelector` - the main hook I had this assumption with

joshxg:
Fixes #1001 

- Adds a default option to the react query client for `notifyOnChangeProps`
- Also adds some props to the`useDetailDocument` type hooks

### DIFF
--- a/packages/inventory/src/Stocktake/api/hooks.ts
+++ b/packages/inventory/src/Stocktake/api/hooks.ts
@@ -53,8 +53,15 @@ const useStocktakeNumber = () => {
 export const useStocktake = (): UseQueryResult<StocktakeFragment> => {
   const stocktakeNumber = useStocktakeNumber();
   const api = useStocktakeApi();
-  return useQuery(api.keys.detail(stocktakeNumber), () =>
-    api.get.byNumber(stocktakeNumber)
+  return useQuery(
+    api.keys.detail(stocktakeNumber),
+    () => api.get.byNumber(stocktakeNumber),
+    // Don't refetch when the edit modal opens, for example. But, don't cache data when this query
+    // is inactive. For example, when navigating away from the page and back again, refetch.
+    {
+      refetchOnMount: false,
+      cacheTime: 0,
+    }
   );
 };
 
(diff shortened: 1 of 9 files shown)


packages/inventory/src/Stocktake/api/hooks.ts
```diff
@@ -53,8 +53,15 @@ const useStocktakeNumber = () => {
...
-  return useQuery(api.keys.detail(stocktakeNumber), () =>
-    api.get.byNumber(stocktakeNumber)
+  return useQuery(
+    api.keys.detail(stocktakeNumber),
+    () => api.get.byNumber(stocktakeNumber),
+    // Don't refetch when the edit modal opens, for example. But, don't cache data when this query
+    // is inactive. For example, when navigating away from the page and back again, refetch.
+    {
+      refetchOnMount: false,
+      cacheTime: 0,
```
joshxg, line 63: without this, whenever the hook `useStocktake()` is mounted, it will refetch and It's used a decent amount! An alternative is to make another hook, but I think this works well, though having the stocktake refetch on the modal opening can sometimes be good. Can revert this or change it up if it's concerning

mark-prins (approved):
🎉  that's so cool! I had wondered about the fetching and never looked into it!
nice work

joshxg:
Yup it's pretty awesome. There are so many people using `react-query`, gathering so many use cases from everyone, there's really not many that it can't do!

Issue #1006, opened 2022-03-09
Update store queries


joshxg:
New API functionality! Good times. 
Would like to make use of these:

- [ ] Add store sorting through API
- [ ] Use the `name` rather than the `code`

PR #1045, opened 2022-03-14
#941 Shift store drop down

### LINKED ISSUE #941: Shift store drop down / autocomplete from login to `Store` package
So that we can reuse it more easily (i.e. I wanted to quickly throw it in to the footer to be able to swap stores easily)

joshxg:
Fixes #941 and #1006 a little bit

- Shifted the store drop down
- Added a default sort to store query - too hacky?
- Getting the name is more annoying - it requires a store ID..
- Then I had an idea and couldn't help myself with `LoginLayout`

### DIFF
--- a/packages/common/src/authentication/AuthContext.tsx
+++ b/packages/common/src/authentication/AuthContext.tsx
@@ -15,6 +15,7 @@ type User = {
 };
 
 interface Store {
+  __typename: 'StoreNode';
   id: string;
   code: string;
 }
@@ -106,7 +107,7 @@ export const AuthProvider: FC = ({ children }) => {
       user: { id: '', name: username },
     };
 
-    setMRUCredentials({ username, store: store });
+    setMRUCredentials({ username, store });
     if (!!token) setLocalStore(store);
     setLocalToken(token);
     setAuthCookie(authCookie);
@@ -118,7 +119,7 @@ export const AuthProvider: FC = ({ children }) => {
     if (!localToken) return;
 
     setLocalStore(store);
-    setMRUCredentials({ username: user?.name ?? '', store: store });
+    setMRUCredentials({ username: user?.name ?? '', store });
     const authCookie = getAuthCookie();
     setAuthCookie({ ...authCookie, store });
   };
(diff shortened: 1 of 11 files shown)


packages/common/src/authentication/AuthContext.tsx
```diff
@@ -15,6 +15,7 @@ type User = {
 };
 
 interface Store {
+  __typename: 'StoreNode';
```
joshxg, line 18: as to use `StoreRowFragment` generated type instead of the `Store` types.. 🤔  

Bad idea?

packages/host/src/components/Login/Login.tsx
```diff
@@ -1,20 +1,17 @@
...
   ArrowRightIcon,
-  Box,
-  Stack,
-  Typography,
-  Autocomplete,
   useTranslation,
   AutocompleteRenderInputParams,
-  defaultOptionMapper,
   LoadingButton,
+  Box,
```
joshxg, line 7: I was getting a little confused with the jsx. These complicated layouts can get so annoying! I had an idea - it's the last commit so I can revert if it doesn't help.

Other than `LoginLayout`, I also started using `StoreSearchInput`

packages/host/src/components/Login/LoginLayout.tsx
```diff
@@ -0,0 +1,101 @@
+import React from 'react';
```
joshxg, line 1: I didn't actually change any of the styling, I just copy pasted and replaced the component with a prop in an effort to separate the ui/styling and the more logically stuffs

packages/common/src/authentication/AuthContext.tsx
mark-prins: I did wonder about doing that. didn't like having the separate types

packages/host/src/components/Login/LoginLayout.tsx
mark-prins: it does make the jsx easier to read here, having the components reduced down like this

mark-prins (approved):
👍  yeah, nice - thanks!

PR #1046, opened 2022-03-14
#1014 filter master lists

### LINKED ISSUE #1014: Filtering master lists
Master list queries need some filters:

- list view: own store
- supplier requisition: own store

joshxg:
Fixes #1014 

- Pretty simple update. If adding an outbound add from master list mutation, would want to use a filter of both `existsForName` (with the customer nameID) and `existsForStore`


mark-prins (approved):
😁

PR #1047, opened 2022-03-14
Add default language

### LINKED ISSUE #1037: Setting the language for new users
When a user is logging in for the first time they'll get this (in dev, in prod will default to en)

image

No language has been selected:

image

Should default to something?

joshxg:
Fixes #1037 

- Sets a default language for users to en

### DIFF
--- a/packages/common/src/intl/intlHelpers.ts
+++ b/packages/common/src/intl/intlHelpers.ts
@@ -20,6 +20,10 @@ export interface TypedTFunction<Keys> {
   ): string;
 }
 
+export const isSupportedLang = (lang: string): boolean => {
+  return lang === 'en' || lang === 'fr' || lang === 'ar';
+};
+
 export const useTranslation = (ns?: Namespace): TypedTFunction<LocaleKey> => {
   const { t } = useTranslationNext(ns);
   return (key, options) => (key ? t(key, options) : '');
--- a/packages/common/src/intl/intlHelpers.ts
+++ b/packages/common/src/intl/intlHelpers.ts
@@ -77,3 +77,9 @@ export const useUserName = (): string => {
   const { user } = useAuthContext();
   return user?.name ?? '';
 };
+
+// TODO: When the server supports a query to find the deployments
+// default language, use a query to fetch it.
+export const useDefaultLanguage = (): SupportedLocales => {
+  return 'en';
+};
--- a/packages/common/src/intl/intlHelpers.ts
+++ b/packages/common/src/intl/intlHelpers.ts
@@ -4,7 +4,10 @@ import { i18n, TOptions } from 'i18next';
 import { LocaleKey } from './locales';
 import { useAuthContext } from '../authentication';
 
-export type SupportedLocales = 'en' | 'fr' | 'ar';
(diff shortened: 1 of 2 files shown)


packages/common/src/intl/intlHelpers.ts
```diff
@@ -20,6 +20,10 @@ export interface TypedTFunction<Keys> {
   ): string;
 }
 
+export const isSupportedLang = (lang: string): boolean => {
+  return lang === 'en' || lang === 'fr' || lang === 'ar';
```
CarlosNZ, line 24: Is there a way to request from the server what languages are supported? Ideally these wouldn't be hardcoded. 

Do you think it's worth making a request to try and load the selected language and then defaulting to "en" if that query fails?

joshxg: > Is there a way to request from the server what languages are supported? Ideally these wouldn't be hardcoded.

Nah, there isn't. 

At the moment we'd still need to commit the language files to the repo, so, avoiding hard coding doesn't give you much. Language support is also rare enough [at the moment] that creating a more dynamic system would cost much more than a little bit of hard coded updates.

> Do you think it's worth making a request to try and load the selected language and then defaulting to "en" if that query fails?

Soz - I'm not getting this one. Request with what to where? I thought maybe you were saying a request to fetch the default language from the server. That query isn't supported, so no point sending one. 

Though I did think setting up for the future was a good idea so I have made a hook `useDefaultLanguage` which can be modified in the future to make a query

CarlosNZ: > Soz - I'm not getting this one. Request with what to where? I thought maybe you were saying a request to fetch the default language from the server. That query isn't supported, so no point sending one.

Sorry, my misunderstanding of where the language strings are coming from. I think what I was saying is, is it possible, instead of having hard-coded "supported" languages, can you try and load whatever language and then switch to default "en" if that fails.

```
const { language } = i18n;
    if (!isSupportedLang(language)) i18n.changeLanguage(defaultLanguage);
```

What happens if you do `i18n.changeLanguage(language)` on a language that is not supported? Do you get an error? If so, what about:
```
try {
  i18n.changeLanguage(language)
} catch {
  i18n.changeLanguage(defaultLanguage)
}

Or something to that effect?

Comment (joshxg) [reply packages/common/src/intl/intlHelpers.ts:24]: doesn't throw

we know exactly the languages supported so we can be explicit

Comment (mark-prins) [reply packages/common/src/intl/intlHelpers.ts:24]: How about unifying with `SupportedLocales`?

```
const locales = ['en', 'ar', 'fr'];

export type SupportedLocales = typeof locales[number];
export const isSupportedLang = (lang: string): boolean =>
  locales.some(locale => lang === locale);
```

then the list is just in one place 😁

mark-prins (approved):
👍  LGTM!

packages/common/src/intl/intlHelpers.ts
joshxg: sweet. Added some `const`s  - `typeof locales[number]` just resolves to a string

PR #1054, opened 2022-03-15
#1052 Clear filter when filter input empty

### LINKED ISSUE #1052: Filter not being cleared when set back to empty string
From [this comment](https://github.com/openmsupply/openmsupply-client/issues/940#issuecomment-1067455581), but adding as new issue as this is a simple fix, whereas that issue is a more comprehensive one that requires further discussion:

> 
> When "searching by comment" on the Outbound/Inbound Shipment pages, when I type something in the search box that has no result, the list is empty as expected. However, when I delete the search query, the original list doesn't come back -- it stays empty.
> 
> Actually -- even when typing search that matches something, when clearing the search only items that have something in their "comment" field show up in the list -- items with empty comments don't return.
> 
> Also -- seems to be only Shipment pages. Searching by comment working as expected on others I tried (e.g. Internal Orders)
> 

Turns out that when clearing the input, the filter field persists but just with empty string value. Simple fix is to explicitly clear the field from the filter object. Oddly, this problem doesn't occur on other filterable queries, so there must be some inconsistency in the back-end regarding how empty string fields are handled in filters.

CarlosNZ:
Fix #1052 

Explicitly clears the filter field when the filter string is empty.

### DIFF
--- a/packages/common/src/hooks/useFilterBy/useFilterBy.tsx
+++ b/packages/common/src/hooks/useFilterBy/useFilterBy.tsx
@@ -45,8 +45,14 @@ export const useFilterBy = (initialFilterBy?: FilterBy | null): FilterState => {
     condition: FilterByConditionByType['string'],
     value: string
   ) => {
-    const newFilter = { [key]: { [condition]: value } };
-    setFilterBy({ ...filterBy, ...newFilter });
+    if (value === '') {
+      const newFilter = { ...filterBy };
+      delete newFilter[key];
+      setFilterBy(newFilter);
+    } else {
+      const newFilter = { [key]: { [condition]: value } };
+      setFilterBy({ ...filterBy, ...newFilter });
+    }
   };
 
   const onChangeDateFilterRule = (
--- a/packages/common/src/hooks/useFilterBy/useFilterBy.tsx
+++ b/packages/common/src/hooks/useFilterBy/useFilterBy.tsx
@@ -46,9 +46,7 @@ export const useFilterBy = (initialFilterBy?: FilterBy | null): FilterState => {
     value: string
   ) => {
     if (value === '') {
-      const newFilter = { ...filterBy };
-      delete newFilter[key];
-      setFilterBy(newFilter);
+      onClearFilterRule(key);
     } else {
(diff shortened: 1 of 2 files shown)


packages/common/src/hooks/useFilterBy/useFilterBy.tsx
```diff
@@ -59,7 +63,9 @@ export const useFilterBy = (initialFilterBy?: FilterBy | null): FilterState => {
   };
 
   const onClearFilterRule = (key: string) => {
-    setFilterBy({ ...filterBy, [key]: null });
+    const newFilter = { ...filterBy };
+    delete newFilter[key];
+    setFilterBy(newFilter);
```
CarlosNZ, line 68: This is not strictly necessary for this fix. However, the filter plugin I'm used to using for Postgraphile was *extremely* fussy and would throw its toys when you passed it null values. So just feels safer to explicitly delete the key rather than set it to `null`.

packages/invoices/src/OutboundShipment/ListView/Toolbar.tsx
```diff
@@ -18,7 +18,7 @@ export const Toolbar: FC<{
   const onDelete = useDeleteSelectedOutbounds();
 
   const key = 'comment' as keyof OutboundRowFragment;
-  const filterString = filter.filterBy?.[key]?.like as string;
+  const filterString = (filter.filterBy?.[key]?.like as string) || '';
```
CarlosNZ, line 21: This just prevents this React error:
Screen Shot 2022-03-15 at 10 31 50 AM

packages/common/src/hooks/useFilterBy/useFilterBy.tsx
joshxg: Nice!

packages/invoices/src/OutboundShipment/ListView/Toolbar.tsx
joshxg: Oh yes, very annoying. I cause this mui error a lot 😁

joshxg (approved):
Nice! This hook is so smelly. Nice work navigating it so cleanly

PR #1058, opened 2022-03-15
Add default to true when creating a stocktake line from a stockline

### LINKED ISSUE #1057: Can't create stocktake lines
When trying to insert a stocktake line, an update mutation is sent instead - oopsies 😁

joshxg:
Fixes #1057 

- Also an off-road `isProduction` bug that I added with a PR recently

### DIFF
--- a/packages/host/src/Host.tsx
+++ b/packages/host/src/Host.tsx
@@ -18,7 +18,7 @@ import {
   ConfirmationModalProvider,
   AuthProvider,
   AlertModalProvider,
-  isProduction,
+  EnvUtils,
 } from '@openmsupply-client/common';
 import { AppRoute, Environment } from '@openmsupply-client/config';
 import { Login, Viewport } from './components';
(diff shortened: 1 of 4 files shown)


packages/host/src/Host.tsx
```diff
@@ -29,8 +29,8 @@ const queryClient = new QueryClient({
     queries: {
       // These are disabled during development because they're
       // annoying to have constantly refetching.
-      refetchOnWindowFocus: isProduction(),
-      retry: isProduction(),
+      refetchOnWindowFocus: EnvUtils.isProduction(),
+      retry: EnvUtils.isProduction(),
```
joshxg, line 33: Offroad bug fix from my merge

packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/hooks.ts
```diff
@@ -62,8 +62,8 @@ const stockLineToDraftLine = (
 ): DraftStocktakeLine => {
   return {
     stocktakeId,
-    countThisLine: false,
-    isCreated: false,
+    countThisLine: true,
+    isCreated: true,
```
joshxg, line 66: default to `isCreated` for lines with a stocktake line. If the user doesn't want to count it, then `countThisLine` will be false.

using `countThisLine` as true is questionable. It means when you add an item, you're counting all the stock lines in the store by default.. 🤔

joshxg:
At the moment, the lines are blank when they're not counted.. so that is a bit of a terrible user experience. Also there's a bug where counted stock lines are doubled up. I'll look at those separately - probably best to get this in

mark-prins (approved):
🎉  I have no idea why this is required.. but it fixes the problem 😁

joshxg:
😁 

the `isCreated` / `isUpdated` flags determine which mutation to send later on and the `isCreated` flag wasn't being set

PR #1060, opened 2022-03-15
#940 Search by name (shipments)

### LINKED ISSUE #940: Shipments: Search by other party name
Currently the Inbound and Outbound shipments only allow filtering the list by comment.
We'd like to filter by `Name` aka `OtherPartyName` - as I'm told this is more useful. 

Personally I'm finding the comment useful when testing.. so I'm now wondering about whether we do the advanced filtering and allow filtering on multiple fields. Also wondering if this is still `v1` territory now. I think push it down the list a bit and do it after some of the other work.

CarlosNZ:
Fix #940 

Just the simple "only search by name" option -- needs back-end changes to filtering to be able to search both comment and name.

### DIFF
--- a/packages/common/src/intl/locales/en/distribution.json
+++ b/packages/common/src/intl/locales/en/distribution.json
@@ -83,5 +83,6 @@
   "messages.must-allocate-all-lines": "Cannot change the status until all lines have been allocated.",
   "messages.select-rows-to-delete": "Select rows to delete",
   "messages.shipment-saved": "Shipment saved 🥳",
-  "warning.nothing-to-supply": "Nothing left to supply!"
+  "warning.nothing-to-supply": "Nothing left to supply!",
+  "placeholder.search-by-name": "Search by name"
 }
--- a/packages/common/src/intl/locales/en/distribution.json
+++ b/packages/common/src/intl/locales/en/distribution.json
@@ -83,6 +83,5 @@
   "messages.must-allocate-all-lines": "Cannot change the status until all lines have been allocated.",
   "messages.select-rows-to-delete": "Select rows to delete",
   "messages.shipment-saved": "Shipment saved 🥳",
-  "warning.nothing-to-supply": "Nothing left to supply!",
-  "placeholder.search-by-name": "Search by name"
+  "warning.nothing-to-supply": "Nothing left to supply!"
 }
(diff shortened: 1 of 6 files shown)


packages/invoices/src/InboundShipment/ListView/Toolbar.tsx
```diff
@@ -13,11 +13,11 @@ import { useDeleteSelectedInbounds, InboundRowFragment } from '../api';
 export const Toolbar: FC<{
   filter: FilterController;
 }> = ({ filter }) => {
-  const t = useTranslation();
+  const t = useTranslation('distribution');
```
mark-prins, line 16: :nit inbound shipments are in `replenishment` and should be using translations from that file if they are specific.. or the translation should be in `common` if used in various places. 

I think this one should be in `common` as it will be used elsewhere. In which case you don't need to specify a namespace when calling `useTranslation` as it defaults to `common`

packages/common/src/intl/locales/en/distribution.json
```diff
@@ -83,5 +83,6 @@
   "messages.must-allocate-all-lines": "Cannot change the status until all lines have been allocated.",
   "messages.select-rows-to-delete": "Select rows to delete",
   "messages.shipment-saved": "Shipment saved 🥳",
-  "warning.nothing-to-supply": "Nothing left to supply!"
+  "warning.nothing-to-supply": "Nothing left to supply!",
+  "placeholder.search-by-name": "Search by name"
```
mark-prins, line 87: I think this is used by more than just the `distribution` namespace and should be in `common`

mark-prins (approved):
👍  nice! working as advertised 😁 
one minor thing: I'd prefer the translation in common, as noted

packages/invoices/src/InboundShipment/ListView/Toolbar.tsx
CarlosNZ: Yeah, sure was wondering about that. There's actually an existing "Search by name" string in "inventory" as well -- should I consolidate them both into `common`?

mark-prins: thanks - sounds good.
I do wonder about running some sort of usage check to determine which translation strings are unused 🤔
surely there's a script out there to do that..

CarlosNZ: > should I consolidate them both into `common`?

Done 8a98121d

> I do wonder about running some sort of usage check to determine which translation strings are unused 🤔
> surely there's a script out there to do that..

Hmmmm... https://github.com/glebm/i18n-tasks

Will try that out later on.

PR #1061, opened 2022-03-15
Tooltips on headers

### LINKED ISSUE #1051: Can't add tooltips to non-sortable column headers
in `Header.tsx`, the tooltip is not applied when a row is not sortable as `Tooltip` needs to be able to inject a ref into it's child, but the child of tooltip when the column is not sortable is `Header`, which does not use `forwardRef`.

Can fix this by either ensuring `Header` uses `forwardRef` (bigger change, typings might be annoying) or wrapping `{Header}` in a component which is capable of taking a ref. i.e. `<div/>` or `<Box/>`

With this fix, the tooltip will show a small blank tooltip when no description is passed which will also need to be resolved

mark-prins:
Fixes #1051 
Don't show a tooltip if there is nothing to say; and forward the ref to the non-sortable header if there is a description.

### DIFF
--- a/packages/common/src/ui/layout/tables/components/Header/Header.tsx
+++ b/packages/common/src/ui/layout/tables/components/Header/Header.tsx
@@ -50,12 +50,30 @@ export const HeaderCell = <T extends RecordWithId>({
     150
   );
 
-  const tooltip = (
+  const showTooltip = !!description || sortable;
+  const tooltip = showTooltip ? (
     <>
       {!!description && <div>{t(description)}</div>}
       {sortable && <div>{t('label.click-to-sort')}</div>}
     </>
+  ) : (
+    ''
   );
+  const HeaderLabel = sortable ? (
+    <TableSortLabel
+      hideSortIcon={false}
+      active={isSorted}
+      direction={direction}
+      IconComponent={SortDescIcon}
+    >
+      <Header column={column} />
+    </TableSortLabel>
+  ) : (
+    <div>
+      <Header column={column} />
+    </div>
+  );
+
   return (
     <TableCell
       role="columnheader"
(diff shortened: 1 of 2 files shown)


packages/common/src/ui/layout/tables/components/Header/Header.tsx
```diff
@@ -50,12 +50,30 @@ export const HeaderCell = <T extends RecordWithId>({
     150
   );
 
-  const tooltip = (
+  const showTooltip = !!description || sortable;
+  const tooltip = showTooltip ? (
```
mark-prins, line 54: simple way to not display Tooltips - they never show for zero length strings

joshxg: Nice!

packages/common/src/ui/layout/tables/components/Header/Header.tsx
```diff
@@ -82,18 +100,7 @@ export const HeaderCell = <T extends RecordWithId>({
...
-            active={isSorted}
-            direction={direction}
-            IconComponent={SortDescIcon}
-          >
-            <Header column={column} />
-          </TableSortLabel>
-        ) : (
-          <Header column={column} />
-        )}
+        {HeaderLabel}
```
joshxg, line 103: Nice. I originally found this because I was thinking about adding them to the edit modal column headers to explain things like what 'available' and 'on hold' meant.

Also now that i'm thinking about tooltips... :)

Maybe a good idea to wrap the `BasicCell` (and maybe some other cells) in a tooltip which displays if the text has overflowed (and is using ellipsis for text overflow) to show the full value. For example the `locationName` doesn't have too much room I could imagine that could be quite annoying.

joshxg (approved):

packages/common/src/ui/layout/tables/components/Header/Header.tsx
mark-prins: good idea! will merge this one, to get rid of those warnings and have a look at a `BasicCell` overflow tooltip

PR #1064, opened 2022-03-15
#1059 Doubled Stocktake rows

### LINKED ISSUE #1059: Double counted stocktake lines
A new uncounted line shows up for every counted line in a stocktake that has a stock line. 

Also, when first adding an item all the lines are blank and so you'd have no idea which to actually add to the stocktake. Not fully sure what to do about this!

joshxg:
Fixes #1059 

- When inserting a stocktake line which was counting an existing stockline, the stock line ID was not being sent, so the API assumed this was a 'new' row created for the item.
- Then I did some refactoring and shuffling things around..


packages/common/src/utils/arrays/ArrayUtils.ts
```diff
@@ -34,6 +35,20 @@ export const ArrayUtils = {
   ): number => {
     return arr.reduce((sum, someEntity) => sum + someEntity[key], 0);
   },
+  immutableUpdateById: <T extends RecordWithId>(
```
mark-prins, line 38: 😁

mark-prins (approved):
🎉  nice! 
and the hooks split out too.. an added bonus 😁

packages/common/src/utils/arrays/ArrayUtils.ts
joshxg: I'm unsure about the name! It's too long :(

Was thinking of some alternatives like `updateWithPatch`.. `patchUpdate` .. 🙄

mark-prins: `patch` can be a verb too so you could shorten by dropping the update, if that's not confusing?
- `patchById`
- `patchImmutable`

joshxg: Oh, yes! nice, thanks!

PR #1066, opened 2022-03-16
#1029 refresh token

### LINKED ISSUE #1029: Refresh Token
Functionality needs reviewing - I _think_ that it isn't refreshing and/or isn't timing users out when the cookie expires.
Would like to test the functionality and maybe, who knows, fix it? If broken.

mark-prins:
Fixes #1029 

The token returned by the refresh query was not being saved, which meant that subsequent refresh requests used the wrong token and failed.

Have also added an alert to catch the state where the cookie has expired and the user is no longer in the auth context. Now shows:

Screen Shot 2022-03-16 at 2 54 59 PM

Have been unable to test locally, so have used the demo server. the refresh is better, but failing on the second refresh request, this returns an `InvalidToken` error.


packages/common/src/authentication/AuthContext.tsx
```diff
@@ -70,10 +70,14 @@ const setAuthCookie = (cookie: AuthCookie) => {
...
-const useRefreshingAuth = (token?: string) => {
+const useRefreshingAuth = (
+  callback: (token?: string) => void,
+  token?: string
+) => {
   const { setHeader } = useGql();
   setHeader('Authorization', `Bearer ${token}`);
-  useGetRefreshToken(token ?? '');
+  const { data } = useGetRefreshToken(token ?? '');
+  useEffect(() => callback(data?.token), [data, callback]);
```
mark-prins, line 80: had to wrap in a callback, as the auth context was being re-rendered a lot

packages/common/src/authentication/AuthContext.tsx
```diff
@@ -99,8 +106,12 @@ export const AuthProvider: FC = ({ children }) => {
   const [localStore, setLocalStore] = useState<Store | undefined>(cookieStore);
   const [localToken, setLocalToken] = useState<string | undefined>(cookieToken);
   const storeId = localStore?.id ?? '';
-
-  useRefreshingAuth(localToken);
+  const saveToken = (token?: string) => {
+    setLocalToken(token);
+    const authCookie = getAuthCookie();
+    setAuthCookie({ ...authCookie, token: token ?? '' });
```
joshxg, line 112: Working nicely!

-----
Stop reading
-----
Random thought dump

 I had thoughts around abstracting some of this away into a `useAuthCookie` hook.. maybe using zustand so it could be used in multiple places (useAuthToken?) but I am not sure it's worth it. 

For example

```
const useAuthCookie = zustand({ ... })
const { store, token, setStore, setToken, setCookie } = useAuthCookie()

useAuthCookie.subscribe(() => {
    // listen for changes and update the actual cookie in the browser
})
```

Though, since there's also the `AuthContext`.. maybe there's a nice way where instead of the `useAuthCookie`, just adding a subscriber to that state to update the cookie... maybe even adding https://www.npmjs.com/package/zustand-persist (Oh, you're still reading? I did stay stop!)

I'm not sure, but I think the store is in a `useState`, in the Context object, in the MRU creds, in a cookie and in the bearer token set in the auth header and there has to be a way to consolidate some of it

packages/common/src/authentication/AuthContext.tsx
```diff
@@ -7,6 +7,7 @@ import { useGql } from '../api';
 import { useGetRefreshToken } from './api/hooks';
 import { useGetAuthToken } from './api/hooks/useGetAuthToken';
 import { AuthenticationResponse } from './api';
+import { useDebounceCallback } from '..';
```
joshxg, line 10: nit: `@common/hooks`

joshxg (approved):
awesome, too bad about the api 😁

packages/common/src/authentication/AuthContext.tsx
mark-prins: oh, that's good. a fine idea - actually a couple of good ideas in there. will pop this pr in and make an issue

packages/common/src/authentication/api/hooks/useGetRefreshToken.ts
```diff
@@ -1,11 +1,23 @@
...
   const api = useAuthApi();
+  const [enabled, setEnabled] = useState(false);
+  const refetchInterval = Math.max(COOKIE_LIFETIME_MINUTES - 1, 1) * 60 * 1000;
+  const debouncedSetEnabled = useDebounceCallback(
+    setEnabled,
+    [token],
+    refetchInterval
+  );
+
+  debouncedSetEnabled(!!token);
```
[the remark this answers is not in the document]
mark-prins: I wondered about that and just hadn't got to the point of experimenting.

PR #1068, opened 2022-03-16
#500 Keyboard shortcuts

### LINKED ISSUE #500: Data table keyboard navigation

CarlosNZ:
Fix #500 

Implements data table and breadcrumb keyboard shortcuts, using existing K-bar package.

Adding as Draft PR for now to get some initial feedback -- see inline for additional notes / questions

Keyboard shortcuts:
`<Escape>` -- go "up" one breadcrumb / url level (almost but not quite the same as "back")
`<Arrow up>` / `<Arrow down>` -- move focus up/down navigation table (only active in Data tables)
`<Enter>` -- "click" focused data table row (only active in Data tables)

Created a new hook "useBreadcrumbs" and moved the `urlParts` logic into it, as well as adding new breadcrumb navigation functions. Might be better to put this into a Context somewhere so we only ever have one instance of it?

Still to-do:
- [x] Styling of "focused" row should match the "hover" style


packages/common/src/hooks/useBreadcrumbs/useBreadcrumbs.tsx
```diff
@@ -0,0 +1,52 @@
...
+    if (navAction === 'up-one') navigateUpOne();
+    setNavAction(null);
+  }, [navAction]);
+
+  const navigateUpOne = () => {
+    console.log('Up one breadcrumb');
+    console.log('parts', urlParts);
+    if (urlParts.length < 2) return;
+    navigate(urlParts[urlParts.length - 2]?.path as string);
+  };
```
CarlosNZ, line 49: Initially tried to return the "navigateUpOne" function directly without having `useEffect`. Problem was the kbar Action would get a previous "urlParts" state (presumably a stale closure). Open to other suggestions?

packages/common/src/ui/layout/tables/DataTable.tsx
```diff
@@ -32,10 +33,44 @@ export const DataTableComponent = <T extends RecordWithId>({
   isDisabled = false,
 }: TableProps<T>): JSX.Element => {
   const t = useTranslation('common');
-  const { setActiveRows, setDisabledRows } = useTableStore();
+  const { setRows, setDisabledRows, setFocusUp, setFocusDown } =
+    useTableStore();
+  const [clickFocusedRow, setClickFocusedRow] = useState(false);
```
CarlosNZ, line 38: This seems a bit hacky, but needed a way to let each row know when it's been clicked (by keyboard) without registering keyboard shortcut inside every row instance. Better ways to do this?

packages/common/src/ui/layout/tables/context/TableContext.ts
```diff
@@ -118,14 +126,18 @@ export const createTableStore = (): UseStore<TableStore> =>
       });
     },
 
-    setActiveRows: (ids: string[]) => {
+    setRows: (ids: string[]) => {
```
CarlosNZ, line 129: Renamed this function at Josh's suggestion (the term "active" was confusing me)

packages/invoices/src/OutboundShipment/DetailView/Footer/Footer.tsx
```diff
@@ -79,7 +79,7 @@ export const FooterComponent: FC = () => {
                 label={t('button.cancel')}
                 color="secondary"
                 sx={{ fontSize: '12px' }}
-                onClick={() => navigate(-1)}
+                onClick={() => setNavAction('up-one')}
```
CarlosNZ, line 82: This is why I exposed this action in a hook outside the Breadcrumb component itself. Cos a straight "back" doesn't quite work if you open an invoice (for example) from an external URL, say. And I figured traversing the breadcrumb / url path is something we'd probably want to do quite a bit?

packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
```diff
@@ -56,6 +68,7 @@ export const DataRow = <T extends RecordWithId>({
           }}
           onClick={onRowClick}
           hover={hasOnClick}
+          selected={isFocused}
```
CarlosNZ, line 71: Using the Mui `selected` state to show focus, which is kind of confusing since we have a separated "selected" state using the checkboxes. It should really be the same as the "hover" state. Is there a simple way to "fake" the hover state so it automatically uses the Mui hover styling?

packages/common/src/ui/layout/tables/context/TableContext.ts
```diff
@@ -152,6 +164,60 @@ export const createTableStore = (): UseStore<TableStore> =>
         };
       });
     },
+    setFocusDown: () => {
```
joshxg, line 167: The difference between these two is just the adding to the current idx, I think?

Did it become really hard to read we consolidated?

packages/common/src/ui/layout/tables/context/hooks/useIsFocused.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  const equalityFn = (
+    oldState: ReturnType<typeof selector>,
+    newState: ReturnType<typeof selector>
+  ) =>
+    oldState?.isFocused === newState?.isFocused &&
+    oldState.rowId === newState.rowId;
+
+  return useTableStore(selector, equalityFn);
+};
```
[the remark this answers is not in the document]
joshxg: Yup sounds good. Just a result of gradually adding them 😁

packages/invoices/src/OutboundShipment/DetailView/Footer/Footer.tsx
joshxg: Yup, a very nice fix

packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
joshxg: I don't know of a simple way. Only to drop down into the browser apis, get the dom node and set the hover state 💩 

Yes using the selected state is a little confusing, but I do like the colour differentiation between your focus and hovering (and also selected..). 

I think it's looks great. Maybe if you want to avoid using the selected state (and use that for the actual selected state..), then you could just apply a background to the table row directly.. or, use the `rowStyle` in the table context 🤔

joshxg (approved):
Will mention it again.. love it!



I think it's great as is but I've made some comments (brain dumps). I'm not as deep into the problem as you so your judgement is far better to know whether they are good or bad ideas. It's all working well so if you want to merge then iterate or iterate before merging or just move on - up to you!

packages/common/src/hooks/useBreadcrumbs/useBreadcrumbs.tsx
CarlosNZ: Oooh -- I didn't realise you could pass a dependency array to it! How did you know that? -- I didn't see any mention on the [kbar docs](https://kbar.vercel.app/docs/concepts/actions) (which are frustratingly sparse tbh)

Will definitely update this then, that'll make things simpler. Thanks

packages/common/src/ui/layout/tables/DataTable.tsx
CarlosNZ: Nice!  I have to admit, I don't fully understand refs, so never quite sure when they're useful. Probably a good time to learn.

For now, I've made a new issue summarising your suggestions in here, will do them together at some point when I have some time to get my head around refs properly :) 

https://github.com/openmsupply/openmsupply-client/issues/1071

packages/common/src/ui/layout/tables/context/TableContext.ts
```diff
@@ -6,6 +6,8 @@ export interface RowState {
   isSelected: boolean;
   isExpanded: boolean;
   isDisabled: boolean;
+  isFocused: boolean;
+  index: number;
```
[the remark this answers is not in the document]
CarlosNZ: Yeah, that makes sense. So currently, it just lists them in the order provided by the back-end, right?

Just thinking this through, brain dump to follow...  -- when you make a change, new data is refetched after the mutation, which causes the table to get content in new order and re-render. That's when `setRows` runs, which re-renders a new list of rows.

Are you suggesting that it would still work this way, but we'd sort the new rows according their ID order in the sorted array rather than the order that they're supplied? And then only update the ID array (based on the current `data` order) when clicking an appropriate header? So if the "setRows" runs as a result of the `useEffect` on `data`  change, how do we know whether to update the the ID array or not (cos it's not clear whether the data changes because a new order was requested (by clicking column header) or because a value changed). I guess you could keep a "headerClicked" state value, but you might have a better suggestion.

Anyway, assuming that all works fine, just wondering what happens to rows that are either missing (from ID array) or for IDs in array that are no longer in the resulting rows (due to the value change causing a row to change pages). I guess you'd just put any new items at the bottom, and omit any that were missing, then update the ID array accordingly.

packages/common/src/ui/layout/tables/context/hooks/useIsFocused.ts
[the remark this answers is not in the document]
CarlosNZ: Cool, added to the same [new issue](https://github.com/openmsupply/openmsupply-client/issues/1071)

packages/invoices/src/OutboundShipment/DetailView/Footer/Footer.tsx
CarlosNZ: Even nicer thanks to that "useRegisterAcctions" dependency array! 3244927f 👍

packages/common/src/ui/layout/tables/DataTable.tsx
CarlosNZ: > I think I like the idea of having a kbar action to focus individual rows (i.e. I could type: `focus 12` and focus row 12. Maybe `click 12` to click row 12. Or maybe one or the other? Not sure if it'd be used by end users, but _I_ think it's wicked 😁

Having dynamic values in a kbar shortcut is *probably* a bit beyond what it can easily do. Not sure how you'd capture key events without telling it in advance what numbers to be watching for. Something to think about...

(Though the rows aren't usually numbered -- so you how are you supposed to know what number to type in without manually counting through them ?? 🤔 )

packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
CarlosNZ: Okay have done it like this: ad2d979b
```
            backgroundColor: isFocused ? 'background.menu' : null,
```

Have also updated the default hover background colour to use a palette value (rather than MUI default). I like having the 'focused' and 'hover' look the same, since they're basically indicating the same thing -- one with the keyboard, one with the mouse.

packages/common/src/ui/layout/tables/context/TableContext.ts
```diff
@@ -152,6 +164,60 @@ export const createTableStore = (): UseStore<TableStore> =>
...
+        // Get currently focused row, if any
+        const [currentFocusId, currentFocusObj] =
+          rows.find(([_, { isFocused }]) => isFocused === true) || [];
+
+        // Deduce what the next row is, wrapping back to bottom if top of list
+        const nextIndex =
+          ((currentFocusObj?.index ?? rows.length) - 1 + rows.length) %
+          rows.length;
+        const [nextId, _] =
+          rows.find(([_, { index }]) => index === nextIndex) || [];
```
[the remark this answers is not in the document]
CarlosNZ: Ah, okay. Sure b52df9fc

(Though I'm not seeing any typescript error either way)

packages/common/src/ui/layout/tables/context/TableContext.ts
CarlosNZ: Yeah, better to combine together with a parameter for up/down? I agree.

Done 4a27af25

packages/common/src/ui/layout/tables/context/TableContext.ts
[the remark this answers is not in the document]
joshxg: Yeah that's all sounding right. 

I did a quick minimal sandbox of how the client side sorting in the detail views is currently working: https://codesandbox.io/s/vibrant-mountain-8qspzm?file=/src/App.tsx

It's not as bad in practise when you're using a modal to edit 😁 

> I guess you could keep a "headerClicked" state value,

This would essentially be the object that stores the current sort state, which when a header is clicked, would change.

>  guess you'd just put any new items at the bottom, and omit any that were missing, then update the ID array accordingly.

Yup .. annoying!

---------

Thanks for thinking so deeply about this, it made me think about it more also. I realised that we previously were not sending updates to the server so frequently. Now that we are, there actually is really no reason to use client side sorting..! In a later iteration, I think asking the backend to support sorting on the `lines` field on requisition/stocktake/invoices would be a good idea!

packages/common/src/ui/layout/tables/DataTable.tsx
joshxg: refs are pretty rarely used but i think useful in the rare cases. Essentially a value that is 'state' but not 'react state' (causing re-renders), I guess!

packages/common/src/hooks/useBreadcrumbs/useBreadcrumbs.tsx
joshxg: Ah yea. I knew the docs weren't great so I just cheated and went straight to the code https://github.com/timc1/kbar/blob/8f75115062504876e1a7992aad5a3c0ce549f7fa/src/useRegisterActions.tsx#L12

CarlosNZ:
Thanks for your feedback, @joshxg . I'll merge this now -- have made an issue with some things to revisit later based on your remaining suggestions.

packages/common/src/ui/components/navigation/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -19,27 +13,23 @@ const Breadcrumb = styled(Link)({
...
+
+  useRegisterActions(
+    [
+      {
+        id: 'navigation:up-one-level',
+        name: '', // No name => won't show in Modal menu
+        shortcut: ['escape'],
+        keywords: 'navigate, back',
+        perform: () => {
+          console.log('parts', urlParts);
```
mark-prins, line 26: debug code?

mark-prins (commented):
just noticed a console while having a nosy through 🙂

packages/common/src/ui/components/navigation/Breadcrumbs/Breadcrumbs.tsx
CarlosNZ: Bummer :( 
Will remove in my next PR

PR #1069, opened 2022-03-16
#1067 printing stub

### LINKED ISSUE #1067: Printing stub implementation
A stub implementation of the printing functionality has been [added to develop](https://github.com/openmsupply/remote-server/pull/900). It would be good to implement something ahead of the full implementation of #806 

A request to `reports` will return a list of reports available, which can then be provided to the `printReport` request

```
query Reporting {
  reports(storeId: "$storeId") {
    ... on ReportConnector {
      nodes {
        category
        id
        name
      }
      totalCount
    }
  }
  printReport(dataId: "$dataId", reportId: "InboundShipmentReport_1", storeId: "$storeId") {
    ... on PrintReportNode {
      __typename
      fileId
    }
  }
}
```

response is 
```
"printReport": {
      "__typename": "PrintReportNode",
      "fileId": "demofile0123456789"
    },
```

and the `fileId` is used to fetch the resulting document via [api URL]/files?id=[fileId]

Thinking to add a print icon button which pops a list of reports.. perhaps like the store switcher for now?

Screen Shot 2022-03-16 at 6 26 17 PM

mark-prins:
Fixes #1067 
Wanted to get something in which would use the stub printing implementation. and this does 😁 

Should be enough to work with the full api implementation when done
I'm not entirely convinced about the popover usage.. though ok for now?

Screen Shot 2022-03-17 at 12 00 57 PM

Couple of things not implemented in the API, though they are hooked up in this PR, awaiting implementation:
- sort
- filter by category

### DIFF
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -1,8 +1,9 @@
 export * from './routes';
 
 interface EnvironmentConfig {
-  API_URL: string;
+  API_HOST: string;
   FILE_URL: string;
+  GRAPHQL_URL: string;
 }
 
 declare global {
@@ -11,9 +12,10 @@ declare global {
   }
 }
 
-const { API_URL = 'http://localhost:4000' } = window.env;
+const { API_HOST = 'http://localhost:4000' } = window.env;
 
 export const Environment: EnvironmentConfig = {
-  API_URL,
-  FILE_URL: `${API_URL.replace('graphql', 'files')}?id=`,
+  API_HOST,
+  FILE_URL: `${API_HOST}/files?id=`,
+  GRAPHQL_URL: `${API_HOST}/graphql`,
 };
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -2,6 +2,7 @@ export * from './routes';
 
 interface EnvironmentConfig {
   API_URL: string;
+  FILE_URL: string;
 }
 
 declare global {
@@ -10,6 +11,9 @@ declare global {
   }
 }
 
-export const Environment: EnvironmentConfig = window.env ?? {
-  API_URL: 'http://localhost:4000',
+const { API_URL = 'http://localhost:4000' } = window.env;
+
+export const Environment: EnvironmentConfig = {
+  API_URL,
+  FILE_URL: `${API_URL.replace('graphql', 'files')}?id=`,
 };
--- a/packages/inventory/src/Stocktake/DetailView/AppBarButtons.tsx
+++ b/packages/inventory/src/Stocktake/DetailView/AppBarButtons.tsx
@@ -6,8 +6,16 @@ import {
   Grid,
   useDetailPanel,
   useTranslation,
+  ReportCategory,
+  LoadingButton,
+  PrinterIcon,
 } from '@openmsupply-client/common';
-import { useIsStocktakeDisabled } from '../api';
+import { useIsStocktakeDisabled, useStocktake } from '../api';
+import {
+  ReportRowFragment,
+  ReportSelector,
+  usePrintReport,
+} from '@openmsupply-client/system';
 
 interface AppBarButtonProps {
   onAddItem: (newState: boolean) => void;
@@ -19,6 +27,13 @@ export const AppBarButtonsComponent: FC<AppBarButtonProps> = ({
   const isDisabled = useIsStocktakeDisabled();
   const { OpenButton } = useDetailPanel();
   const t = useTranslation('common');
+  const { data } = useStocktake();
+  const { print, isPrinting } = usePrintReport();
+
+  const printReport = (report: ReportRowFragment) => {
+    if (!data) return;
+    print({ reportId: report.id, dataId: data?.id || '' });
+  };
 
   return (
     <AppBarButtonsPortal>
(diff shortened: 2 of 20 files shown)


packages/config/src/index.ts
```diff
@@ -10,6 +11,9 @@ declare global {
   }
 }
 
-export const Environment: EnvironmentConfig = window.env ?? {
-  API_URL: 'http://localhost:4000',
+const { API_URL = 'http://localhost:4000' } = window.env;
+
+export const Environment: EnvironmentConfig = {
```
mark-prins, line 16: 🏃 

while not ideal.. it does have all of the awfulness in one place 😬

packages/inventory/src/Stocktake/DetailView/AppBarButtons.tsx
```diff
@@ -19,6 +27,13 @@ export const AppBarButtonsComponent: FC<AppBarButtonProps> = ({
   const isDisabled = useIsStocktakeDisabled();
   const { OpenButton } = useDetailPanel();
   const t = useTranslation('common');
+  const { data } = useStocktake();
+  const { print, isPrinting } = usePrintReport();
```
mark-prins, line 31: this is really `downloadReport` aka `getAPrintableVersionOfThisThingy`
it's not ideal for a user.. but ok?
I experimented with `window.print` - and that's fine for an HTML page, but printing a PDF in chrome made the browser crash 😢

packages/system/src/Report/api/hooks.ts
```diff
@@ -0,0 +1,72 @@
...
+  const { mutate, isLoading } = useMutation<
+    string,
+    unknown,
+    PrintReportParams,
+    unknown
+  >(params => api.get.print(params), {
+    onSuccess: fileId => {
+      if (!fileId) throw new Error('Error printing report');
+      const url = `${Environment.FILE_URL}${fileId}`;
+      const win = window.open(url, '_blank');
```
mark-prins, line 63: was using `useOpenInNewTab` but decided just to run with this, since `navigate` won't help with external URLs 🤷

packages/config/src/index.ts
```diff
@@ -10,6 +11,9 @@ declare global {
   }
 }
 
-export const Environment: EnvironmentConfig = window.env ?? {
-  API_URL: 'http://localhost:4000',
+const { API_URL = 'http://localhost:4000' } = window.env;
```
joshxg, line 14: The default here won't work with the replace down there will it?

I wonder if this API_URL is better though?

```
interface EnvironmentConfig {
  API_URL: string;
  GRAPHQL_URL: string;
  FILE_URL: string;
}
...
const { API_URL = 'http://localhost:4000' } = window.env;

export const Environment: EnvironmentConfig = {
  API_URL,
   GRAPHQL_URL: `${Environment.API_URL}/graphql}`
  FILE_URL: `${Environment.API_URL/files?id=}`,
};
```

packages/system/src/Report/api/hooks.ts
```diff
@@ -0,0 +1,72 @@
...
+      if (win) {
+        win.focus();
+        // win.print(); // crashes chrome if the file is a PDF :shrug:
```
joshxg, line 66: I'm only getting .txt files - is that right?

packages/inventory/src/Stocktake/DetailView/AppBarButtons.tsx
joshxg: Yeah I tried to communicate that this endpoint using the `report` terminology is not correct and is confusing (my suggestion for gql I think was `documentToFile`

So.. as is software developer tradition, my naming is better than everyones! `useDocumentAsFile`... 

🤷 

`usePrintReport` lgtm

packages/invoices/src/InboundShipment/DetailView/AppBarButtons.tsx
```diff
@@ -29,6 +44,19 @@ export const AppBarButtonsComponent: FC<AppBarButtonProps> = ({
...
+          onClick={printReport}
+        >
+          <LoadingButton
+            variant="outlined"
+            startIcon={<PrinterIcon />}
+            isLoading={isPrinting}
+          >
+            {t('button.print')}
+          </LoadingButton>
+        </ReportSelector>
```
joshxg, line 58: Nice!

joshxg (approved):
Nice 😁 

Very cool to have that in.

Re the popover: I think there's only a couple of use cases where a document has multiple ways to print, isn't there? Pick slips and the invoice itself? It looks good but yes, I'm not sure about it 😁

packages/system/src/Report/api/hooks.ts
mark-prins: It's not right. but it is what the API is doing. 
did that make sense? I mean - yes.. the api is only serving txt files. The request was to have the api serve PDFs and I believe that the txt is just the stub and will be changed soon

packages/config/src/index.ts
mark-prins: ah. I'm going to miss your little bursts of enlightenment. that's much betterer.
`API_ROOT` maybe?

packages/inventory/src/Stocktake/DetailView/AppBarButtons.tsx
mark-prins: it is better described as `useDocumentAsFile`, and you're right, I've picked up on the api terminology and used that without thinking it through. I see this all as temporary ( or hope that it is anyway.. )

PR #1076, opened 2022-03-17
#1063 The outbound API Hook shuffle

### LINKED ISSUE #1063: Organise package api/hooks better
The api/hooks files in each of the packages are generally pretty packed and a bit difficult to grok. 

Would be good to make it simpler to read and find what you're looking for. Also would be great to not have to import so many individual hooks 😁 

Some options:

### Individual files for each hook

- Can likely pair this with a test file in the future which would be nice ✅ 
- A lot of files ❌
- can search file names easily with an ide ✅ 

### Partition by functionality
- `mutations` v `queries` / `query` v `update` v `insert` 
-  not as many files ✅ 
-  hooks that do more than one thing confusion ❌
- Can search by functionality easily ✅ 

### Partition by entity
- `document` v `lines` i.e invoice v invoice lines
-  not as many files ✅ 
-  hooks that do more than one thing confusion ❌
- Can search by 'feature' easily ✅ 

### Combo
- Individual files ✅ 
- Name space object to consolidate - end up with an api like `useOutbound.document.insert()`, `useOutbound.line.update()` - i like name spacing obvs ✅ 
- Could structure the files in `hooks/lines` to get search by feature ✅ 

### Do nothing
- It's fiiiine

I like combo, what do you like? @mark-prins @CarlosNZ

joshxg:
fixes #1063 

- uhh.. this is what i ended up with.
- I like it, but i can see how others might not
- it also maybe makes it feel a little like you're reaching into an objects properties.. which you are.. but you're also not, so makes everything maybe a little more magical? 
- An alternative is to just leave the hooks in individual files and call it a day


packages/invoices/src/OutboundShipment/DetailView/SidePanel/SidePanel.tsx
```diff
@@ -14,7 +14,7 @@ import { RelatedDocumentsSection } from './RelatedDocumentsSection';
 export const SidePanelComponent = () => {
   const { success } = useNotification();
   const t = useTranslation('distribution');
-  const { data } = useOutbound();
+  const { data } = useOutbound.document.get();
```
mark-prins, line 17: that's very neat - I like how clear that is! the `.get()` to return data makes sense

packages/invoices/src/OutboundShipment/DetailView/SidePanel/RelatedDocumentsSection.tsx
```diff
@@ -12,12 +12,12 @@ import {
...
 } from '@openmsupply-client/common';
 import { AppRoute } from '@openmsupply-client/config';
-import { useOutboundFields } from '../../api';
+import { useOutbound } from '../../api';
 
 const RelatedDocumentsSectionComponent = () => {
   const t = useTranslation('distribution');
   const d = useFormatDate();
-  const { requisition } = useOutboundFields('requisition');
+  const { requisition } = useOutbound.document.fields('requisition');
```
mark-prins, line 20: provides a nice distinction between `document` and `lines` too

mark-prins (approved):
🎉  Wow! yeah, I like it.

when using the hooks, it is really clear, and I like the 'document' / 'lines' separation etc.
the only part that feels a little strange is the namespaced hook.. calling `use...` with the drill down into lower level objects; just because I haven't seen it before. Give it a week or so and it will feel very normal I expect.

nice to have separate files, since I use cmd+p all the time to find files and the grouping of them is very handy to use

joshxg:
> the only part that feels a little strange is the namespaced hook.. calling use... with the drill down into lower level objects;

Yup me too.

I wondered about `Outbound.document.useUpdate` and maybe that's better. With hooks needing to be specifically called in a component only and the only real indication outside of the abstraction being the naming of `useX`.. maybe it's a better naming..

mark-prins:
> wondered about Outbound.document.useUpdate and maybe that's better.

FWIW I prefer `useOutbound.document.update`, maybe because it keeps the weird in one place and the rest of the naming is nice an consistent

PR #1078, opened 2022-03-17
Add supply requested quantity feature

### LINKED ISSUE #1055: Set response supplied to requested
from https://github.com/openmsupply/openmsupply-client/issues/100

> Set automatically set the quantity to be supplied for every line to be equal to the requested quantity.

- Can be an action in the drop down bulk actions
- Just use the batch update mutation. A specific mutation would be better in the future

joshxg:
Fixes #1055 

image

- Pretty much does what it says on the tin.. a button which sets all the supply quantities to the requested quantities 
- The wording here was difficult for me, not sure if it's good


packages/requisitions/src/ResponseRequisition/DetailView/AppBarButtons/SupplyRequestedQuantityButton.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const t = useTranslation('distribution');
+  const { mutate: supplyRequestedQuantity } = useSupplyRequestedQuantity();
+  const getConfirmation = useConfirmationModal({
+    onConfirm: supplyRequestedQuantity,
+    message: t('messages.supply-to-requested'),
+    title: t('heading.supply-to-requested'),
+  });
+
+  return (
+    <ButtonWithIcon
```
mark-prins, line 21: I wonder if there's some way to disable the button if all lines have already been set to requested?
On my test requisition I've got a lot of lines with requested=0, which might not be possible? but does make things a bit strage.

We could just run though and check to see if all lines have requested=quantity to supply - there's nothing provided by the api is there?

mark-prins (approved):
👍  a thing of beauty!
worked nicely 😁

packages/requisitions/src/ResponseRequisition/DetailView/AppBarButtons/SupplyRequestedQuantityButton.tsx
joshxg: Nah, nothing by the API. Also Requested equal to zero I think is possible?

 I did wonder also if it was a good idea and this is the lazy way. 

I figured there was three options:
- Alert saying "Everything is already at the requested quantity"  and have the user push "OK"
- No alert, just send the request (my lazy way)
- No alert, no request.. just pretend it happened..

🤷 😁

mark-prins: let's leave it - asking the user to press OK again to clear an alert is also annoying for some people.
will wait and see what other people think

PR #1081, opened 2022-03-17
#1079 Requested to suggested

### LINKED ISSUE #1079: requested to suggested
Use the `useSuggestedQuantity` mutation in internal orders

joshxg:
Fixes #1079 

- Pretty much does what it says on the tin.. a button which sets all the requested quantities to the suggested quantities 
- The wording here was difficult for me, not sure if it's good


packages/common/src/intl/locales/en/replenishment.json
```diff
@@ -2,11 +2,13 @@
   "button.delete-lines": "Delete lines",
   "button.new-shipment": "New Shipment",
   "button.save-and-confirm-status": "Confirm {{status}}",
+  "button.requested-to-suggested": "Requested to Suggested",
```
mark-prins, line 5: oh no! this could go either way!
`Suggested to Requested` could mean the same thing.. as in, take the suggested and put it in requested.

maybe something like `Use Suggested`? ooh, what about `Request Suggested`?

packages/requisitions/src/RequestRequisition/DetailView/AppBarButtons/AppBarButtons.tsx
```diff
@@ -43,20 +41,9 @@ export const AppBarButtonsComponent: FC<AppBarButtonProps> = ({
...
-        />
-        <MasterListSearchModal
-          open={modalController.isOn}
-          onClose={modalController.toggleOff}
-          onChange={masterList => {
-            modalController.toggleOff();
-            addFromMasterList(masterList);
-          }}
-        />
+        <UseSuggestedQuantityButton />
```
mark-prins, line 44: could you change the order please?
I think: Add item should be first, since that's most often used, and should be easiest to find. having it in the middle makes it hard to see..
So the button order:
1. Add item
2. Add from master list
3. Request Suggested (?)
4. Print
5. More

packages/requisitions/src/RequestRequisition/DetailView/AppBarButtons/UseSuggestedQuantityButton.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const { mutate: setRequestedToSuggested } = useSuggestedQuantity();
+  const getConfirmation = useConfirmationModal({
+    onConfirm: setRequestedToSuggested,
+    message: t('messages.requested-to-suggested'),
+    title: t('heading.requested-to-suggested'),
+  });
+
+  return (
+    <ButtonWithIcon
+      Icon={<EditIcon />}
```
mark-prins, line 21: at one point Manu has a lightning bolt icon for the 'wizard' actions.. that might be nice here. 
leave it for now, I can add that in later and see how it looks

mark-prins (approved):
👍  awesome!

I have a concern about suggested quantity which I hadn't noticed before.. this is now calculated for the table, but on the edit modal it is taking the value from the line, which might not match. Might make an issue for that.

packages/requisitions/src/RequestRequisition/DetailView/AppBarButtons/AppBarButtons.tsx
joshxg: Ah yup!

packages/common/src/intl/locales/en/replenishment.json
joshxg: `Request Suggested` it is (because it was the last one 😁 )

packages/common/src/intl/locales/en/replenishment.json
```diff
@@ -2,11 +2,13 @@
   "button.delete-lines": "Delete lines",
```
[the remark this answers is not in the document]
joshxg: whoops!

joshxg:
> I have a concern about suggested quantity which I hadn't noticed before.. this is now calculated for the table, but on the edit modal it is taking the value from the line, which might not match. Might make an issue for that.

Should also allow you to remove the `suggestedQuantity` function all together which is good!

packages/requisitions/src/RequestRequisition/DetailView/AppBarButtons/UseSuggestedQuantityButton.tsx
joshxg: Oh! That's right. I did wonder if I should use the popover similarly to how you've done the `Print` with these extra actions. The wizard modal was cool though.

For this one (and the response one), I'd imagine the API should change to accept lines and for this to become an action in the drop down.

Maybe another button like mobile: "One click order", which populates the requisition with all of the items which have a suggested quantity > 0 and sets the requested to the suggested

PR #1083, opened 2022-03-18
update alert

### LINKED ISSUE #1075: Bug: have to login twice?
New problem I've noticed since the latest changes to the refresh token. After being logged out (or manually logging out), then next time I log in, I immediately get the "Session timed out" message, and returned to the Login screen. 

Subsequent login works fine.

I guess there might be an old version of the cookie that is somehow still being sent after that first login?

Screen Shot 2022-03-18 at 8 55 50 AM

mark-prins:
Fixes #1075 

Can now log out and in again without being prompted. 
In a dev environment, where the refresh token isn't working, you'll still have an issue after being logged out and will need to refresh the page. after that it's fine.

### DIFF
--- a/packages/common/src/authentication/AuthContext.tsx
+++ b/packages/common/src/authentication/AuthContext.tsx
@@ -76,12 +76,11 @@ const useRefreshingAuth = (
 ) => {
   const { setHeader } = useGql();
   setHeader('Authorization', `Bearer ${token}`);
-  const { data, isSuccess } = useGetRefreshToken(token ?? '');
+  const { data, enabled, isSuccess } = useGetRefreshToken(token ?? '');
   useEffect(() => {
-    if (isSuccess) callback(data?.token);
-  }, [data, isSuccess, callback]);
+    if (isSuccess && enabled) callback(data?.token ?? '');
+  }, [enabled, isSuccess, data]);
 };
-
 const AuthContext = createContext<AuthControl>({
   isLoggingIn: false,
   login: () =>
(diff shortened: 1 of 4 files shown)


packages/common/src/authentication/AuthContext.tsx
```diff
@@ -76,12 +76,11 @@ const useRefreshingAuth = (
 ) => {
   const { setHeader } = useGql();
   setHeader('Authorization', `Bearer ${token}`);
-  const { data, isSuccess } = useGetRefreshToken(token ?? '');
+  const { data, enabled, isSuccess } = useGetRefreshToken(token ?? '');
```
mark-prins, line 79: this isn't great. The callback is setting state on the component, causing a re-render which will fire the `useEffect` again. fine, except that the `useGetRefreshToken` is only returning data after a timeout ( to prevent issues when developing.. without that the refresh will wipe out your token and log you out ). until the refresh returns, the `data` is undefined and isn't useful as a referential dependency.

all of which is to say.. this is working but I'm not that happy with it. 🤷

packages/common/src/authentication/AuthContext.tsx
```diff
@@ -101,22 +100,23 @@ export const AuthProvider: FC = ({ children }) => {
   const i18n = useI18N();
   const defaultLanguage = useDefaultLanguage();
   const { mutateAsync, isLoading: isLoggingIn } = useGetAuthToken();
-  const { token: cookieToken, store: cookieStore, user } = getAuthCookie();
-  const [localStore, setLocalStore] = useState<Store | undefined>(cookieStore);
-  const [localToken, setLocalToken] = useState<string | undefined>(cookieToken);
-  const storeId = localStore?.id ?? '';
+  const authCookie = getAuthCookie();
+  const [cookie, setCookie] = useState<AuthCookie | undefined>(authCookie);
```
mark-prins, line 104: changed to cookie - the `user` was outside of local state and so was a bit weird. 
was using the `user` to trigger the logged out modal, which meant after login the newly assigned user wasn't being passed to the logged out modal, and you get prompted again.

packages/host/src/components/Navigation/RequireAuthentication.tsx
```diff
@@ -1,20 +1,24 @@
...
   const { token } = useAuthContext();
   const location = useLocation();
+  const navigate = useNavigate();
 
-  if (!token) {
-    return (
-      <Navigate to={`/${AppRoute.Login}`} state={{ from: location }} replace />
-    );
-  }
+  useEffect(() => {
```
mark-prins, line 14: changed to an effect, with no deps, so that it only gets called on load. this way, the logged out alert can be shown - otherwise, this will redirect and the alert never gets displayed 🙄

CarlosNZ (approved):
Cool, looks good, seems to be working nicely. I think I get your concerns about it being less than an ideal solution. Perhaps when we have some time (and I understand how it's working a bit better) we can come up with a more straightforward solution that doesn't require that timeout hack.

mark-prins:
thanks - yeah, would be good to review the process. I still think it can be better!

PR #1085, opened 2022-03-19
#578 Filter supplier requisition lines

### LINKED ISSUE #578: Supplier requisition lines: filter by item name
Add a filter input for supplier requisition lines: https://docs.google.com/spreadsheets/d/1H0Rv1DEOgW0cE07wpfc_KivLf8TcNhYDqlwRp92x-To/edit#gid=0&range=A234

CarlosNZ:
Fix #578

Also filters internal order lines the same way.

Since we're just filtering an already queried list of items, I've just used a simple local `useState` value for the filter string rather than the more comprehensive `useFilterBy` hook (which re-queries the database).

See inline comments for more details.


packages/common/src/ui/components/inputs/SearchBar/SearchBar.tsx
```diff
@@ -38,7 +40,7 @@ export const SearchBar: FC<SearchBarProps> = ({
       setLoading(false);
     },
     [onChange],
-    500
+    debounceTime
```
CarlosNZ, line 43: Made the debounce time configurable (though leaving existing value as default). Seemed a bit weird seeing a loading spinner for a list filter that can be instant since we already have all the data -- have set debounceTime to `0` in that case.

packages/common/src/ui/components/navigation/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -22,10 +22,7 @@ export const Breadcrumbs: React.FC = () => {
         name: '', // No name => won't show in Modal menu
         shortcut: ['escape'],
         keywords: 'navigate, back',
-        perform: () => {
-          console.log('parts', urlParts);
-          navigateUpOne();
-        },
+        perform: () => navigateUpOne(),
```
CarlosNZ, line 25: Leftover from previous PR 🙄

packages/common/src/utils/regex/RegexUtils.tsx
```diff
@@ -23,4 +23,9 @@ export const RegexUtils = {
       />
     );
   },
+  // Case-insensitive match of partial string -- same as SQL "LIKE"
+  matchSubstring: (substring: string, testString: string) => {
+    const matcher = new RegExp(substring, 'i');
+    return matcher.test(testString);
+  },
```
CarlosNZ, line 30: I couldn't find a simple partial string matcher anywhere already -- let me know if I've missed an existing one.

packages/requisitions/src/RequestRequisition/DetailView/Toolbar/Toolbar.tsx
```diff
@@ -1,27 +1,43 @@
...
   ]);
 
   return (
-    <AppBarContentPortal sx={{ display: 'flex', flex: 1, marginBottom: 1 }}>
+    <AppBarContentPortal
+      sx={{
+        display: 'flex',
+        flex: 1,
+        marginBottom: 1,
+        flexDirection: 'column',
```
CarlosNZ, line 38: Made outermost flex box `column` and placed the Search and Dropbdown components in their own `<Grid>` outside the one the other components are in -- otherwise Search box was pulling the `<ToolbarActions>` stuff over to the right and it would move as the SearchBar expanded -- not pretty.

packages/common/src/utils/regex/RegexUtils.tsx
mark-prins: no, there isn't one! because the filtering is all server-side we haven't had much need for client side utils.
You could add a test - but I think not worth it yet.

packages/requisitions/src/ResponseRequisition/DetailView/ContentArea.tsx
```diff
@@ -1,20 +1,27 @@
 import React from 'react';
 import { DataTable, useTranslation } from '@openmsupply-client/common';
 import { useResponseLines, ResponseLineFragment } from '../api';
+import { RegexUtils } from '@openmsupply-client/common';
```
mark-prins, line 4: :nit nicer to have from `import { RegexUtils } from '@common/utils';`

packages/requisitions/src/RequestRequisition/DetailView/ContentArea.tsx
```diff
@@ -1,19 +1,27 @@
 import React from 'react';
 import { DataTable, useTranslation } from '@openmsupply-client/common';
 import { RequestLineFragment, useRequestLines } from '../api';
+import { RegexUtils } from '@openmsupply-client/common';
```
mark-prins, line 4: :nit `import { RegexUtils } from '@common/utils';`

packages/requisitions/src/RequestRequisition/DetailView/Toolbar/Toolbar.tsx
mark-prins: Originally I didn't like the RH layout of the item filter, I looked at 'Internal Orders' first and thought on that screen it would be better on the left..
...but then saw response reqs, and it is better on the right there. And consistency is important! and on internal orders, I like having all the fields related to the requisition together and all the 'view' elements ( filters and such) together.

so.. 👍  and 😁

mark-prins (approved):
👍  that's great thanks!

I do have a request though.. 

We have been using hooks in preference to passing props around. The approach here is fine, but a departure from the approach taken elsewhere.

Ideally.. extend `useRequestLines` to store the item filter internally and return `itemFilter` and `setItemFilter`. note that `useState` wouldn't work in here, as a different instance will be called by different components (🙄 ). This would work though:

```
export const useItemFilter = zustand<{
  itemFilter: string;
  setItemFilter: (itemFilter: string) => void;
}>(set => ({
  setItemFilter: (itemFilter: string) =>
    set(state => ({ ...state, itemFilter })),
  itemFilter: '',
}));
```

and then 
```
const { itemFilter, setItemFilter } = useItemFilter();
...
if (on) {
      return sorted.filter(
        item =>
          item.itemStats.availableStockOnHand <
            item.itemStats.averageMonthlyConsumption * minMonthsOfStock &&
          RegexUtils.matchSubstring(itemFilter, item.name)
      );
    } else {
      return sorted.filter(({ item: { name } }) =>
        RegexUtils.matchSubstring(itemFilter, name)
      );
    }
``` in the hook


then you can remove the props from `Toolbar` and instead use 
```const { itemFilter, setItemFilter } = useRequestLines();```
and `ContentArea` and `DetailView` just revert back to what they were.. with the ContentArea fetching lines. Removes a lot of inter-dependence between components 😄

File: packages/requisitions/src/RequestRequisition/DetailView/Toolbar/Toolbar.tsx (hunk 4 above)
Comment (CarlosNZ) [reply packages/requisitions/src/RequestRequisition/DetailView/Toolbar/Toolbar.tsx:34]: Yeah, I went through the exact same thought process 😄

File: packages/requisitions/src/ResponseRequisition/api/hooks.ts [hunk 7]
```diff
@@ -18,8 +18,10 @@ import {
   useNotification,
   useTranslation,
   Column,
+  RegexUtils,
 } from '@openmsupply-client/common';
 import { getResponseQueries, ListParams } from './api';
+import { useItemFilter } from '../../RequestRequisition/api';
```
CarlosNZ, line 24: Not sure if this is okay. The Request and Response requisition modules seem quite well-separated, so is it bad to share a common filter string?

packages/requisitions/src/ResponseRequisition/DetailView/ContentArea.tsx
CarlosNZ: These are now in the hooks, and I've just included the import with a large bunch of existing imports from `@openmsupply-client/common`. That okay?

CarlosNZ:
> I do have a request though..
> 
> We have been using hooks in preference to passing props around. The approach here is fine, but a departure from the approach taken elsewhere.

Okay cool, have done that: cb4331fb

packages/requisitions/src/ResponseRequisition/api/hooks.ts
mark-prins: can't see an issue with it - and it's nice to be using the same hook from different components

mark-prins (approved):
😁  very nice!

PR #1086, opened 2022-03-20
Select items on stocktake creation

### LINKED ISSUE #1018: Stocktake: Select items when creating
Coming from [this comment](https://github.com/openmsupply/openmsupply-client/issues/582#issuecomment-1063524444) and that it is annoying to add one item at a time.

Could we show a list of stock items with a checkbox next to each and a 'select all' checkbox. 
On create, the stocktake is created and lines added for each selected item. Items can then be added and removed as they can now - no change to the edit / detail view

mark-prins:
Fixes #1018 

Adds the option to choose items when creating a stocktake.
One more thing I would like to have in is the ability to select all, of the currently filtered results. 
That's going to be a little more complex and is a nice to have, so I'll stop here and get this in and make a new issue for that.


packages/common/src/hooks/useDialog/useDialog.tsx
```diff
@@ -20,6 +20,7 @@ export interface ModalProps {
   nextButton?: React.ReactElement<{
     onClick: () => Promise<boolean>;
   }>;
+  slideAnimation?: boolean;
```
mark-prins, line 23: this one was neat 😉 
The slide animation make the 'OK & Next' functionality look great - but.. it does cause an issue with the Popper component used by the autocomplete, if using the `open` prop 🙄 
With the animation, the popper shows in the correct place vertically, but on the hard left of the screen! any interaction with the page fixes it.. but it is super weird. 

So, have added the option to disable the animation.

packages/common/src/ui/components/inputs/Autocomplete/AutocompleteList.tsx
```diff
@@ -17,12 +17,16 @@ export type AutocompleteListProps<T> = {
   loading?: boolean;
   loadingText?: React.ReactNode;
   noOptionsText?: React.ReactNode;
-  onChange?: AutocompleteOnChange<T>;
+  onChange?: AutocompleteOnChange<T | T[]>;
```
mark-prins, line 20: this is because I've added `multiple` as an optional prop and the return shape is now an item or an array

packages/inventory/src/Stocktake/ListView/CreateStocktakeButton.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { StockItemSelectModal } from 'packages/system/src';
+
+export const CreateStocktakeButton: React.FC = () => {
+  const t = useTranslation(['distribution', 'common']);
+  const { mutate } = useInsertStocktake();
+  const modalController = useToggle();
+
+  return (
+    <>
+      {modalController.isOn && (
```
mark-prins, line 16: doing this so that the modal is only mounted when requested - which means the item query is only sent when you want to create a stocktake too.. all to make things as snappy as possible

packages/inventory/src/Stocktake/api/api.ts
```diff
@@ -170,6 +170,17 @@ export const getStocktakeQueries = (sdk: Sdk, storeId: string) => ({
     });
     const { insertStocktake } = result;
     if (insertStocktake.__typename === 'StocktakeNode') {
+      if (itemIds) {
+        const insertStocktakeLines = itemIds.map(itemId => ({
+          id: FnUtils.generateUUID(),
+          stocktakeId: insertStocktake.id,
+          itemId,
+        }));
+        await sdk.upsertStocktakeLines({
```
mark-prins, line 179: the return shape makes error catching difficult here, so I didn't bother 😁 
The mutation always returns a BatchMutationResponse regardless of errors, and you have to drill down to get the insert lines response. Decided that it doesn't really matter anyway - the key mutation here is the stocktake insert, and if the lines insert fails that's pretty obvious.

packages/system/src/Item/Components/StockItemSelectModal/StockItemSelectModal.tsx
```diff
@@ -0,0 +1,144 @@
...
+      {option.name}
+    </span>
+  </li>
+);
+
+const ItemInput: FC<AutocompleteRenderInputParams> = props => {
+  const { InputProps, ...rest } = props;
+  const { startAdornment, ...restInputProps } = InputProps ?? {};
+  const t = useTranslation('common');
+  const length =
```
mark-prins, line 58: now that I've done it.. this could be changed to use the `itemIds` array

packages/system/src/Name/Components/CustomerSearchModal/CustomerSearchModal.tsx
```diff
@@ -19,8 +19,8 @@ export const CustomerSearchModal: FC<NameSearchProps> = ({
       onClose={onClose}
       title={t('customers')}
       optionKey="name"
-      onChange={(_, name: NameRowFragment | null) => {
-        if (name) onChange(name);
+      onChange={(_, name: NameRowFragment | NameRowFragment[] | null) => {
```
mark-prins, line 22: these three are just because I changed the shape of the prop

CarlosNZ (commented):
Thanks Mark, just one comment I made, plus --  the "search by code" doesn't seem to be working in the autocomplete list (it works on the "Add item" dialog within the Stocktake page)

packages/common/src/hooks/useDialog/useDialog.tsx
mark-prins: oops. not my intention to remove it completely. 1ea2dbdb

mark-prins:
Thanks @CarlosNZ  - well spotted. I've added search by code in 1ea2dbdb as well as fixing the 'ok & next' button.

CarlosNZ (approved):
Great, thanks Mark, that's all working well now.

Just one tiny thing, you may or may not agree. I think it might be slightly better if the search box kept its state after checking items. It's slightly disconcerting having the whole list pop back when ticking, and I don't think there's any problem with the user forgetting they have a filter on. Plus, if the user wants to, for example, select all the items containing "Benzyl", it'd be easier if they didn't have to re-enter their "benzyl" search term again after ticking the first one.

mark-prins:
thanks! I hadn't noticed the input value clearing - and I agree. 
Have added in the state to retain the text

PR #1091, opened 2022-03-21
#1074 mSupply man hover background

### LINKED ISSUE #1074: Spinning man logo has wrong BG color on hover
Just a tiny thing:

Screen Shot 2022-03-17 at 10 43 25 PM

CarlosNZ:
Fix #1074 

Was a tiny fix -- see what you think. (Seems useful to have the full `sx` props passed down from the `IconButton` anyway.)


mark-prins (approved):
🎉  awesome! 

that looks much better - thanks! I have a vague concern that devs might in the future abuse the power of having `sx` available.. and we'll have non-standard styling on things.
No matter, that's what code reviews are for.

CarlosNZ:
> I have a vague concern that devs might in the future abuse the power of having `sx` available.. and we'll have non-standard styling on things.

Yes! I generally like to keep styling out of markup (as much as practical), so we should try and limit `sx` to special-case overrides only.

PR #1094, opened 2022-03-21
Update entered-by label

### LINKED ISSUE #1093: Entered by caption
The original request for user was to display `Created by` (#827), however the user returned by the API isn't always the created-by user, and the caption won't quite match for Inbounds or Response reqs

Reference [#931](https://github.com/openmsupply/remote-server/issues/931)

`Request Requisition` -> entered by
`Outbound Shipment` -> entered by
`Stocktake` -> entered by
`Response Requisition` -> edited by
`Inbound Shipment` -> edited by

We'll need to update the label text - and possibly have the requirement changed?

mark-prins:
Fixes #1093 

fwiw I think that 'entered by' is more helpful and I'd prefer if that was returned. Have updated the label anyway so that we're not misleading.

### DIFF
--- a/packages/common/src/intl/locales/en/distribution.json
+++ b/packages/common/src/intl/locales/en/distribution.json
@@ -18,7 +18,6 @@
   "heading.comment": "Comment",
   "heading.create-outbound-shipment": "Create Outbound Shipment",
   "heading.edit-item": "Edit Item",
-  "label.entered-by": "Entered by",
   "heading.supply-to-requested": "Set supply to requested",
   "heading.tax": "Tax",
   "heading.edit-tax-rate": "Edit tax rate",
--- a/packages/requisitions/src/ResponseRequisition/DetailView/AdditionalInfoSection.tsx
+++ b/packages/requisitions/src/ResponseRequisition/DetailView/AdditionalInfoSection.tsx
@@ -1,64 +1,56 @@
-import React, { FC, memo } from 'react';
+import React, { FC } from 'react';
 import {
   Grid,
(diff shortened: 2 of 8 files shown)


packages/common/src/intl/locales/en/distribution.json
```diff
@@ -18,7 +18,6 @@
   "heading.comment": "Comment",
   "heading.create-outbound-shipment": "Create Outbound Shipment",
   "heading.edit-item": "Edit Item",
-  "label.entered-by": "Entered by",
```
mark-prins, line 21: not required - as the translation is in common

packages/requisitions/src/ResponseRequisition/DetailView/AdditionalInfoSection.tsx
```diff
@@ -0,0 +1,56 @@
+import React, { FC } from 'react';
```
mark-prins, line 1: just lifted the components out.. making it consistent with the other side panels

CarlosNZ (approved):
Cool, thanks.

One tiny slightly unrelated thing since you're poking around in language strings -- I notice there is a lone `label.sent` in `replenishment.json`  which is not being rendered correctly on the requisitions page. Probably need to chuck it in "common" with the other statuses:
Screen Shot 2022-03-23 at 4 01 33 PM

mark-prins:
🤔  the 'Sent` status isn't valid for response requisitions.. 
looks like dodgy data that I had added to the server 😬

PR #1103, opened 2022-03-23
#1095 Use gradient logo for Login and AppDrawer

### LINKED ISSUE #1095: mSupply logo with gradient
Low priority, but try restoring the colour gradient on the mSupply logo (but **not** on the bouncing loader)

CarlosNZ:
Fix #1095 

Made the `size` prop optional, so can be used with enumerated sizes *or* specific `width`/`height` values. So now `MSupplyGuyGradient` and `MSupplyGuy` are exactly the same (except for the gradient) and can be used interchangeably.


packages/common/src/ui/icons/MSupplyGuy.tsx
```diff
@@ -19,15 +19,15 @@ const sizes = {
...
+  size?: 'large' | 'medium';
 };
 
 const SvgGuy: FC<SvgIconProps & { fill: string }> = props => {
   const { fill, ...svgProps } = props;
   return (
     <SvgIcon {...svgProps} viewBox="3 1 206 310">
       <g>
-        <linearGradient id="6000022348f0" x1="0.5" y1="0" x2="0.5" y2="1">
+        <linearGradient id="linear-grad" x1="0.5" y1="0" x2="0.5" y2="1">
```
CarlosNZ, line 30: I'm assuming there was no particular reason that this `id` had such an unhelpful name?

mark-prins: yeah, laziness 😁

mark-prins (approved):
👍  Nice! 
Not sure why there isn't an issue with the loading indicator now.. but there isn't. So the gradient is back in 😁 

I have a suggestion - it was multiple files so I've committed it to cdc259c7 and you can revert / update as you see fit.

basically.. couldn't see the point of having separate `MSupplyGuy` and `MSupplyGuyGradient` now. The only reason for the two was because the loading indicator didn't work with the gradient.

However.. the change here does mean that the loading indicator has a hover animation effect, which is not really what we want:



so.. I recommend changing the names: applying the gradient to both, and having an animated version and a non-animated. So that's what I've done 😁

CarlosNZ:
> So that's what I've done 😁

Okay, great. Just needed to also pass the existing style prop through as well when no named size is used, so custom sizes will continue to work (as used on the Login screen)
```
const sizeStyle = size ? sizes[size] : { ...svgProps.style };
```

8a73c648

mark-prins:
argh! thanks 😁  I had a mental note to do that.. and lost it..

PR #1104, opened 2022-03-23
Lerna speedup -- is this useful?

CarlosNZ:
Small tweak to only re-compile packages that have had changes since the last commit. Doesn't actually speed up the process very much as most packages pull from other ones so any change recompiles any that are connect.

Not sure if actually useful, what do we think?


package.json
```diff
@@ -6,7 +6,7 @@
     "build": "lerna run --scope @openmsupply-client/* build",
     "serve": "lerna run --scope @openmsupply-client/* --parallel serve",
     "clean": "lerna run --scope @openmsupply-client/* --parallel clean",
-    "compile": "lerna run --scope @openmsupply-client/* --parallel tsc",
+    "compile": "lerna run --scope @openmsupply-client/* --parallel tsc --since HEAD~1",
```
mark-prins, line 9:
```suggestion
    "compile": "lerna run --scope @openmsupply-client/* --parallel tsc --since HEAD",
```

I believe it should be since `HEAD` rather than the first ancestor.
To check - I made a change in `Host.tsx` and running with HEAD only, then only the host package is compiled.
With the ancestor level then it compiles four packages (dashboard, requisitions, host, system). These are compiled even if I revert my changes.

mark-prins (approved):
🎉   Yes - definitely worth adding in! thanks for finding that.

I think the ref needs changing - and if you do, then the compile is super fast 😁

PR #1109, opened 2022-03-24
Indicate internal stores

### LINKED ISSUE #1108: Indicate store for customers
The SupplierSearchModal has an indication to show that the supplier is internal i.e. an mSupply store.

Screen Shot 2022-03-24 at 12 17 55 PM

The CustomerSearchModal wants to have this too!
And InternalSupplierSearchModal I suppose 🙄

mark-prins:
Fixes #1108 

Mainly just lifting out the option render method, so that all three search modals do the same thing.
Then went a bit further and fixed console errors and allowed searching by name or code 🤷


packages/common/src/utils/regex/RegexUtils.tsx
```diff
@@ -28,4 +28,19 @@ export const RegexUtils = {
     const matcher = new RegExp(substring, 'i');
     return matcher.test(testString);
   },
+  // Case-insensitive version of 'startsWith'
+  matchString: (substring: string, testString: string) => {
+    const matcher = new RegExp(`^${substring}`, 'i');
+    return matcher.test(testString);
+  },
+  // matches test string against an object's name or code properties
+  matchNameOrCode: (
```
mark-prins, line 37: mSupply generally only searches using startsWith, so I've kept that tradition.. although I find the `includes` more helpful myself

packages/system/src/Name/Components/CustomerSearchModal/CustomerSearchModal.tsx
```diff
@@ -18,7 +23,13 @@ export const CustomerSearchModal: FC<NameSearchProps> = ({
       options={data?.nodes ?? []}
       onClose={onClose}
       title={t('customers')}
-      optionKey="name"
+      renderOption={NameOptionRenderer}
+      getOptionLabel={(option: NameRowFragment) => option.name}
```
mark-prins, line 27: adding the option label, as Autocomplete was unhappy with `[object, object]` as the label 🙄

packages/common/src/utils/regex/RegexUtils.tsx
```diff
@@ -28,4 +28,19 @@ export const RegexUtils = {
     const matcher = new RegExp(substring, 'i');
     return matcher.test(testString);
   },
+  // Case-insensitive version of 'startsWith'
+  matchString: (substring: string, testString: string) => {
```
CarlosNZ, line 32: Just wondering if we want to make these two functions (matchString and matchSubstring) have slightly more helpful names. Maybe `matchStringIn` and `matchStringStart`? Just something so it's clear from the function name which is which.

packages/common/src/utils/regex/RegexUtils.tsx
```diff
@@ -28,4 +28,19 @@ export const RegexUtils = {
...
+    return matcher.test(testString);
+  },
+  // matches test string against an object's name or code properties
+  matchNameOrCode: (
+    item: { name: string; code: string },
+    searchString: string
+  ) => {
+    return (
+      RegexUtils.matchString(searchString, item.name) ||
+      RegexUtils.matchString(searchString, item.code)
```
CarlosNZ, line 43: This is good, I've never been able to figure out how to refer to a value from another property of the same object! Thanks :)

packages/common/src/utils/regex/RegexUtils.tsx
```diff
@@ -28,4 +28,19 @@ export const RegexUtils = {
...
   },
+  // Case-insensitive version of 'startsWith'
+  matchString: (substring: string, testString: string) => {
+    const matcher = new RegExp(`^${substring}`, 'i');
+    return matcher.test(testString);
+  },
+  // matches test string against an object's name or code properties
+  matchNameOrCode: (
+    item: { name: string; code: string },
+    searchString: string
```
CarlosNZ, line 39: Could take this a step further and allow matching against *any* object?

Function would check for matches against any of the properties of the passed-in object, and when you call it you just pass an object with only the fields you want matched.

CarlosNZ (commented):
Cool -- couple of little suggestions, see what you think.

packages/common/src/utils/regex/RegexUtils.tsx
CarlosNZ: Also -- once you've made any changes you want to do here, I'll write some tests for the new functions.

CarlosNZ: > Function would check for matches against any of the properties of the passed-in object, and when you call it you just pass an object with only the fields you want matched.

This works, I think ;) 
```ts
matchObjectStrings: (object: { [key: string]: string }, searchString: string) =>
    Object.keys(object).some(key =>
      RegexUtils.matchSubstring(object[key] as string, searchString)
    ),
```

mark-prins: oh that's good! I see what you mean now 😁 

Was just doing it and realised that we shouldn't.. filtering all string props would include `id` and therefore return unexpected results. potentially other props could be added if the fragment is updated. Safer to be explicit about the props to filter.

CarlosNZ: Oh, I was thinking that when you call the function, you just pass in an subset of the object containing the fields you want to check.

E.g. you have an item `item` with `{id: 123, name: "Drug A", price: 2.34, code: "1234", description: "Some drug"}`, then you'd just call it like: `matchObjectStrings({code: item.code, name: item.name}, <searchString>)` to match against just the "name" and "code" fields.

I feel like eventually we'll likely have a lot of different filters matching different things, so it'd be good to handle all possibilities without writing a new function for each. Can always wait to generalise until we need it though.

mark-prins: ah, right. I thought it was pretty neat just checking all string props.

I had this
```
  // returns true if the search string is contained in any of the properties specified by the given keys
  matchObjectProperties: function <T>(
    substring: string,
    object: T,
    keys: Array<keyof T>
  ) {
    keys.some(key => RegexUtils.includes(substring, String(object[key])));
  },
```

so that I could
```
  options.filter(option =>
    RegexUtils.matchObjectProperties(state.inputValue, option, ['name', 'code'])
  );
```

which does the same thing.. I think I understand now 😄

CarlosNZ: Yup, an array of field names would work well. Can it be optional though, so it'll match *all* keys if not specified. (parameter default would be `Object.keys(object)` -- is that possible, to use another parameter to generate the default value for another? 🤷‍♂️)

CarlosNZ (approved):
Great, thanks for the updates. And tests too! 👍🏼

PR #1110, opened 2022-03-24
Stocktake create: Select all option

### LINKED ISSUE #1089: Stocktakes: select all items
Couple of issues off the back of #1018 

- would like to be able to select all items which are shown in the autocomplete filter
- when deleting a stocktake, the notistack notification is shown twice

mark-prins:
Fixes #1089 
Adds a 'select all' option to the item selection modal.
Also updates the filter so that either name or code is matched, using `startsWith` rather than `includes`.

Some issues with ts typing, as you'll see and overall this makes the component much more controlled than I would like. If we require something like this modal elsewhere, we'll need to lift it out, and maybe refine a little.


packages/inventory/src/Stocktake/api/hooks.ts
```diff
@@ -323,7 +323,6 @@ export const useDeleteSelectedStocktakes = () => {
           number: numberSelected,
         });
         const successSnack = success(deletedMessage);
-        successSnack();
```
CarlosNZ, line 326: Ah -- this is why it was showing the notification twice! Nice.

packages/system/src/Item/Components/StockItemSelectModal/StockItemSelectModal.tsx
```diff
@@ -70,24 +73,44 @@ export const StockItemSelectModal = ({
...
+            flex={1}
+            style={{ verticalAlign: 'bottom' }}
+            display="flex"
+            alignItems="center"
+          >
+            {t('label.items-selected', { count: selectedItems.length })}
+          </Typography>
+          <Typography textAlign="right" flex={1}>
+            {t('label.select-all')}
+            <Checkbox onChange={selectAll} />
```
CarlosNZ, line 111: Would be nice if the checkbox value showed the "partial" state in response to selecting/un-selecting individual items, like the Tables do:
Screen Shot 2022-03-25 at 2 22 54 PM

Although... I'm not sure what the logic would be when there is a filter applied -- should the "Select all" checkbox state reflect *all* the items or just the currently filtered ones? I want to say *all* the items, but then when you *click* "Select All" it only updates the *filtered* items. And that's kind of confusing so maybe we should just leave it as it is now. Thoughts?

packages/system/src/Item/Components/StockItemSelectModal/StockItemSelectModal.tsx
```diff
@@ -113,9 +136,11 @@ export const StockItemSelectModal = ({
           disabled={saving}
           variant="ok"
           onClick={async () => {
-            setSaving(true);
-            await onChange(itemIds);
-            setSaving(false);
+            if (selectedItems instanceof Array) {
```
CarlosNZ, line 139: Why this check? When can `selectedItems` ever NOT be an Array?

packages/system/src/Item/Components/StockItemSelectModal/StockItemSelectModal.tsx
```diff
@@ -126,17 +151,23 @@ export const StockItemSelectModal = ({
...
+            height={365}
             width={600}
             disableCloseOnSelect
             multiple
             getOptionLabel={option => `${option.code} ${option.name}`}
+            filterOptions={(options, state) =>
+              options.filter(option =>
+                RegexUtils.matchNameOrCode(option, state.inputValue)
+              )
+            }
```
CarlosNZ, line 163: Will need to update function name as a result of your other PR changes.

Just out of interest -- when you merge a branch to `main` do you have any pre-merge checks? Cos this branch is currently valid, and `main` is currently valid, but once you merge this into `main`, there will be compile errors due to the function name change from the other branch.

CarlosNZ (commented):
Nice :)

You probably want to merge `main` into here and fix the resulting errors (due to function name change) before we merge this.

packages/system/src/Item/Components/StockItemSelectModal/StockItemSelectModal.tsx
mark-prins: no - just the pre-commit check. It's theoretically possible that the merge makes a hash of things - but hasn't been an issue yet.

packages/system/src/Item/Components/StockItemSelectModal/StockItemSelectModal.tsx
mark-prins: I'll put that one down to a previous version. It's deffo an array now.

packages/system/src/Item/Components/StockItemSelectModal/StockItemSelectModal.tsx
mark-prins: The intention was to make it possible to select a filtered list of items, similar to what you can do now in mSupply.
For that - the 'select all' action has to operate on the filtered / displayed list of items. In which case it would make sense that the checked state reflected the visible list. Just not sure if it's possible. Will see.

mark-prins: turns out, not that hard. and much nicer to use 😁 
thanks!
97926b7b

CarlosNZ (approved):
Looking great, thanks for the changes, Indeterminate state looks great.

There is one little quirk that would be good to fix -- although I had a quick play around and couldn't solve it.

The "clear selection" icon (X) in the input field is confusing. It looks like it should clear the text input, but it actually clears the selection.
Screen Shot 2022-03-25 at 5 17 50 PM

I thought this was a bug at first, but it's the intended behaviour. However for the [examples on the Mui docs](https://mui.com/components/autocomplete/#checkboxes) page, the selected items are being collected in the input field so it makes more sense (and with the Select All it's extra confusing cos the Select All can already be used to toggle/clear the selection).

Ideally it'd be good to override the behaviour of the icon to clear the input field not the selection, but not sure that's possible (based on a quick browse of the API docs). But even if we could hide it completely, that would be better do you think? (There's a `clearIcon` prop that can *change* the icon, so hopefully it can be made to hide or disable the icon instead)

mark-prins:
heh. Funny you should say that.. I found the same problem with the clear icon. I had added a tooltip.. but still found myself clicking it to clear the input.
Have put a fix in for it.. thanks for the prompt to get it done! 😉

CarlosNZ (approved):
Nice :) 

Small suggestion for that typescript issue, not sure if it actually helps 🤷‍♂️

packages/common/src/ui/components/inputs/Autocomplete/AutocompleteList.tsx
```diff
@@ -31,6 +31,11 @@ export type AutocompleteListProps<T> = {
   limitTags?: number;
   inputValue?: string;
   clearText?: string;
+  // annoying. typing this to T | T[] causes ts complaints, as it types the value as (T | T[])[]
+  // as the correct typing relies on other props here
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  value?: any;
```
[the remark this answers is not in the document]
mark-prins: ah! thanks 😁  that's awesome, just what I need. I went and merged before seeing this.. will sneak it in the next one.

Should be `value?: T extends any[] ? T[] : T;`?

PR #1111, opened 2022-03-24
Refresh outbound list cache on delete

### LINKED ISSUE #1107: Bug: Deleting shipment rows from Outbound Shipment table doesn't refresh the table.
The deletion works okay and there is a Snackbar notification, but we have to refresh the table for the line to actually disappear from the UI.

Seems to be just on Outbound shipments, Inbound is working as expected. Haven't tried with any other tables.

mark-prins:
Fixes #1107 

Just needed to update the cache - the cache key was missed in some recent refactoring

### DIFF
--- a/packages/invoices/src/OutboundShipment/DetailView/OutboundServiceLineEdit/useServiceLineColumns.tsx
+++ b/packages/invoices/src/OutboundShipment/DetailView/OutboundServiceLineEdit/useServiceLineColumns.tsx
@@ -9,6 +9,7 @@ import {
   ColumnAlign,
   NonNegativeNumberInputCell,
   TextInputCell,
+  useFormatCurrency,
 } from '@openmsupply-client/common';
 import { ServiceItemSearchInput } from '@openmsupply-client/system';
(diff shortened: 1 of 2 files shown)


packages/invoices/src/OutboundShipment/DetailView/OutboundServiceLineEdit/useServiceLineColumns.tsx
```diff
@@ -63,7 +65,7 @@ export const useServiceLineColumns = (
       label: 'label.total',
       align: ColumnAlign.Right,
       width: 75,
-      accessor: ({ rowData }) => get.serviceChargeTotal(rowData),
+      accessor: ({ rowData }) => c(get.serviceChargeTotal(rowData)),
```
mark-prins, line 68: offroad - I noticed that the service item currency format was incorrect. 
( was showing 7.07777777777...  when I had $7 with a tax of 10% 🙄  )

CarlosNZ (approved):
LGTM!

PR #1112, opened 2022-03-25
#1099 Dropdown menu -- actions always available

### LINKED ISSUE #1099: "Select" operation dropdown should only be displayed when it has content.
Currently, on some pages (e.g.Supplier requisitions), the operation menu has no options, so the control (showing "Select") does nothing. Suggestion would be to update such that:

- if no operations are available, show nothing
- if only one operation is available, just show that operation as a button (e.g. "Delect selected lines" on "Internal Orders")
- if more than one operation, show the drop-down selector as it currently does

CarlosNZ:
Fix #1099 

Not _exactly_ what was proposed in Issue #1099 as there was a few snags found along the way. I have gone more for overall consistency as we discussed -- there is always a button or dropdown there, but if its action is invalid it will notify you (such as deleting lines on a Completed invoice).

It was a bit tricky to keep track of all the different places this was used, as each of them had their "Delete" actions handled *slightly* differently. Have tried to make the overall behaviour consistent, but haven't restructured anything, just copied the same patterns when required.

I'm not sure about changing to a clickable link when there's only one option -- see what you think.

See inline comments for specifics


packages/common/src/ui/components/menus/DropdownMenu/DropdownMenu.tsx
```diff
@@ -80,25 +80,27 @@ export const DropdownMenu: FC<DropdownMenuProps> = ({
...
+          size="small"
+          labelId={`action-drop-down-label-${label}`}
+          variant="outlined"
+          IconComponent={ChevronDownIcon}
+        >
+          {children}
+        </StyledSelect>
+      </FormControl>
+    );
+  else return <div>{children}</div>;
```
CarlosNZ, line 105: Not sure about this -- might be better to leave it as it was. It's probably a bit risky to have a "Delete" button so easily accessible, at least when it's inside a Dropdown you have to be a little more intentional, therefore less likely to accidentally click it.

CarlosNZ, line 105: Not sure about this -- might be better to leave it as it was. It's probably a bit risky to have a "Delete" button so easily accessible, at least when it's inside a Dropdown you have to be a little more intentional, therefore less likely to accidentally click it.

packages/requisitions/src/ResponseRequisition/api/api.ts
```diff
@@ -123,6 +125,25 @@ export const getResponseQueries = (sdk: Sdk, storeId: string) => ({
...
+    //       line =>
+    //         line.response.__typename === 'DeleteRequestRequisitionLineError'
+    //     );
+    //   if (failedLines.length === 0) {
+    //     return result.batchRequestRequisition.deleteRequestRequisitionLines;
+    //   }
+    // }
+
+    throw new Error('Could not delete requisition lines!');
+  },
```
CarlosNZ, line 146: Was part-way through adding this "Delete" action when I realised there is not yet any back-end API for `deleteResponseRequisitionLine` (only `deleteRequestRequisitionLine`). However I figured I may as well leave the boilerplate in as a placeholder (it's in a few places, basically a copy of the "Request" equivalent) until it's ready.

packages/requisitions/src/ResponseRequisition/api/hooks.ts
```diff
@@ -200,6 +207,43 @@ export const useCreateOutboundFromResponse = () => {
   });
 };
 
+export const useDeleteResponseLines = () => {
```
CarlosNZ, line 210: As per previous comment -- this calls a mutation that doesn't actually do anything yet (waiting for back-end)

packages/requisitions/src/ResponseRequisition/api/hooks.ts
```diff
@@ -200,6 +207,43 @@ export const useCreateOutboundFromResponse = () => {
...
+  const selectedRows = useTableStore(state =>
+    lines.filter(({ id }) => state.rowState[id]?.isSelected)
+  );
+
+  const onDelete = async () => {
+    if (isDisabled) {
+      info(t('label.cant-delete-disabled-requisition'))();
+      return;
+    }
+    info('Deleting response lines not yet implemented in API')();
```
CarlosNZ, line 232: Hardcoded this string as its only meant to be temporary

packages/requisitions/src/RequestRequisition/api/hooks.ts
```diff
@@ -320,6 +321,10 @@ export const useDeleteRequestLines = () => {
   );
 
   const onDelete = async () => {
+    if (isDisabled) {
+      info(t('label.cant-delete-disabled'))();
```
mark-prins, line 325:
```suggestion
      info(t('label.cant-delete-disabled-requisition'))();
```

the key `label.cant-delete-disabled` isn't in `replenishment` namespace

packages/common/src/ui/components/menus/DropdownMenu/DropdownMenu.tsx
```diff
@@ -80,25 +80,27 @@ export const DropdownMenu: FC<DropdownMenuProps> = ({
...
-        {children}
-      </StyledSelect>
-    </FormControl>
-  );
+  if (Array.isArray(children))
+    return (
+      <FormControl size="small">
+        <InputLabel
+          shrink={false}
+          sx={{ color: '#8f90a6', '&.Mui-focused': { color: '#8f90a6' } }}
```
mark-prins, line 88:
```suggestion
          sx={{ color: 'gray.main', '&.Mui-focused': { color: 'gray.main' } }}
```

if you wouldn't mind. not sure how those colours slipped through

packages/common/src/ui/components/menus/DropdownMenu/DropdownMenu.tsx
mark-prins: possibly, given that the action in almost all cases is 'delete'. It's also more consistent having the drop down everywhere.

I really like the ease of having the button available without the select.. but am worrying that you are right. Also.. didn't want to abandon all this work! 

Am favouring reverting back to the select.. but could go either way.. thoughts?

mark-prins, line 105: if we do stick with the button - it was looking a little lost, and I wondered about this:

```
const StyledContainer = styled('div')(({ theme }) => ({
  borderWidth: 1,
  borderStyle: 'solid',
  borderColor: theme.palette.gray.light,
  borderRadius: 5,
  '& > li': { height: '100%' },
}));
```

and 
```
  else return <StyledContainer>{children}</StyledContainer>;
```

giving

Screen Shot 2022-03-25 at 3 21 11 PM

packages/invoices/src/InboundShipment/DetailView/Toolbar.tsx
```diff
@@ -73,7 +73,7 @@ export const Toolbar: FC = () => {
             />
           </Box>
         </Grid>
-        <DropdownMenu disabled={isDisabled} label={t('label.select')}>
+        <DropdownMenu label={t('label.select')}>
```
mark-prins, line 76: I've always had a problem with 'Select' here.. what do you think about changing the label to 'Actions' instead?
The word select doesn't really tell me what is going to happen when I click it 🤷

mark-prins (approved):
👍  very nice!

I'm now having second thoughts about exposing the delete button directly - along with the concern that the look is inconsistent. Even if we don't have the button bit - changing to have the toast notification consistent is a win 😁

packages/requisitions/src/RequestRequisition/api/hooks.ts
CarlosNZ: Ah, I knew I'd miss one! Thanks.

packages/common/src/ui/components/menus/DropdownMenu/DropdownMenu.tsx
CarlosNZ: Yeah, let's leave it how it was until we can be bothered adding a Warning modal when deleting stuff. Now that I've done this, it's an easy change to put back when we need it.

CarlosNZ: Reverted 97b222e9
Made new issue: #1115

CarlosNZ: Yeah, much better. Have made reference to this on [new issue](#1115)

packages/invoices/src/InboundShipment/DetailView/Toolbar.tsx
CarlosNZ: Yeah good idea: 3863acf2

Issue #1101, opened 2022-03-23
Give better feedback about database errors


CarlosNZ:
Currently, if there is a database error when making a query the UI shows "No records to display". This could be misleading -- user thinks there are no items when there was actually a database problem. 

We should probably catch database errors more specifically and actually display some kind of error alert to the user.

mark-prins:
Had created a new issue with the following comment.. but had intended that the changes be done on this issue.

Following on from #1113 - that one has added a global error catch in the graphql request implementation.

There's also an error handling change in #1125 which throws an error in the GqlContext when the response has an error, which allows graphql-request to respond in the `onError` callback. That part could be lifted out into an error boundary and use notistack to toast on error as a general handler for error objects being returned.

Would be nice to make use of different levels of error boundaries in graphql-request, such as noted in [this comment](https://github.com/openmsupply/openmsupply-client/pull/1113#issuecomment-1078748253) and respond differently given the severity of the problem

PR #1113, opened 2022-03-25
authentication check

### LINKED ISSUE #1096: Redirect to login screen when logged out due to inactivity
Should be able to detect when token disappears, then re-direct to `/login` page

mark-prins:
Fixes #1096, #1101 

Yes, that's right! two-for-one time! 😁 

These two are somewhat interrelated, so I've combined them in one PR. uh, sorry.

Few things
- Checking the response of every gql request and if it has the 'unauthenticated' error, then pop that into local storage. Ideally I would have updated the `AuthContext` directly but there is a problem of dependency. The AuthContext uses a gql request so has to sit under the gql provider. Also slightly annoying using the AuthContext without referencing hooks directly.. so it was good to sidestep that
- The `LoggedOutAlert` is now more generic and so has been renamed
- The `AuthContext` is running an interval, checking the cookie and if it has expired, or no token - sets an `AuthError`
- The `AuthenticationAlert` is using the `AuthError` from `AuthContext` and will re-render if that is updated by the interval in `AuthContext`

### DIFF
--- a/packages/common/src/api/GqlContext.tsx
+++ b/packages/common/src/api/GqlContext.tsx
@@ -1,14 +1,83 @@
 import React, { FC, useMemo, useEffect, useState, useCallback } from 'react';
 import { createContext } from 'react';
-import { GraphQLClient } from 'graphql-request';
+import {
+  BatchRequestDocument,
+  GraphQLClient,
+  RequestDocument,
+  Variables,
+} from 'graphql-request';
+import { AuthError } from '../authentication/AuthContext';
+import { LocalStorage } from '../localStorage';
 
-export const createGql = (url: string): { client: GraphQLClient } => {
-  const client = new GraphQLClient(url, { credentials: 'include' });
+class GQLClient extends GraphQLClient {
+  private client: GraphQLClient;
+  private emptyData: object;
+
+  constructor(url: string, options?: RequestInit) {
+    super(url, options);
+    this.client = new GraphQLClient(url, options);
+    this.emptyData = {};
+  }
+
+  public rawRequest<T, V = Variables>(
+    query: string,
+    variables?: V,
+    requestHeaders?: RequestInit['headers']
+  ): Promise<{
+    data: T;
+    extensions?: any;
+    headers: Headers;
+    status: number;
+  }> {
+    return this.client.rawRequest(query, variables, requestHeaders);
+  }
+
+  public request<T, V = Variables>(
+    document: RequestDocument,
+    variables?: V,
+    requestHeaders?: RequestInit['headers']
+  ): Promise<T> {
+    const response = this.client.request(document, variables, requestHeaders);
+    // returning an empty object in order to give the caller a stable reference
+    // without it, the page will re-render continuously
+    return response.then(
+      data => data ?? this.emptyData,
+      ({ response }) => {
+        if (response && response.errors) {
+          if (
+            response.errors.some(
+              ({ message }: { message?: string }) =>
+                message === 'Unauthenticated'
+            )
+          ) {
+            LocalStorage.setItem('/auth/error', AuthError.Unauthenticated);
+          }
+        }
+        return this.emptyData;
+      }
+    );
+  }
+  public batchRequests<T, V = Variables>(
+    documents: BatchRequestDocument<V>[],
+    requestHeaders?: RequestInit['headers']
+  ): Promise<T> {
+    return this.client.batchRequests(documents, requestHeaders);
+  }
+  public setHeaders = (headers: RequestInit['headers']): GraphQLClient =>
+    this.client.setHeaders(headers);
+  public setHeader = (key: string, value: string): GraphQLClient =>
+    this.client.setHeader(key, value);
+  public setEndpoint = (value: string): GraphQLClient =>
+    this.client.setEndpoint(value);
+}
+
+export const createGql = (url: string): { client: GQLClient } => {
+  const client = new GQLClient(url, { credentials: 'include' });
   return { client };
 };
 
 interface GqlControl {
-  client: GraphQLClient;
+  client: GQLClient;
   setHeader: (header: string, value: string) => void;
   setUrl: (url: string) => void;
 }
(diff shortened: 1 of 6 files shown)


packages/common/src/api/GqlContext.tsx
```diff
@@ -1,14 +1,83 @@
...
+  GraphQLClient,
+  RequestDocument,
+  Variables,
+} from 'graphql-request';
+import { AuthError } from '../authentication/AuthContext';
+import { LocalStorage } from '../localStorage';
 
-export const createGql = (url: string): { client: GraphQLClient } => {
-  const client = new GraphQLClient(url, { credentials: 'include' });
+class GQLClient extends GraphQLClient {
```
mark-prins, line 12: extending it solely to avoid typing problems elsewhere. the fun of using a codegen 🙄

packages/common/src/api/GqlContext.tsx
```diff
@@ -1,14 +1,83 @@
...
+          if (
+            response.errors.some(
+              ({ message }: { message?: string }) =>
+                message === 'Unauthenticated'
+            )
+          ) {
+            LocalStorage.setItem('/auth/error', AuthError.Unauthenticated);
+          }
+        }
+        return this.emptyData;
```
mark-prins, line 56: as noted.. the reason that internal orders were endlessly rendering was because the gql response was returning `null` for `data` which can't be used as a stable reference for the component.

Instead, create an object in the class and return that, so it is stable.

packages/common/src/authentication/AuthContext.tsx
```diff
@@ -9,6 +9,12 @@ import { useGetAuthToken } from './api/hooks/useGetAuthToken';
 import { AuthenticationResponse } from './api';
 
 export const COOKIE_LIFETIME_MINUTES = 60;
+const TOKEN_CHECK_INTERVAL = 60 * 1000;
+
+export enum AuthError {
+  PermissionDenied = 'PermissionDenied',
```
mark-prins, line 15: No idea what this will look like as an api response - should be easy to add in when the api implements though

packages/common/src/authentication/AuthContext.tsx
```diff
@@ -9,6 +9,12 @@ import { useGetAuthToken } from './api/hooks/useGetAuthToken';
 import { AuthenticationResponse } from './api';
 
 export const COOKIE_LIFETIME_MINUTES = 60;
+const TOKEN_CHECK_INTERVAL = 60 * 1000;
+
+export enum AuthError {
```
mark-prins, line 14: Initially called `AuthenticationError` but there's already one in `./api` 🙄

packages/common/src/authentication/AuthContext.tsx
```diff
@@ -102,6 +110,7 @@ export const AuthProvider: FC = ({ children }) => {
   const { mutateAsync, isLoading: isLoggingIn } = useGetAuthToken();
   const authCookie = getAuthCookie();
   const [cookie, setCookie] = useState<AuthCookie | undefined>(authCookie);
+  const [error, setError] = useLocalStorage('/auth/error');
```
CarlosNZ, line 113: Just trying to clarify my understanding -- is the value of the AuthError in localStorage only used to tell the AuthenticationAlert which message to display?

CarlosNZ (approved):
Great, seems to be working well, although I can't test what a real authentication error from the server looks like.

I'm assuming we'll need to tweak this once we have more standardised auth/permission errors from the server?

josh-griffin:
:)

FWIW, my dump of what I was thinking re errors:

- Create an OmSupplyError class to be used with some specific fields to be able to check things - like an error 'type' (Or more specific error classes.. 🤷  )
- Queries should generally not throw errors often. `useErrorBoundary: true` for most queries and just throw an error wrapped in OmSupplyError to be caught by specific error boundaries: PermissionErrorBoundary, NotFoundBoundary which can just render `<>{children}{hasError ? <AlertModal/> : null}</>` for example so this logic isn't needed on every page.
- Possible alternative I had been thinking of was to modify the client like this to throw errors, then  use a function for the `useErrorBoundary` on the query client for [default options](https://react-query.tanstack.com/reference/QueryClient#queryclientsetdefaultoptions) to check the error is for example an auth error, and return `true`, to use the boundary for those errors globally.
- Mutations to also throw an `OmSupplyError` but should usually be handled in the `onError` callback of a mutation to just toast. But without using an error boundary, the `error` prop is also set so can be used for more specific things also

----

I'm really interested in the constant rendering when `Internal orders detail page will re-render infinitely when you are logged out` .. seems like a good learning opportunity - how can I repro that?

😁

packages/common/src/api/GqlContext.tsx
```diff
@@ -1,14 +1,83 @@
...
+    // returning an empty object in order to give the caller a stable reference
+    // without it, the page will re-render continuously
+    return response.then(
+      data => data ?? this.emptyData,
+      ({ response }) => {
+        if (response && response.errors) {
+          if (
+            response.errors.some(
+              ({ message }: { message?: string }) =>
+                message === 'Unauthenticated'
```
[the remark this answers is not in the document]
mark-prins: excellent! thanks

packages/common/src/authentication/AuthContext.tsx
mark-prins: yes - using local storage because I need to retain state and can't access the AuthContext from the Gql provider

mark-prins:
Thanks @josh-griffin! That's super helpful. I started off with an error boundary and stopped because I didn't want the unhappy man handling all the auth errors.. didn't really consider having multiple boundaries! 🙄 Will tidy up a few things and then look at reworking this area.

> I'm really interested in the constant rendering when Internal orders detail page will re-render infinitely when you are logged out .. seems like a good learning opportunity - how can I repro that?

if you comment out line 56 in `GqlContext` and then try to view a requisition while logged out - any req query will do it, I was using the internal order list. If the token is empty you'll get an 'Unauthenticated' response, and the `data` returned is undefined. The page is using that as a reference and being undefined it re-renders and fires off the query again.. which repeats the process. 

To stop it, I've added a variable to the context and will return that on error, so that react-query can use it as a stable reference. It also means that you get the message 'no results' which doesn't apply because there was a problem - but that's where the error boundary comes in! 😁

PR #1119, opened 2022-03-28
#1097 Table column widths

### LINKED ISSUE #1097: Table column widths -- improve
Currently some unnecessarily wide and we have to resort to side-scrolling when the right-hand sidebar slides out.

Ideally more columns could have an automatic width based on content. Suggestion is to use the current hard-coded widths as *max* widths instead. 

Need to investigate to see what will work / look good.

CarlosNZ:
Fix #1097 

General gist is:
- remove defaults for column `minWidth` and `width` (Table will auto-width unless custom width style is specified)
- remove "flex" layouts on the table elements (Header, Body, Row) as it was making a mess of the column widths per row once defaults were removed

Have add custom widths to some columns, but they're pretty arbitrary just based on me eye-balling it with the existing data -- will probably want to evolve them over time.

Have checked all the tables, including ones with grouped lines, and they seem to be behaving themselves.

### DIFF
--- a/packages/common/src/ui/components/inputs/CurrencyInput/CurrencyInput.tsx
+++ b/packages/common/src/ui/components/inputs/CurrencyInput/CurrencyInput.tsx
@@ -7,7 +7,7 @@ import { useCurrency } from '@common/intl';
 
 interface CurrencyInputProps extends RCInputProps {
   onChangeNumber: (value: number) => void;
-  maxWidth?: number;
+  maxWidth?: number | string;
 }
 
 // TODO: It would be nice if we were to just use the BasicTextInput or
--- a/packages/common/src/ui/layout/tables/DataTable.tsx
+++ b/packages/common/src/ui/layout/tables/DataTable.tsx
@@ -114,9 +114,6 @@ export const DataTableComponent = <T extends RecordWithId>({
         <TableHead
           sx={{
             backgroundColor: 'background.white',
-            flex: 1,
-            display: 'flex',
-            flexDirection: 'column',
             position: 'sticky',
             top: 0,
             zIndex: 'tableHeader',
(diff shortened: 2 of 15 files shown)


packages/common/src/ui/components/inputs/CurrencyInput/CurrencyInput.tsx
```diff
@@ -7,7 +7,7 @@ import { useCurrency } from '@common/intl';
 
 interface CurrencyInputProps extends RCInputProps {
   onChangeNumber: (value: number) => void;
-  maxWidth?: number;
+  maxWidth?: number | string;
```
CarlosNZ, line 10: Included string so other units can be passed down (%, ems, etc) as per standard css.

mark-prins:
The border radius on the expando suffers a little:

Screen Shot 2022-03-29 at 8 59 27 AM

the problem is with the `th` having a background colour, which is due to the use of the table in the edit modals.
maybe we just lose the corners on the top?

`border-radius: 0 0 16px 16px;`

Screen Shot 2022-03-29 at 9 05 04 AM

mark-prins (approved):
🎉  that's fantastic thanks!
looking much better - and should be easier to maintain using tables rather than a mix of table & flex.

The columns widths added all look good on my screen 😁

CarlosNZ:
> The border radius on the expando suffers a little:

> the problem is with the `th` having a background colour, which is due to the use of the table in the edit modals. maybe we just lose the corners on the top?

Ah, good catch. Not sure why it behaves different than it did before. Anyway, can be fixed by setting `overflow: hidden` on the Mini-table container (same level as the border setting)

32006bc5

mark-prins:
One other thing which I forgot to mention.. there is now a hover state on rows which we didn't have before - noticeable particularly in the outbound edit table, where the rows aren't really active, so shouldn't have the hover state.

CarlosNZ:
> One other thing which I forgot to mention.. there is now a hover state on rows which we didn't have before - noticeable particularly in the outbound edit table, where the rows aren't really active, so shouldn't have the hover state.

I didn't think anything relating to the row hover state was changed here, was it? 

I agree the hover rows in the Edit Modal look wrong -- not sure when that was introduced (although it is just following the default Mui hover behaviour). I've made a new issue to deal with it. (Hopefully just a matter of turning off a prop in the Mui Table component). https://github.com/openmsupply/openmsupply-client/issues/1128

PR #1121, opened 2022-03-28
#1114 Stockdate date fix

### LINKED ISSUE #1114: Stocktake -- date should be blank by default
Currently it defaults to 1/1/1970, probably should be current date, shouldn't it?

If the current date is not appropriate for some reason, then it should probably have an "empty" date value rather than 1/1/1970

CarlosNZ:
Fix #1114 

Turns out the 1/1/1970 date wasn't being saved, it was just a small problem with the Date formatter not handling `null` values correctly.

Slightly tangential: noticed that the sort headers weren't working on this table. Turns out that API isn't accepting those columns as sort keys -- have made an issue for back-end: https://github.com/openmsupply/remote-server/issues/981


mark-prins (approved):
👍  awesome! that's something which was bothering me.. nice to have it sorted!

PR #1122, opened 2022-03-28
#1120 Page title hook

### LINKED ISSUE #1120: Update page title
Currently the page title always shows "localhost:3003/dashboard"
Screen Shot 2022-03-28 at 8 00 42 PM

We should update it depending on the current location.

Suggest "**\<PageName\> | mSupply**" format

CarlosNZ:
Fix #1120 

I realise this doesn't *need* to be a hook -- could just be a simple function that calls `document.title = newTitle`, and indeed it's not really using the returned items `pageTitle` and `setPageTitle`. However, if we ever need to do dynamic title updating the `setPageTitle` function would be useful.

Have only implemented this on Dashboard, Outbound and Inbound shipments so far. If this approach is okay, I'll add `usePageTitle` to the other pages before merging. :)

### DIFF
--- a/packages/common/src/hooks/usePageTitle/usePageTitle.tsx
+++ b/packages/common/src/hooks/usePageTitle/usePageTitle.tsx
@@ -1,0 +1,15 @@
+import { useEffect, useState } from 'react';
+import { useTranslation } from '@common/intl';
+
+export const usePageTitle = (title: string) => {
+  const t = useTranslation('common');
+  const [pageTitle, setPageTitle] = useState(title);
+
+  useEffect(() => {
+    const newTitle = `${title} | ${t('app')}`;
+    setPageTitle(newTitle);
+    document.title = newTitle;
+  }, [title]);
+
+  return { pageTitle, setPageTitle };
+};
--- a/packages/common/src/intl/locales/en/common.json
+++ b/packages/common/src/intl/locales/en/common.json
@@ -1,4 +1,5 @@
 {
+  "app": "mSupply",
   "app.admin": "Admin",
   "app.catalogue": "Catalogue",
   "app.customer-requisition": "Requisitions",
--- a/packages/common/src/hooks/index.ts
+++ b/packages/common/src/hooks/index.ts
@@ -16,6 +16,7 @@ export * from './useIsMountedRef';
 export * from './useIsScreen';
 export * from './useNotification';
 export * from './useOpenInNewTab';
+export * from './usePageTitle';
 export * from './usePagination';
 export * from './useQueryParams';
 export * from './useSortBy';
--- a/packages/common/src/hooks/usePageTitle/index.ts
+++ b/packages/common/src/hooks/usePageTitle/index.ts
@@ -1,0 +1,1 @@
+export * from './usePageTitle';
--- a/packages/dashboard/src/DashboardService.tsx
+++ b/packages/dashboard/src/DashboardService.tsx
@@ -1,24 +1,28 @@
 import React from 'react';
-import { Grid } from '@openmsupply-client/common';
+import { Grid, usePageTitle, useTranslation } from '@openmsupply-client/common';
 import { StockWidget } from './widgets';
 import {
   InboundShipmentWidget,
   OutboundShipmentWidget,
 } from '@openmsupply-client/invoices';
 
-const Dashboard: React.FC = () => (
-  <Grid
-    container
-    sx={{
-      backgroundColor: 'background.toolbar',
-      paddingBottom: '32px',
-    }}
-    justifyContent="space-evenly"
-  >
-    <InboundShipmentWidget />
-    <OutboundShipmentWidget />
-    <StockWidget />
-  </Grid>
-);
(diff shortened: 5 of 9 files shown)


packages/common/src/hooks/usePageTitle/usePageTitle.tsx
```diff
@@ -0,0 +1,15 @@
+import { useEffect, useState } from 'react';
+import { useTranslation } from '@common/intl';
+
+export const usePageTitle = (title: string) => {
+  const t = useTranslation('common');
+  const [pageTitle, setPageTitle] = useState(title);
+
+  useEffect(() => {
+    const newTitle = `${title} | ${t('app')}`;
```
CarlosNZ, line 9: We happy with this title format?

packages/common/src/intl/locales/en/common.json
```diff
@@ -1,4 +1,5 @@
 {
+  "app": "mSupply",
```
CarlosNZ, line 2: Thought this seemed appropriate for the "root" app name, though can change to `app.name` if that's preferable.

packages/common/src/hooks/usePageTitle/usePageTitle.tsx
mark-prins: yup. looks very slick next to this site 😁 

Screen Shot 2022-03-29 at 10 18 15 AM

packages/common/src/intl/locales/en/common.json
mark-prins: I like `app.name` but it would be confusing next to the other entries. what you have is probably best

mark-prins (approved):
👍  that's very pleasing to see.. I had only noticed the missing page title a short time ago and it has bugged me since

I do wonder if the state should be in `HostContext` rather than creating a new state within the hook, though it would make it slightly clumsier to use perhaps. something like

```
const { setPageTitle } = useHostContext();
setPageTitle('blah');
```

though I do like setting a value by calling the `set...` method rather than a `use..` method 🤷 

Another thought, is for the Host to set the page title based on the route - it should be possible to set the title without explicitly adding the setter to every page, we just need a mapping between the route and the translation key

CarlosNZ:
> I do wonder if the state should be in `HostContext` rather than creating a new state within the hook, though it would make it slightly clumsier to use perhaps. something like

Yeah that seems good.

> Another thought, is for the Host to set the page title based on the route - it should be possible to set the title without explicitly adding the setter to every page, we just need a mapping between the route and the translation key

Yeah, I considered that, but there was a little discrepancy between the route and the name in terms of plurals -- we'd want the title to be be "Outbound Shipments" not "Outbound Shipment", whereas we don't want to pluralise "Dashboard", so can't have a fixed rule. Also, would have to have a bit of logic to figure out whether to take the last part of the route or the 2nd-to-last: `/distribution/outbound-shipment` should become "Outbound Shipments", but `/distribution/outbound-shipment/6` should also become just "Outbound Shipments".

I do like the idea of doing it automatically though (bit tedious to add them all in -- lazy!) so how about using an automatic title inferrer by default, but allowing specific setting through`setPageTitle` if required? 

I'll make the logic:
- use last path part, unless it includes a number, in which case use 2nd-to-last
- convert to Title Case (using `lodash` handy case conversion methods)
- always pluralise (but set a custom title for "Dashboard")

CarlosNZ:
> we just need a mapping between the route and the translation key

I realise now that you didn't mean automatic case conversion of route -> title, as that wouldn't even work with translations. You meant a custom map of routes to keys. That should be fine to do, much simpler than what I was suggesting, heh.

mark-prins:
You're a step ahead of me.. I was typing out a reply when yours popped in!
yeah - I was first wondering about mapping a route portion to a translation key and then about a static map of routes to keys. It's still a little tedious, but fewer files to touch.. which is something?

we could use the same translation as the menu items too, and just ignore plurals like we do there 🤷

CarlosNZ:
> we could use the same translation as the menu items too, and just ignore plurals like we do there 🤷

Yeah, it should be quite straightforward -- just have a `switch` block with simple regexes for each (group of) routes in each case. Doing that now.

Just to clarify:
- I'll put the `pageTitle` state and a `setPageTitle` method in `useHostContext`, along with an internal function to map routes to titles
- A `useEffect` in `<Site/> component (lowest level that is inside Router) to execute `setPageTitle` whenever the `location` changes.

Sound about right?

mark-prins:
perfect. sounds good to me 😁

josh-griffin:
So uh, funny story..

image

I just came here to check if we had written this hook so that I could take it real quick.

Open source, right? Thanks!

CarlosNZ:
Was easier to start a new branch based on suggestions/discussion: https://github.com/openmsupply/openmsupply-client/pull/1127

CarlosNZ:
> So uh, funny story..
> 
> image
> 
> I just came here to check if we had written this hook so that I could take it real quick.
> 
> Open source, right? Thanks!

Anyone know who this @josh-griffin guy is? Never heard of him.

PR #1123, opened 2022-03-28
Item detail mock

### LINKED ISSUE #1118: Item Detail : mock data
With [PR #965](https://github.com/openmsupply/remote-server/pull/965) some mock data is available for items.
Time to try a detail view!

mark-prins:
Fixes #1118 

Sorry! there's a few changes to bring this in.. nothing complicated though, just a few files involved.
Have added a new skeleton for the detail form and some common elements used to construct this new style of detail page.


mark-prins:
example skeleton and page:

Screen Shot 2022-03-29 at 11 48 49 AM

Screen Shot 2022-03-29 at 11 48 41 AM

packages/common/src/ui/components/domain/Dashboard/StatsPanel.tsx
```diff
@@ -10,12 +10,14 @@ export interface StatsPanelProps {
   isLoading: boolean;
   stats: Stat[];
   title: string;
+  width?: number;
```
mark-prins, line 13: so that could constrain the width on the item detail page

packages/common/src/ui/components/navigation/Tabs/Tabs.tsx
```diff
@@ -23,4 +23,5 @@ export const Tab = styled(MuiTab)({
   height: 50,
   fontWeight: 'bold',
   textTransform: 'none',
+  whiteSpace: 'nowrap',
```
mark-prins, line 26: looks nicer 🤷  
the other tab usages don't seem troubled by the change

packages/common/src/ui/forms/Detail/Container.tsx
```diff
@@ -0,0 +1,8 @@
+import React, { FC } from 'react';
+import { Grid } from '@mui/material';
+
+export const DetailContainer: FC = ({ children }) => (
```
mark-prins, line 4: The naming of all these components is ropey at best.. could probably do better

packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/hooks/useNextItem.test.tsx
```diff
@@ -69,6 +69,20 @@ const getLines = (): InvoiceLineNode[] => [
         availableMonthsOfStockOnHand: 0,
         averageMonthlyConsumption: 0,
       },
+      atcCategory: '',
```
mark-prins, line 72: annoying. this is the quick fix.. but it isn't great 🙄

packages/system/src/Item/DetailView/Tabs/MasterLists.tsx
```diff
@@ -0,0 +1,34 @@
...
+import {
+  DataTable,
+  TableProvider,
+  useColumns,
+  createTableStore,
+  Box,
+} from '@openmsupply-client/common';
+
+export const MasterListsTab = () => {
+  const { data, isLoading } = useMasterLists();
```
mark-prins, line 16: yes, this is nonsense. but it was fun to do 😁

packages/system/src/Item/api/hooks/useItem/useItem.ts
```diff
@@ -1,11 +1,22 @@
...
+  const itemId = useItemId();
+  return useItemById(itemId);
+};
+
+export const useItemById = (itemId: string | undefined) => {
   const api = useItemApi();
   return useQuery(
-    api.keys.detail(itemId ?? ''),
-    () => api.get.byId(itemId ?? ''),
+    api.keys.detail(itemId || ''),
```
mark-prins, line 18: oh. oops - that was unintentional, just rewrote this when adding in the option to not specify an id

packages/system/src/Name/api/operations.graphql
```diff
@@ -10,6 +10,31 @@ fragment NameRow on NameNode {
   }
 }
 
+fragment Name on NameNode {
```
mark-prins, line 13: popped this in while I was here..

packages/common/src/ui/forms/Detail/Section.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+interface DetailSectionProps {
+  title: string;
+}
+export const DetailSection: FC<DetailSectionProps> = ({ children, title }) => (
+  <Grid container flex={1} flexDirection="column" gap={1}>
+    <Grid item display="flex">
+      <Box flex={1} flexBasis="40%"></Box>
+      <Box flex={1} flexBasis="60%">
+        <Typography fontWeight={700} fontSize={16}>
```
CarlosNZ, line 12: Be good to standardise font-sizes etc. into the theme. Conversation for another day perhaps, just putting it out there for now.

packages/common/src/ui/forms/Detail/InputWithLabelRow.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  return (
+    <Box
+      display="flex"
+      alignItems="center"
+      gap={1}
+      justifyContent="space-around"
+    >
+      <Box flex={1} style={{ textAlign: 'end' }} flexBasis="40%">
+        <FormLabel sx={{ fontWeight: 'bold', ...sx }} {...labelPropsRest}>
```
CarlosNZ, line 33:
```suggestion
        <FormLabel sx={{ ...sx, fontWeight: 'bold' }} {...labelPropsRest}>
```
Would this be better? (in case `fontWeight` already has a value, we would want to override it)

packages/system/src/Item/DetailView/Tabs/General.tsx
```diff
@@ -0,0 +1,141 @@
...
+            }}
+          />
+          <DetailInputWithLabelRow
+            label={t('label.universal-code')}
+            inputProps={{
+              value: data?.msupplyUniversalCode,
+              disabled: isDisabled,
+            }}
+          />
+        </DetailSection>
```
CarlosNZ, line 76: Would it be possible to generate these from the query response fields themselves rather than hard-coding each field? Iterate over the `data` object fields? Could even use the field name itself as the key for the language string (i.e. ``label={t(`label.${fieldName}`)}``)

Then we can use the same DetailView component for other Detail modals. And if the data fields from the back-end changes, we won't have to change the front-end.

I guess I'm suggesting a generalise Detail view component like the DataTable component is for List views.

packages/system/src/Item/DetailView/Tabs/General.tsx
```diff
@@ -0,0 +1,141 @@
...
+      </Grid>
+
+      <Grid
+        container
+        flex={1}
+        flexDirection="column"
+        style={{ maxWidth: 500 }}
+        gap={4}
+      >
+        <DetailSection title={t('title.packaging')}>
```
CarlosNZ, line 86: I guess same comment for the sections?

packages/system/src/Item/DetailView/Tabs/MasterLists.tsx
CarlosNZ: Lol. Is there any items that belong to more than one Master list?

CarlosNZ (approved):
Thanks Mark, looking really good.

I have a few random thoughts aside from my inline comments. We can discuss them at some point when we want to refine this a bit more:

- The stats panels are great, the only problem is that the page has to be really wide (~1400px) for them to all show up nicely. I think this is wider than most tablets, isn't it?
Screen Shot 2022-03-29 at 4 39 22 PM
   Once they start collapsing they don't look so good, as there is a lot of wasted space on the left:
Screen Shot 2022-03-29 at 4 42 58 PM
  It might be better if the three panels stayed in one row, but the whole row moved under the Name/Code block. Not sure what the best option might be.
- I feel like the `disabled` text colour is a bit *too* pale for this situation. I know it's a global setting so would need to be careful before changing it, but we might want to have a different theme colour for disabled text within DetailView components.

CarlosNZ:
> It might be better if the three panels stayed in one row, but the whole row moved under the Name/Code block. Not sure what the best option might be.

Actually, just had another thought about this, since I just realised we have the ID in the Breadcrumb, which is probably not ideal. If we changed the breadcrumb ID to be the name, then put the code on the same line further to the right then we could have all three stats panels underneath taking the full width:

Quick mockup showing 3 stats panels fitting on more "realistic" width screen:

Screen Shot 2022-03-29 at 4 50 17 PM

packages/system/src/Item/DetailView/Tabs/MasterLists.tsx
mark-prins: yeah, reasonably common, though you're right - there will be items on only one list. mSupply currently has the the list of master lists though..

packages/common/src/ui/forms/Detail/Section.tsx
mark-prins: yeah. we did have an issue for font sizes; and maybe should use relative sizes than pixels

packages/common/src/ui/forms/Detail/InputWithLabelRow.tsx
mark-prins: I did it deliberately so that it _could_ be overridden by a consumer 😆

packages/system/src/Item/DetailView/Tabs/General.tsx
mark-prins: that is a good idea - not sure how to handle different input components, data validation or grouping of items though. that would require some mapping, like a configuration object and then it isn't completely dynamic.

I'm still favouring the declarative syntax for component use, so would keep this rather longhand version. We have planned to use JSON schema forms and this is great use case for that.

mark-prins:
> Quick mockup showing 3 stats panels fitting on more "realistic" width screen:

that's looking sweet! 😁 

We'd have to change when items are editable though, as the name and code need to be input components 🤔 
There's also no guarantee of 3 stats panels - likely that we have to cater for more, in which case Richard's second design might need to be used, or have a scrollable area in the AppBar.

I do like the idea of switching the id and the item name in the breadcrumb.. the prominence of the ID is kinda weird

mark-prins:
just opened this up on my laptop screen and the stats panels are ridiculous.
Have added the item name, moved name & code out of toolbar and removed the flex-wrap on the stats panels.
Looking better now.. though I have left the original grey on the stats panels. I think it was ok

PR #1125, opened 2022-03-29
Printing updates

### LINKED ISSUE #1124: Printing updates
Some changes to the API need to be brought through:
 
example 1:

Screen Shot 2022-03-29 at 12 25 52 PM

example 2:

```
{
    "data": null,
    "errors": [
        {
            "message": "Invalid value for argument \"filter.category.equalTo\", enumeration type \"ReportCategory\" does not contain the value \"INBOUND_SHIPMENT\"",
            "locations": [
                {
                    "line": 2,
                    "column": 62
                }
            ]
        }
    ]
}
```

mark-prins:
Fixes #1124 

enables pdf generation!

There's a small issue - in that the server sets the `Content-Disposition` header to `attachment` which forces the browser to download the file rather than open in a new tab. I've asked for that to be changed, as that is even less user friendly when trying to print.

As a small extra I've added a warning message when no reports exist.


packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/hooks/useNextItem.test.tsx
```diff
@@ -69,6 +69,20 @@ const getLines = (): InvoiceLineNode[] => [
         availableMonthsOfStockOnHand: 0,
         averageMonthlyConsumption: 0,
       },
+      atcCategory: '',
```
mark-prins, line 72: also in #1123 but I had to add it in for this change to compile

packages/system/src/Report/api/hooks.ts
```diff
@@ -29,7 +29,7 @@ const useReportApi = () => {
 export const useReports = (category?: ReportCategory) => {
   const api = useReportApi();
   const initialFilterBy = category
-    ? { category: { equalTo: ReportCategory.InboundShipment } }
+    ? { category: { equalTo: category } }
```
mark-prins, line 32: oopsie!

packages/system/src/Report/api/operations.graphql
```diff
@@ -34,19 +34,12 @@ query printReport($storeId: String!, $dataId: String!, $reportId: String!) {
...
       error {
-        description
-        ... on RecordNotFound {
-          __typename
-          description
-        }
         ... on FailedToFetchReportData {
           __typename
           description
+          errors
```
mark-prins, line 40: the error shape had changed

CarlosNZ:
> There's a small issue - in that the server sets the `Content-Disposition` header to `attachment` which forces the browser to download the file rather than open in a new tab. I've asked for that to be changed, as that is even less user friendly when trying to print.

Ah -- I wouldn't have noticed as Firefox chucks it in a new tab regardless :)

CarlosNZ (approved):
Cool, working nicely.

Just a couple of comments, not sure they can be dealt with here, but you may want to create issues for them if there isn't already:

1) There are "Print" buttons on pages that probably shouldn't have them, such as the ListViews. When you click them, they give a "No printer detected" message, which is probably not what we want. For now, do you want to just change that message to "Not available" or something? (If not removing the buttons)
2) The stocktake report doesn't work -- the query returns `DocGenerationError(    "Failed to render 'template'",)`. Presumably this is a back-end problem? Just thought I'd mention it in case it's something you can deal with here.

mark-prins:
Thanks - didn't check stocktake report, clearly 😁  mea culpa on that one - it was using a requisition report and passing in a stocktake as the item. I've fixed that - and have added an error notification if something like that happens again.

The print buttons on the list views.. I'll remove them. We don't have any way to print those yet..

PR #1129, opened 2022-03-29
#1128 Row hover fix

### LINKED ISSUE #1128: Disable row hover effect on Edit modal tables
As mentioned in [this comment](https://github.com/openmsupply/openmsupply-client/pull/1119#issuecomment-1081231646).

By default there is a onHover effect for table rows. It seems wrong when used in Edit modals, so we should disable it here
Screen Shot 2022-03-29 at 3 16 03 PM
.

CarlosNZ:
Fix #1128 

This crept in when I was updating the row hover styles the other day. Didn't realise I the hover style also over-rode the effect of the `hover` prop in `<TableRow>`.

Have removed the `hover` prop completely and made the hover style conditional.


mark-prins (approved):
👍  ah, right. hadn't realised that the hover style overrides the prop.

PR #1135, opened 2022-03-30
Prevent duplicate alert modals

### LINKED ISSUE #1134: AuthenticationAlert hidden
If you are on a detail page and the session times out, you see this:

Screen Shot 2022-03-30 at 8 19 02 AM

or even this:

Screen Shot 2022-03-30 at 9 08 01 AM

but I think you should see this:

Screen Shot 2022-03-30 at 8 19 31 AM

mark-prins:
Fixes #1134 

Well, that turned out to be a good one to sort out. The `AlertModal` is used in multiple places, at different levels in the document hierarchy. Turns out, the provider wasn't adding anything and multiple instances of the modal were being created 🙄 

A simple test.. In the `Internal Orders` detail page add

```
const [error, setError] = useLocalStorage('/auth/error');
```

and 

```
  React.useEffect(() => {
    setTimeout(() => {
      if (!error) setError(AuthError.Unauthenticated);
    }, 1000);
  }, []);
```

and then change the requisition number in the URL to an invalid one for your store.
you should get two modals, overlayed, with the 'requisition not found' appearing on top.

This PR fixes that.

### DIFF
--- a/packages/common/src/ui/components/modals/AlertModal/AlertModal.tsx
+++ b/packages/common/src/ui/components/modals/AlertModal/AlertModal.tsx
@@ -1,44 +1,38 @@
-import React from 'react';
-import { Grid, Typography } from '@mui/material';
-import { DialogButton } from '../../buttons';
-import { AlertIcon } from '@common/icons';
-import { BasicModal } from '@common/components';
-
+import React, { useContext } from 'react';
+import { AlertModalContext } from './AlertModalContext';
 export interface AlertModalProps {
+  important?: boolean;
   message: React.ReactNode;
   open: boolean;
   onOk: () => void;
   title: string;
 }
 
+// This is really just a convenience component
+// allowing you to use the modal in a declarative syntax
+// without creating multiple overlaying modals
+// Set the important prop only if this is a critical alert
+// which should only be superceded by other critical alerts
+// Apart from the important / non-important distinction,
+// the latest caller wins, and will be displayed
 export const AlertModal: React.FC<AlertModalProps> = ({
+  important,
   message,
   onOk,
   open,
   title,
 }) => {
-  return (
-    <BasicModal open={open} width={400} height={150}>
-      <Grid padding={4} container gap={1} flexDirection="column">
-        <Grid container gap={1}>
-          <Grid item>
-            <AlertIcon color="primary" />
-          </Grid>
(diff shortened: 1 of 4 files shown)


packages/common/src/ui/components/modals/AlertModal/AlertModal.tsx
```diff
@@ -1,44 +1,38 @@
...
+import { AlertModalContext } from './AlertModalContext';
 export interface AlertModalProps {
+  important?: boolean;
   message: React.ReactNode;
   open: boolean;
   onOk: () => void;
   title: string;
 }
 
+// This is really just a convenience component
```
mark-prins, line 11: like it says.. you could just use the context and do this stuff yourself - but I wanted to prevent people from falling into the same trap that we had.

Doing it like this instantiates only one modal and allows callers to manipulate the state declaratively, which has the added benefit of meaning that I don't have to change all the current usages 😁

packages/common/src/ui/components/modals/AlertModal/AlertModalContext.tsx
```diff
@@ -1,28 +1,32 @@
-import { createContext } from 'react';
+import React, { createContext } from 'react';
 
 export interface AlertModalState {
   open: boolean;
-  message: string;
+  message: string | React.ReactNode;
```
mark-prins, line 5: was already mixed mode, I've just updated the state type

packages/common/src/ui/components/modals/AlertModal/AlertModalProvider.tsx
```diff
@@ -18,12 +53,14 @@ export const AlertModalProvider: FC = ({ children }) => {
...
     () => ({
       setState,
-      setMessage: (message: string) =>
+      setMessage: (message: string | React.ReactNode) =>
         setState(state => ({ ...state, message })),
       setTitle: (title: string) => setState(state => ({ ...state, title })),
       setOnOk: () => {},
       setOpen: (open: boolean) => setState(state => ({ ...state, open })),
       ...alertModalState,
+      setImportant: (important: boolean) =>
```
mark-prins, line 62: like using `important!` in css 🤷

CarlosNZ (approved):
LGTM!

Thx.

PR #1136, opened 2022-03-30
Remove store select from login

### LINKED ISSUE #1089: Stocktakes: select all items

mark-prins:
Fixes #1089 
Does what it says on the tin.
Uses most recently used store, or picks the first one from the (alphabetically sorted) list.

Should probably have waited for the API changes, but what's a little rework amongst friends? 🤷

### DIFF
--- a/packages/common/src/authentication/AuthContext.tsx
+++ b/packages/common/src/authentication/AuthContext.tsx
@@ -126,7 +126,7 @@ export const AuthProvider: FC = ({ children }) => {
   // or the first store in the list
   // TODO: return default store rather than first - currently not provided by API
   const getStore = async () => {
-    if (!!mostRecentlyUsedCredentials && !!mostRecentlyUsedCredentials.store)
+    if (mostRecentlyUsedCredentials?.store)
       return mostRecentlyUsedCredentials.store;
 
     const { nodes } = await getStores();
--- a/packages/common/src/authentication/AuthContext.tsx
+++ b/packages/common/src/authentication/AuthContext.tsx
@@ -6,6 +6,7 @@ import { addMinutes } from 'date-fns';
 import { useGql } from '../api';
 import { useGetRefreshToken } from './api/hooks';
 import { useGetAuthToken } from './api/hooks/useGetAuthToken';
+import { useStores } from './api/hooks/useStores';
 import { AuthenticationResponse } from './api';
 
 export const COOKIE_LIFETIME_MINUTES = 60;
@@ -44,8 +45,7 @@ interface AuthControl {
   isLoggingIn: boolean;
   login: (
     username: string,
-    password: string,
-    store?: Store
+    password: string
   ) => Promise<AuthenticationResponse>;
   logout: () => void;
   mostRecentlyUsedCredentials?: MRUCredentials | null;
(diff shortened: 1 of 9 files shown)


packages/common/src/authentication/AuthContext.tsx
```diff
@@ -121,8 +122,22 @@ export const AuthProvider: FC = ({ children }) => {
   };
   useRefreshingAuth(saveToken, cookie?.token);
 
-  const login = async (username: string, password: string, store?: Store) => {
+  // returns MRU store, if set
+  // or the first store in the list
+  // TODO: return default store rather than first - currently not provided by API
+  const getStore = async () => {
+    if (!!mostRecentlyUsedCredentials && !!mostRecentlyUsedCredentials.store)
```
CarlosNZ, line 129:
```suggestion
    if (mostRecentlyUsedCredentials?.store)
```
Same thing?

packages/common/src/authentication/api/hooks/useStores.ts
```diff
@@ -0,0 +1,13 @@
+import { useMutation } from 'react-query';
+import { useQueryParams } from '@common/hooks';
+import { useAuthApi } from './useAuthApi';
+
+export const useStores = () => {
+  const api = useAuthApi();
+  const initialListParameters = { initialSortBy: { key: 'code' } };
+  const { filterBy, first, offset } = useQueryParams(initialListParameters);
+  const getStores = api.get.stores({ filterBy, first, offset });
+  const { mutate, ...rest } = useMutation(getStores);
```
CarlosNZ, line 10: You can explain this to me in person next week if that's easier, but I'm just wondering why this is a mutation?

CarlosNZ (approved):
Great, works nicely.

Just one thing I wondered -- when you log in the store label is quite subtle at the bottom of the page, so I could imagine someone logging in and not realising they're in a different store to what they thought.

Do you think it might be worth having a SnackBar notification as soon as you log in saying "Welcome to XXXX store" or something so it would alert the user clearly to which store they've logged in to? I dunno, might be a silly distraction, see what you think.

packages/common/src/authentication/AuthContext.tsx
mark-prins: yeah, thanks

packages/common/src/authentication/api/hooks/useStores.ts
mark-prins: using it so that I could called `mutateAsync` and `await` the response, and so that I could call explicitly.

Mutations allow you to call them when required, as opposed to queries which are called when you call `useQuery`, so it would run when the page loads. In this case, the request only needs to be called if there is no MRU, in most cases it is not required.

You can prevent / delay execution with a query by setting `enabled`, so in this case I could call 

``` useQuery(getStores, { enabled: !mostRecentlyUsedCredentials?.store }) ```

I thought that a mutation is clearer, and it is also helpful to be able to await. Given how quick the query is, there's little need to prefetch the results

mark-prins:
> Do you think it might be worth having a SnackBar notification as soon as you log in saying "Welcome to XXXX store" or something so it would alert the user clearly to which store they've logged in to? I dunno, might be a silly distraction, see what you think.

I remembered this morning about this - was a good idea and I missed it when merging. Have created #1146

PR #1138, opened 2022-03-31
#1130 Name details modal

### LINKED ISSUE #1130: Supplier and Customer detail: mock data implementation
With some mock data available we can put in a placeholder for the name detail pages.
There's no design mockup, so just make something that looks ok.

Here's what mSupply looks like, for reference:

Screen Shot 2022-03-30 at 8 14 18 AM

The item detail page is the closest we have - though for names we don't have stats and so is a simpler page. Perhaps for this one, just a detail modal rather than a full page. At least for now while there aren't many fields available and the fields are all read-only

CarlosNZ:
Fix #1130 

Also updated `DetailInputWithLabelRow` component to show regular text instead of input fields when `disabled`, as discussed. See what you think.

Overall I think the Modal is looking a little bland -- could do with a splash of colour or something, not sure. Although perhaps best not to try to pretty it up too much until we make them editable?
Screen Shot 2022-04-01 at 9 37 32 AM


packages/common/src/hooks/useDialog/useDialog.tsx
```diff
@@ -24,6 +25,7 @@ export interface ModalProps {
 
   okButton?: JSX.Element;
   width?: number;
+  sx?: SxProps<Theme>;
```
CarlosNZ, line 28: I'm using `sx` to set the maxWidth (which is 600px by default).

packages/common/src/utils/formatters/formatters.ts
```diff
@@ -8,6 +8,7 @@ export const Formatter = {
     `${withParens ? '(' : ''}${(tax ?? 0).toFixed(2)}%${withParens ? ')' : ''}`,
   date: (date: Date): string =>
     `${date.getDate()}/${date.getMonth() + 1}/${date.getFullYear()}`,
+  localisedDate: (date: Date): string => format(date, 'P'),
```
CarlosNZ, line 11: I couldn't see a localised date function anywhere else, so this would probably be useful?

mark-prins (approved):
😁  Awesome!!

I was concerned how the text would look, as distinct from a disable input - thinking that it wouldn't look like a form, and that might be a problem. But having seen it, I agree.. not being a form is better!

I've added a few comments.. just trying to be helpful really. Looks great and is fantastic to have the functionality in place.

packages/common/src/hooks/useDialog/useDialog.tsx
```diff
@@ -127,6 +130,7 @@ export const useDialog = (dialogProps?: DialogProps): DialogState => {
         onClose={handleClose}
         width={width}
         height={height}
+        sx={{ ...sx }}
```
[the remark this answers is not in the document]
CarlosNZ: Ah yes, of course. Think I was getting a little lost with the `sx` being passed around all over the place in this component 😄

packages/common/src/hooks/useDialog/useDialog.tsx
CarlosNZ: Cool, I've added the Transition as suggested, looks good.

And yeah, I'm always reluctant to add fixed widths to elements where I can avoid it ;)

packages/common/src/utils/formatters/formatters.ts
CarlosNZ: There's a `useFormatDate` hook in there, but it seems to be returning US-style dates instead of localised ones. You can pass a DateTimeFormatOptions object to it to specify formatting, but we want the date to automatically show in the user's local format regardless of where they are. It seems the `date-fns` functions is the only way to (easily) auto-format in local date format. 

Perhaps the `useFormatDate` function needs to be updated to use this? I'll make a new issue to figure it out and we can address it next week?

packages/system/src/Name/DetailModal/DetailModal.tsx
```diff
@@ -0,0 +1,112 @@
...
+  DetailInputWithLabelRow,
+  DetailSection,
+  Checkbox,
+  Grid,
+  Formatter,
+  Typography,
+  Box,
+  BasicSpinner,
+} from '@openmsupply-client/common';
+import { Link } from '@mui/material';
```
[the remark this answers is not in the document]
CarlosNZ: Yup, gotcha. I actually tried to use the React Router `Link` first, but I could get it to handle external links properly (it would always add the link to the local host root 🙄)

Anyway, `MuiLink` in "common" now.

packages/common/src/utils/formatters/formatters.ts
CarlosNZ: New issue: https://github.com/openmsupply/openmsupply-client/issues/1145 
Let's discuss next week.

packages/common/src/ui/forms/Detail/InputWithLabelRow.tsx
```diff
@@ -17,25 +19,28 @@ interface InputWithLabelRowProps {
 export const DetailInputWithLabelRow: FC<InputWithLabelRowProps> = ({
   label,
   inputProps,
-  Input = <BasicTextInput style={{ width: '100%' }} {...inputProps} />,
+  Input = <BasicTextInput {...inputProps} />,
+  DisabledInput = <Typography>{inputProps?.value as string}</Typography>,
```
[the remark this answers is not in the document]
CarlosNZ: Yep, I hear you. I've made another little mini-component `<NumericTextDisplay>` (to be the "disabled" partner of `<NumericTextInput>`) which just right-aligns the content in a fixed-width space, so you get:

Screen Shot 2022-04-01 at 9 13 36 PM


Here's another little mystery though: you'll notice in that image, the gap above "Volume per outer pack" is larger than the gap above "Volume per pack". This is completely inexplicable: the components are exactly the same, and neither of the label or text are close to their available width. In the Inspector I can see absolutely nothing different in terms of CSS values, one just gets a height of 40 and the other remains the height of its children (~24). It's something to do with the label length though -- reduce it by a couple of characters and the height'll match the narrower ones. Very strange behaviour... 🤷‍♂️

CarlosNZ:
Thanks for the suggestions, @mark-prins -- have updated accordingly, see last 2 commits and comments.

PR #1143, opened 2022-04-01
Default error handling for queries

### LINKED ISSUE #1101: Give better feedback about database errors

mark-prins:
Fixes #1101 

Adds a generic toast on query and mutation errors. I think mutations should be handled in each specific case and will require tailored messaging. Even so, a default catch-all could be helpful.


packages/common/src/intl/locales/en/common.json
```diff
@@ -47,6 +47,7 @@
   "description.stock-on-hand": "Stock on Hand (Est. remaining)",
   "description.supply-quantity": "Amount to Supply",
   "error.404": "Click the back arrow or use the menu to find your way again.",
+  "error.general-query-error": "There was an communicating with the server 😢",
```
mark-prins, line 50: ok, perhaps the emoji isn't very professional. But 🤷

packages/host/src/QueryErrorHandler.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { useQueryClient } from 'react-query';
+
+export const QueryErrorHandler = () => {
+  const client = useQueryClient();
+  const { error } = useNotification();
+  const t = useTranslation();
+
+  const generalError = t('error.general-query-error');
+
+  useEffect(() => {
```
mark-prins, line 13: this is because I want to use toast notifications.. and the notifications are using hooks which I cannot access outside of a component or hook. Otherwise this would just be at the top of `Host`
It also means that I can `useTranslation`, though I expect we won't ever hit the general error message anyway..

The toasting also requires being under the `SnackbarProvider` so I had to add it lower down the stack than just in `Host`

CarlosNZ (commented):
Looks good, but can't actually find a way to test it -- how do you simulate these errors?

Also, we don't seem to be capturing connection errors. When I set the server to something incorrect in "config.js", the queries all just show "No records to display" and don't give any errors in the UI.

One other test I tried. Turned my Wi-Fi off then clicked on a record in a table. Loading skeleton appeared and stayed there indefinitely. When I restored my connection, after a few more seconds, this came up:
Screen Shot 2022-04-01 at 10 01 34 PM

Is that what's expected?

mark-prins:
> Looks good, but can't actually find a way to test it -- how do you simulate these errors?

One that I was using when testing this was in `useRequestLineChartData`, comment out the `onError` and go to an internal order detail page, then click on an item. They were all throwing errors for me

mark-prins:
Ok - Have made a bunch of changes which should address all of the issues.
@CarlosNZ  would you want to have another look through?

The main issue I had was tracking down the problem with Internal Orders - finally found why that was endlessly rendering!
Problem there was with the hooks - the `useRequests` hook was being called by the `ListView` and the `Toolbar` within it; when the query errored the state changed for both which rendered them again, which queried again..

To resolve, I've changed the toolbar usage to query the cached version only by setting enabled to false. that way only the page query fails and everything is more stable.

packages/common/src/api/GqlContext.tsx
```diff
@@ -58,7 +58,7 @@ class GQLClient extends GraphQLClient {
             throw new Error(details || error.message);
           }
         }
-        return this.emptyData;
+        throw new Error('Unknown error');
```
mark-prins, line 61: required to set the error state in react-query

packages/common/src/authentication/AuthContext.tsx
```diff
@@ -129,7 +129,7 @@ export const AuthProvider: FC = ({ children }) => {
     if (mostRecentlyUsedCredentials?.store)
       return mostRecentlyUsedCredentials.store;
 
-    const { nodes } = await getStores();
+    const { nodes } = (await getStores()) || {};
```
mark-prins, line 132: this was causing a secondary error when the stores couldn't be fetched

packages/common/src/authentication/api/api.ts
```diff
@@ -63,11 +65,18 @@ export const getAuthQueries = (sdk: Sdk) => ({
...
-        password,
-      });
-      return authTokenGuard(result);
+      try {
+        const result = await sdk.authToken({
+          username,
+          password,
+        });
+        return authTokenGuard(result);
+      } catch (e) {
```
mark-prins, line 74: login page needs a special case.

packages/common/src/ui/layout/tables/DataTable.tsx
```diff
@@ -77,19 +77,17 @@ export const DataTableComponent = <T extends RecordWithId>({
...
-      <Box
-        sx={{
-          display: 'flex',
-          flex: 1,
-          justifyContent: 'center',
-          alignItems: 'center',
-        }}
-      >
-        <CircularProgress />
+      <Box sx={{ padding: 2 }}>
```
mark-prins, line 84: passing the error down here, and letting the data table do the error display. much neater than repeating this on a bunch of pages

packages/common/src/ui/layout/tables/types.ts
```diff
@@ -15,15 +15,16 @@ export interface QueryResponse<T> {
...
 
 export interface TableProps<T extends RecordWithId> {
-  isDisabled?: boolean;
+  children?: ReactNode;
   columns: Column<T>[];
   data?: T[];
+  dense?: boolean;
+  ExpandContent?: FC<{ rowData: T }>;
+  isDisabled?: boolean;
+  isError?: boolean;
```
mark-prins, line 24: adding in `isError` and alphabetising while I'm at it..

packages/host/src/QueryErrorHandler.tsx
```diff
@@ -0,0 +1,38 @@
+import React, { useEffect, useState } from 'react';
+import { useNotification } from '@common/hooks';
+import { useTranslation } from '@common/intl';
+import { useQueryClient } from 'react-query';
+
+export const QueryErrorHandler = () => {
+  const client = useQueryClient();
+  const { error } = useNotification();
+  const t = useTranslation();
+  const [errorMessage, setErrorMessage] = useState<string | null>(null);
```
mark-prins, line 10: Adding in the state so that a specific error is only toasted once 😁

packages/requisitions/src/RequestRequisition/api/hooks.ts
```diff
@@ -225,19 +228,16 @@ export const useDeleteRequests = () => {
 };
 
 export const useDeleteSelectedRequisitions = () => {
-  const { data: rows } = useRequests();
+  const { data: rows } = useRequests({ enabled: false });
```
mark-prins, line 231: this gave me so much grief!

mark-prins:
This is how the data table looks when there's an error:

Screen Shot 2022-04-04 at 4 10 26 PM

CarlosNZ (approved):
Okay, great, seems good. Thanks.

> One that I was using when testing this was in useRequestLineChartData, comment out the onError and go to an internal order detail page, then click on an item. They were all throwing errors for me

One little thing when I do this. When clicking on a line, I get the Toast notification for the first one, but when I click on other lines in the same table (including the same one again), there's no more notification. Have to refresh the page for the Toast to start showing up again. Does some state value need resetting after the Toast?

mark-prins:
> One little thing when I do this

That was deliberate, to suppress multiple - even when returning to the page. Could set cache timeout and have it toast each time you visit, I just figured that it could be annoying seeing the same message every time. Will see how it looks though..

mark-prins:
Added a location aware re-toast 😉 
8fe2490a

Issue #1140, opened 2022-03-31
Table column widths


mark-prins:
Some issues with column width specification

- percentage values have no effect for `min-width` or `max-width`
- `max-width` has no effect on the first column

mark-prins:
the percentage values still don't work.. but I don't really care any more 🤷‍♂️ 
The `max-width` is likely due to `NameAndColorColumn` having a `minWidth: 400` on it 😁

PR #1144, opened 2022-04-01
Usability issues

### LINKED ISSUE #1132: Usability Issues
@richardmoizeau has picked up a few things while creating documentation. There will be more coming.. this is just a place to pop in the ones found so far and to knock off a few at a time rather than creating a single issue per item

- [x] `Cancel` buttons - in the detail views the 'cancel' doesn't actually cancel it just returns you to the list. The name should be 'Close' or 'Return to List' or 🤷 
- [x] Shipments > it is not clear that you should enter a % for tax rate
- [x] Shipments > List: Invnum should be replaced by Shipment ID or Shipment Number
- [x] Shipments > List: Customer column is too wide
- [x] Shipment > List: Customer to be replaced by Customer Name
- [x] Shipment > List: Entered should be replaced by Entered on or Created on 
- [x] Shipments > List: Total should be replaced by Total Value
- [x] Shipments > List: add Customer Ref
- [ ] Shipment > List: would it be possible to have the below item always at the bottom of the page ?
Screen Shot 2022-03-30 at 2 28 26 PM
- [x] Shipment > List: cannot navigate to another page if I'm not at the bottom of the list (using vertical scroll bar). It seems that the issue is only with the 'last page'. When clicking on it, it is activating the main check box
- [x] Shipment: no warning dialog when deleting selected lines (bulk actions)
- [ ] O/Shipment: created a line for an item with no stock (placeholder). When I use the "Allocate Placeholder Lines" function, notification says 1 line has been allocated but in fact... no :) Suggestion for notif: 
```
Allocated X line (when line has been fully allocated)
Partially allocated X line (when line has been partially allocated)
X lines not allocated 
```
- [ ] O/Shipment: when I try to confirm allocation with unallocated line, I get the dialog. When I click on `OK`, I'm redirected to the login page
- [ ] when i try to delete a line when status is "shipped", notif says "no you can't unless it's new" but actually you can delete line when status is allocated or picked. Suggestion: You cannot delete a line when status is shipped. 
- [ ]

mark-prins:
Fixes #1132 (mostly)

Have tidied up column widths, added column descriptions and fixed the checkbox-goes-over-the-pagination problem.

The one that is still bugging me isn't raised in the issue, which is what happens to pagination when the table is wider than the viewable screen:

Screen Shot 2022-04-01 at 11 18 32 AM

If you remove the `width: 100%` on the `TableProvider` it is fine, but means that tables narrower than the viewable screen don't stretch. Added a `minWidth` instead still causes the problem. Have left it for now..


packages/common/src/ui/layout/tables/DataTable.tsx
```diff
@@ -108,6 +108,7 @@ export const DataTableComponent = <T extends RecordWithId>({
         flexDirection: 'column',
         height: '100%',
         overflowX: 'unset',
+        tableLayout: 'fixed',
```
mark-prins, line 111: a possible fix for the percentage column widths problem (#1140) which didn't actually help.

packages/common/src/ui/layout/tables/DataTable.tsx
```diff
@@ -156,6 +157,7 @@ export const DataTableComponent = <T extends RecordWithId>({
           insetBlockEnd: 0,
           backgroundColor: 'white',
           justifyContent: 'flex-end',
+          zIndex: 100,
```
mark-prins, line 160: fixes the checkbox-over-pagination problem

packages/common/src/ui/layout/tables/utils/ColumnDefinitionSetBuilder.ts
```diff
@@ -101,6 +102,7 @@ const getColumnLookup = <T extends RecordWithId>(): Record<
   },
   invoiceNumber: {
     key: 'invoiceNumber',
+    align: ColumnAlign.Right,
```
mark-prins, line 105: this showed up a problem with the width of the column - the header wasn't in the right place.

packages/invoices/src/InboundShipment/ListView/ListView.tsx
```diff
@@ -54,7 +54,7 @@ export const InboundListView: FC = () => {
       ['invoiceNumber', { maxWidth: 80 }],
       'createdDatetime',
       'allocatedDatetime',
-      ['comment', { maxWidth: 300 }],
+      ['comment', { width: '100%' }],
```
mark-prins, line 57: doing this to squash the name column, without it, the name spreads too much on a wide screen

packages/common/src/ui/layout/tables/DataTable.tsx
CarlosNZ: In the CSS inspector, it informs me: `table-layout has no effect on this element since it’s not a table.`

It would need to be on the `<MuiTable>` component to have an effect, you've got it on the `TableContainer` component (which is just a `div`).

But personally, I'm not so in favour of fixed column widths anyway -- I think it's nicer to let content dictate widths and just set maximums.

packages/invoices/src/InboundShipment/ListView/ListView.tsx
CarlosNZ: Why not just put a max-width on the "name" column?

CarlosNZ (commented):
Couple of inline comments.

Also:
> percentage values have no effect for min-width or max-width
> max-width has no effect on the first column

Where are you seeing this problem, specifically? I tried a couple and the ones I tried all seemed to respect percentage min/max, and could apply a max-width to first column.

packages/common/src/ui/layout/tables/DataTable.tsx
mark-prins: Doh! and thanks - have removed it.

packages/invoices/src/InboundShipment/ListView/ListView.tsx
mark-prins: Some of the names used by clients are really long - and many don't have comments. Basically just trying to cater for long names when they exist

and it didn't work for me - setting a `max-width` on name doesn't change anything unless I also add the `100%` width to comment

mark-prins:
> Where are you seeing this problem, specifically?

I'm using chrome by default - though the max-width isn't working in firefox either:

Screen Shot 2022-04-02 at 12 32 59 PM

**Firefox**
Screen Shot 2022-04-02 at 12 33 05 PM

**Chrome**
Screen Shot 2022-04-02 at 12 33 18 PM

Issue #1150, opened 2022-04-04
Determine locale by system setting


CarlosNZ:
This will need some investigation to see whether it's worth it. 

~~Consider using this package: https://www.npmjs.com/package/os-locale~~
Actually, this one seems more promising: https://www.npmjs.com/package/get-user-locale
Or, more specific to our i18n library: https://www.npmjs.com/package/i18next-browser-languagedetector

If it's reliable, then we should also try and use the os locale as the default language (rather than just assuming English)

jmbrunskill:
Don't you already have the locale from i18next-browser-languagedetector?

This blog seems to imply you could use the detected language with the JS built in Intl.DateTimeFormat function
https://phrase.com/blog/posts/localizing-react-apps-with-i18next/#Date_Formatting

See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat

Not sure if that helps (haven't tried any of this!)

CarlosNZ:
Thanks @jmbrunskill, yes we should be able to use i18next-browser-languagedetector. Will add it to the original post.

Cheers.

PR #1151, opened 2022-04-04
#1145 Date formatters

### LINKED ISSUE #1145: Date formatters -- tidy up
As per [this comment thread](https://github.com/openmsupply/openmsupply-client/pull/1138#discussion_r840111999)

The current `useFormatDate` seems to be returning US-style dates -- we need something that can localise the date display regardless of where it is used.

Also -- I'm not sure why `useFormatDate` needs to be a "hook" -- it doesn't actually use any React state or other hooks, it just returns a plain JS function.

My suggestion would be to remove `useFormatDate` and consolidate date formatters into the existing "Formatters" module, specifically taking advantage of `date-fns` locale-aware functionality, although not sure the best file/function structure -- let's discuss!

CarlosNZ:
Fix #1145

Large number of file changes, but most are just changing existing implementations to use the new localised hook `useFormatDateTime` (in DateUtils).

Deprecated `useFormatDate`

Quite a bit of fiddling around for not much actual change in the end, but at least the `useFormatDateTime` hook will  be easy to add additional localised date/time formatters to going forward.

Was planning to just make utility functions and not create a hook, but we need to access `useCurrentLanguage` so needs to be a hook.

For now, it just users browser language to determine whether to use US or GB for English, but I've made a new issue to investigate this further: #1150


packages/common/src/ui/layout/tables/columns/types.ts
```diff
@@ -39,7 +39,7 @@ export type ColumnDataAccessor<T extends RecordWithId> = (params: {
 
 export type Translators = {
   t: ReturnType<typeof useTranslation>;
-  d: ReturnType<typeof useFormatDate>;
+  d: (date: string | number | Date) => string;
```
CarlosNZ, line 42: Could be used for several different date formatting methods (though only 2 currently), but all should follow this signature

packages/common/src/utils/formatters/formatters.ts
```diff
@@ -6,9 +6,6 @@ export const Formatter = {
   // TODO: Do something better than naively rounding to 2 decimal places
   tax: (tax: number, withParens = true) =>
     `${withParens ? '(' : ''}${(tax ?? 0).toFixed(2)}%${withParens ? ')' : ''}`,
-  date: (date: Date): string =>
-    `${date.getDate()}/${date.getMonth() + 1}/${date.getFullYear()}`,
-  localisedDate: (date: Date): string => format(date, 'P'),
```
CarlosNZ, line 11: Never used, can't see a reason for it.

packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/ItemCharts/ConsumptionHistory.tsx
```diff
@@ -29,13 +29,9 @@ export const ConsumptionHistory: React.FC<ConsumptionHistoryProps> = ({
...
   const theme = useTheme();
-  const d = useFormatDate();
+  const { customDate } = useFormatDateTime();
   const { data, isLoading } = useRequestLineChartData(id);
-  const dateFormatter = (date: string) =>
-    d(new Date(date), {
-      month: 'short',
-      day: '2-digit',
-    });
+  const dateFormatter = (date: string) => customDate(date, 'dd MMM');
```
CarlosNZ, line 34: Does the same thing as what it replaces, just uses `date-fns` functions (localised) rather that Intl.DateTime

packages/common/src/intl/utils/DateUtils.ts
```diff
@@ -0,0 +1,69 @@
...
+  isToday,
+  isThisMonth,
+  isAfter,
+  isBefore,
+  isEqual,
+  format,
+  parseISO,
+  fromUnixTime,
+} from 'date-fns';
+import { enGB, enUS, fr, ar } from 'date-fns/locale';
```
CarlosNZ, line 16: Not thrilled about having to import these. Ideally we would lazy-load them as required, as part of the Intl Context (updating when "changeLanguage" is run), but not sure how to do this / if it's worth it.

packages/common/src/intl/utils/DateUtils.ts
```diff
@@ -0,0 +1,69 @@
...
+        : enGB
+      : getLocaleObj[language];
+
+  const localisedDate = (date: Date | string | number): string =>
+    format(dateInputHandler(date), 'P', { locale });
+
+  const customDate = (
+    date: Date | string | number,
+    formatString: string
+  ): string => format(dateInputHandler(date), formatString, { locale });
```
CarlosNZ, line 65: Strictly we can do all date formatting with the one method (customDate), but useful to create a handful of "preset" formats (added as required)

packages/common/src/intl/utils/DateUtils.ts
mark-prins: agree - not ideal. would be nice to lazy load if we can.. but that's a problem for another day!

packages/common/src/intl/utils/DateUtils.ts
mark-prins: yes, the 'preset' idea is important. Good to have something like the expiry date formatted consistently throughout, and easy to update if required

packages/common/src/utils/formatters/formatters.ts
mark-prins: I think a remnant from a time when the api was returning dates, or being parsed on fetch

packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/ItemCharts/ConsumptionHistory.tsx
mark-prins: potentially a candidate for a preset date format too?

mark-prins (approved):
🎉  I like that a lot - very nicely wrapping up a few problems with formatting and inconsistent usage 😁

packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/ItemCharts/ConsumptionHistory.tsx
CarlosNZ: Yup: 78394429

PR #1152, opened 2022-04-04
Round some numbers

### LINKED ISSUE #1147: Rounding issues
A couple that have come up:

**Internal order**

Screen Shot 2022-04-03 at 9 10 32 AM

**Item details**

IMAGE 2022-04-03 14:47:06

mark-prins:
Fixes #1147 

To test:

`Acetylsalicylic Acid 300mg tabs` or `Amoxicillin 250mg tabs` in Opua is giving some funky numbers.
Also in Opua - create an inbound shipment and add `Acetylsalicylic Acid 300mg tabs` - hopefully before I fix up some data on the server 😉  Currently it is giving 33 1/3 which does not look great as decimals..

Results:

Screen Shot 2022-04-04 at 5 02 42 PM

Screen Shot 2022-04-04 at 4 50 27 PM


CarlosNZ (commented):
I guess we should probably localise these as well, since different languages display them differently (for example, doesn't French use `,` for decimal place?)

Pretty sure we can do all number presentation (including rounding) using JS `Intl.NumberFormat` objects: [link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat)

It can do localised currency display too. 

If you want to merge this as it is for now, I can make a generalised "Number Formatter" object later on, with specific methods that output localised versions, basically a parallel to the DateFormatter methods. Up to you.

mark-prins:
thanks - added 87734b1b for the intl version 😉

PR #1155, opened 2022-04-04
Quick fix for API authentication changes

CarlosNZ:
Seems to be working and compiles :)

@mark-prins -- in "operations.generated.ts", the `AuthTokenQuery` type updated automatically when I ran the codegen, but `query authToken` in the same file I had to update manually. Is this what's supposed to happen?

Anyway, please modify this as you see fit and merge it as soon as you're happy, since we need to get this working again ASAP. 

You probably want to re-merge main into your Database Queries branch too once this is merged, as it looks like they may have some slight overlap?

### DIFF
--- a/packages/common/src/authentication/api/api.ts
+++ b/packages/common/src/authentication/api/api.ts
@@ -29,8 +29,7 @@ const authTokenGuard = (
 
   if (authTokenQuery?.authToken?.__typename === 'AuthTokenError') {
     switch (authTokenQuery.authToken.error.__typename) {
-      case 'DatabaseError':
-      case 'InternalError':
+      case 'InvalidCredentials':
         return {
           token: '',
           error: { message: authTokenQuery.authToken.error.description },
--- a/packages/common/src/authentication/api/api.ts
+++ b/packages/common/src/authentication/api/api.ts
@@ -32,7 +32,7 @@ const authTokenGuard = (
       case 'InvalidCredentials':
         return {
           token: '',
-          error: { message: authTokenQuery.authToken.error.description },
+          error: { message: '' },
         };
     }
   }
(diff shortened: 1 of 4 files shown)


mark-prins:
> I had to update manually. Is this what's supposed to happen?

No - that would mean the query in `operations.graphql` is incorrect. should never have to update the generated file manually

CarlosNZ:
> 



> > I had to update manually. Is this what's supposed to happen?
> 
> No - that would mean the query in `operations.graphql` is incorrect. should never have to update the generated file manually

Ah, right. So `operations.graphql` is the "master" query definitions. Have updated that file now, re-ran codegen -- works as exected: 2aad7a31

packages/common/src/authentication/api/api.ts
```diff
@@ -29,8 +29,7 @@ const authTokenGuard = (
 
   if (authTokenQuery?.authToken?.__typename === 'AuthTokenError') {
     switch (authTokenQuery.authToken.error.__typename) {
-      case 'DatabaseError':
-      case 'InternalError':
+      case 'InvalidCredentials':
```
mark-prins, line 32: This would return the message from the server, which is not the intention here. We need to have it translatable, which won't be provided by the server.

The original was catching systemic errors and showing those, and with other Auth errors ( e.g. InvalidCredentials ) showing the message from `app.json` : `error.login`

Given that we don't have access to `useTranslation` here, I'd recommend returning 
```
{
          token: '',
          error: { message: '' },
}
```

To return the original, translated string. Otherwise it shows this regardless of language:

Screen Shot 2022-04-05 at 10 36 54 AM

otherwise, a specific token to indicate there was an auth error. Or always show the invalid user - though it is helpful to also show if the server is down, which is what the last part will do.

CarlosNZ: Ah yes, good point. Thanks df398907

mark-prins (approved):
🎉  great! let's get it merged then 😁

PR #1156, opened 2022-04-05
#1146 store changes on login

### LINKED ISSUE #1146: Login: Store changes
Need to check and implement a few extra things
 - is the list of stores specific to a user? We'd talked about returning stores on the `/me` query
 - request default store from api, possibly also the `me` endpoint
 - Show the current store on login, as indicated in [this comment](https://github.com/openmsupply/openmsupply-client/pull/1136#pullrequestreview-928174935)

mark-prins:
Fixes #1146 

The list of stores is now user-specific. In addition:
- There's a toast notification on changing store / logging in which informs the user of the current store
- The store selector is using the store name which is now available, instead of store code
- The store selector has been tidied a little
- There's an alert modal if the user has no stores ( e.g. `kawakawa` which actually does in mSupply, but aren't coming through in API, so is a handy test )


packages/common/src/authentication/AuthContext.tsx
```diff
@@ -108,11 +104,12 @@ export const AuthProvider: FC = ({ children }) => {
   const i18n = IntlUtils.useI18N();
   const defaultLanguage = IntlUtils.useDefaultLanguage();
   const { mutateAsync, isLoading: isLoggingIn } = useGetAuthToken();
-  const { mutateAsync: getStores } = useStores();
   const authCookie = getAuthCookie();
   const [cookie, setCookie] = useState<AuthCookie | undefined>(authCookie);
   const [error, setError] = useLocalStorage('/auth/error');
   const storeId = cookie?.store?.id ?? '';
+  const { mutateAsync: getStores } = useUserDetails();
```
mark-prins, line 111: changed to a `useQuery` then back to a mutation - the auth header has to be set when querying, and loading the store list before login is a problem for that!

packages/common/src/authentication/AuthContext.tsx
```diff
@@ -124,20 +121,27 @@ export const AuthProvider: FC = ({ children }) => {
...
-  const getStore = async () => {
-    if (mostRecentlyUsedCredentials?.store)
+  const getStore = async (token?: string) => {
+    const userDetails = await getStores(token);
+    const defaultStore = userDetails?.defaultStore;
+    const stores = userDetails?.stores.nodes;
+
+    if (
+      mostRecentlyUsedCredentials?.store &&
+      stores?.some(store => store.id === mostRecentlyUsedCredentials?.store?.id)
```
mark-prins, line 131: don't log a user in to a MRU store if they don't have access to it

packages/common/src/authentication/AuthContext.tsx
```diff
@@ -124,20 +121,27 @@ export const AuthProvider: FC = ({ children }) => {
...
+    ) {
       return mostRecentlyUsedCredentials.store;
+    }
 
-    const { nodes } = await getStores();
-    return !!nodes && nodes?.length && nodes?.length > 0
-      ? nodes?.[0]
-      : undefined;
+    if (!!defaultStore) return defaultStore;
```
mark-prins, line 136: default store is now available! as a separate property though, rather than a prop on a store.. so I have to do it this way

packages/common/src/authentication/AuthContext.tsx
```diff
@@ -124,20 +121,27 @@ export const AuthProvider: FC = ({ children }) => {
...
+
+    return !!stores && stores?.length > 0 ? stores?.[0] : undefined;
   };
 
   const login = async (username: string, password: string) => {
     const { token, error } = await mutateAsync({ username, password });
-    const store = await getStore();
+    setHeader('Authorization', `Bearer ${token}`);
```
mark-prins, line 143: adding this as a catch for subsequent queries which need to use the new token and might be called before the cookie is updated

packages/common/src/authentication/api/api.ts
```diff
@@ -72,16 +64,18 @@ export const getAuthQueries = (sdk: Sdk) => ({
...
-          first,
-          offset,
-        });
-
-        return result.stores;
-      },
+    me: async (token?: string) => {
+      const result = await sdk.me(
+        {},
+        {
```
mark-prins, line 70: manually setting the header here, to ensure the correct token.
Waiting for the gql context to update state sometimes doesn't work, and this is more reliable

packages/host/src/components/Footer/StoreSelector.tsx
```diff
@@ -8,51 +8,64 @@ import {
...
       Content={
         <PaperPopoverSection label={t('select-store')}>
           {isLoading ? (
             <CircularProgress size={12} />
           ) : (
             <Box
-              style={{ maxHeight: '200px', overflowY: 'auto' }}
+              style={{
+                overflowX: 'hidden',
+                textOverflow: 'ellipsis',
```
mark-prins, line 65: oh. yeah - the overflow isn't working and I couldn't see how to make it work as I expected it to. gave up in the end 🤷

CarlosNZ: It works if you move the horizontal overflow style properties to the `<FlatButton>` component instead. So the `sx` property on `FlatButton` should be:
```
sx={{
        whiteSpace: 'nowrap',
        overflowX: 'hidden',
        textOverflow: 'ellipsis',
        display: 'block',
        textAlign: 'left',
      }}
```
`display: block` is just to override the default `display: inline-flex` which messes up the ellipsis.

packages/host/src/components/Footer/StoreSelector.tsx
```diff
@@ -8,51 +8,64 @@ import {
...
+                overflowY: 'auto',
+                maxHeight: POPPER_HEIGHT - 100,
```
CarlosNZ, line 67: `maxHeight` not actually doing anything cos there is a fixed `height` prop. It looks better if you remove the `height` prop and just let this `maxHeight` do its thing, IMHO.

CarlosNZ (approved):
Cool, working nicely. The welcome toast is very effective.

Just a couple of small style comments.

packages/host/src/components/Footer/StoreSelector.tsx
mark-prins: thanks. yeah, flex 🙄

packages/common/src/authentication/api/hooks/useUserDetails.ts
```diff
@@ -0,0 +1,15 @@
...
+  const api = useAuthApi();
+  return useMutation<UserNode, unknown, string | undefined, unknown>(
+    api.get.me
+  );
+};
+
+export const useUserStores = (token?: string) => {
+  const api = useAuthApi();
+  return useQuery(api.keys.me(token || ''), api.get.stores(), { cacheTime: 0 });
+};
```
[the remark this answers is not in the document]
mark-prins: the token is coming out of the auth cookie which could expire, and yes - you should be redirected to login when the cookie expires. might still slip between the cracks though.

I've added the check in AuthContext instead, and updated this hook, thanks.

packages/host/src/components/Footer/StoreSelector.tsx
mark-prins: yeah - removing flex tidies that up. have removed height now.

PR #1158, opened 2022-04-05
#1157 automate finding unused strings

### LINKED ISSUE #1157: Remove unused language strings (automated)
https://github.com/mxmvshnvsk/i18n-unused

CarlosNZ:
Fix #1157 

Had this one mostly sorted a while back, but had to [troubleshoot a problem](https://github.com/mxmvshnvsk/i18n-unused/issues/16) with the developer of the script.

Uses https://www.npmjs.com/package/i18n-unused -- package added to devDependencies.

Have added 3 yarn shortcuts in package/scripts:
- `i18n-unused-display`
- `i18n-unused-mark`
- `i18n-unused-remove`

Changes in language files here as a result of running `yarn i18n-unused-remove`.


packages/common/src/intl/locales/ar/common.json
```diff
@@ -1,6 +1,6 @@
...
-    "button.ok": "نعم",
-    "button.open-the-menu": "افتح القائمة",
-    "app.admin": "Admin",
-    "app.catalogue": "catalogue"
-}
+  "button.ok": "نعم",
+  "button.open-the-menu": "افتح القائمة",
+  "app.admin": "Admin",
+  "app.catalogue": "catalogue"
+}
```
CarlosNZ, line 6: Whitespace changes due to Prettier reformatting on save -- shouldn't change again.

i18n-unused.config.js
```diff
@@ -0,0 +1,19 @@
+// Info: https://www.npmjs.com/package/i18n-unused
+
+module.exports = {
+  localesPath: 'packages/common/src/intl/locales',
+  srcPath: 'packages',
+  ignorePaths: [
```
mark-prins, line 6: I don't think `ignorePaths` is necessary - in that it runs in the same amount of time without it. Tried with the glob syntax

```
  ignorePaths: ['**/node_modules'],
```

and then simply removed. Did you have an issue which required adding the prop in?

i18n-unused.config.js
```diff
@@ -0,0 +1,19 @@
...
+    'packages/common/node_modules',
+    'packages/android/node_modules',
+    'packages/config/node_modules',
+    'packages/dashboard/node_modules',
+    'packages/host/node_modules',
+    'packages/inventory/node_modules',
+    'packages/invoices/node_modules',
+    'packages/requisitions/node_modules',
+    'packages/system/node_modules',
+    'packages/template/node_modules',
```
mark-prins, line 16:
```suggestion
    'packages/common/node_modules',
    'packages/config/node_modules',
    'packages/dashboard/node_modules',
    'packages/host/node_modules',
    'packages/host/dist',
    'packages/inventory/node_modules',
    'packages/invoices/node_modules',
    'packages/requisitions/node_modules',
    'packages/system/node_modules',
    'packages/template/node_modules',
    'packages/android',
    'packages/android/node_modules',
    '../node_modules',
```

ok - so I tried running `i18n-unused display-missed` and that has highlighted the need for the ignore prop, and some bits which will need adding..
Still disappointed that we have to be explicit with all the paths, but I'll just have to live with disappointment.

packages/common/src/intl/locales/en/catalogue.json
```diff
@@ -31,4 +30,4 @@
   "title.packaging": "Packaging",
   "title.pricing": "Pricing",
   "title.stock-on-hand": "Stock on Hand"
-}
+}
```
mark-prins, line 33: I wonder if weblate will add the line break back in and cause a commit war? 🤔

mark-prins (approved):
🎉  that's awesome! Very useful - and something I'd wondered about 😁 

Would be even better if we could suppress the need to tell me that `Unused translations count: 0` for each file.. but hey.

When I run it now, I get `Total unused translations count: 87` - perhaps you need to run it again? Or am I doing something wrong? The unused are correctly detected, such as 'app.admin' in common.. which is now in `app` and just `admin`

package.json
```diff
@@ -14,7 +14,10 @@
     "build-storybook": "build-storybook -s ./packages/host/public ",
     "generate": "graphql-codegen --config codegen.yml",
     "android:build-remote_server": "cd ./packages/android && ./build_remote_server_libs.sh",
-    "android:build": "yarn build && npx cap copy && cd ./packages/android && ./gradlew assembleRelease"
+    "android:build": "yarn build && npx cap copy && cd ./packages/android && ./gradlew assembleRelease",
+    "i18n-unused-display": "i18n-unused display-unused",
+    "i18n-unused-mark": "i18n-unused mark-unused",
```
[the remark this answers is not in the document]
CarlosNZ: Yeah, have replaced "mark" with "missing": 786c3e67

I had thought that we'd get alerted for missing strings when we try and compile, but can't hurt to have it.

packages/common/src/intl/locales/en/catalogue.json
CarlosNZ: I've made a note to the developer about that, should be fixed in upcoming release.

CarlosNZ:
> When I run it now, I get `Total unused translations count: 87` - perhaps you need to run it again? Or am I doing something wrong? The unused are correctly detected, such as 'app.admin' in common.. which is now in `app` and just `admin`

Yup, I think I must have re-merged from `main` after my initial "remove" run, then forgot to do it again. Have re-run `remove` again now, and "display" correctly shows "Total: 0" 4eee624a

i18n-unused.config.js
CarlosNZ: Thanks, have used your suggestion below 👍🏼

CarlosNZ:
Will merge now, we can see how this goes -- might end up replacing it with a different tool if it's too annoying. 🤷‍♂️

PR #1161, opened 2022-04-05
#1100 Don't mutate if no changes

### LINKED ISSUE #1100: Don't run mutation from Modal editor if no changes
Currently on Modals for editing details (e.g. Inbound Shipments), when clicking "Okay", the update mutation is running even if nothing has been changed.

This introduces a slight delay that shouldn't be needed, and could also cause problems if the back-end is keeping an activity log (it would show as an edit even though it wasn't).

Should check if data has changed, and if not just do the same thing as "Cancel"

CarlosNZ:
Fix #1100

Just used the existing `isDirty` state to conditionally run the mutation for saving. Plus set `isDirty` back to `false` in a couple of places where it needed to be.

Currently just affects the Inbound/Output shipment line edit modals. Any other places we should have this? (These are the only places where `useDirtyCheck` hook is being used)

### DIFF
--- a/packages/common/src/hooks/useDirtyCheck/useDirtyCheck.ts
+++ b/packages/common/src/hooks/useDirtyCheck/useDirtyCheck.ts
@@ -2,7 +2,7 @@ import { useRef } from 'react';
 
 type DirtyState = {
   isDirty: boolean;
-  setIsDirty: (open: boolean) => void;
+  setIsDirty: (value: boolean) => void;
 };
 
 export const useDirtyCheck = (): DirtyState => {
--- a/packages/common/src/hooks/useDirtyCheck/useDirtyCheck.ts
+++ b/packages/common/src/hooks/useDirtyCheck/useDirtyCheck.ts
@@ -1,16 +1,11 @@
-import { useRef } from 'react';
+import { useState } from 'react';
 
 type DirtyState = {
   isDirty: boolean;
-  setIsDirty: (open: boolean) => void;
+  setIsDirty: (on: boolean) => void;
 };
 
 export const useDirtyCheck = (): DirtyState => {
-  const ref = useRef<boolean>(false);
-  return {
-    isDirty: ref.current,
-    setIsDirty: isDirty => {
-      ref.current = isDirty;
-    },
-  };
+  const [isDirty, setIsDirty] = useState(false);
+  return { isDirty, setIsDirty };
 };
--- a/packages/common/src/hooks/useDirtyCheck/useDirtyCheck.ts
+++ b/packages/common/src/hooks/useDirtyCheck/useDirtyCheck.ts
@@ -1,11 +1,16 @@
-import { useState } from 'react';
+import { useRef } from 'react';
 
 type DirtyState = {
   isDirty: boolean;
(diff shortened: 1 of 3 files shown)


josh-griffin (commented):
Oh hi!

Looks right.. but I don't remember making [this hook](https://github.com/openmsupply/openmsupply-client/blob/main/packages/common/src/hooks/useDirtyCheck/useDirtyCheck.ts) - being global state seems incorrect. Won't this be hunchy funchy when opening and working with different items? (i.e. set as dirty for one, then it's not reset when you open the modal for another item?

How the hook is being used looks like it might be better of as a simple `useRef` .. the setting of the dirty state isn't required to be state as it's not being used in the render path, just as a condition on some functions 🤔  

Maybe could also be used for a warning when clicking cancel, but then it's modal on modal 🤮 

Also, off road but the bundle size has blown up a lot [this](https://github.com/openmsupply/openmsupply-client/pull/1161#issuecomment-1089434245) v [this](https://github.com/openmsupply/openmsupply-client/pull/1081#issuecomment-1071703476) 

You're a coding MACHINE!!!



OK Bye 👋

mark-prins (approved):
👍  Looks good thanks!

An example of the `useDirtyCheck` change suggested by Josh:

```
import { useRef } from 'react';

type DirtyState = {
  isDirty: boolean;
  setIsDirty: (open: boolean) => void;
};

export const useDirtyCheck = (): DirtyState => {
  const ref = useRef<boolean>(false);
  return {
    isDirty: ref.current,
    setIsDirty: isDirty => {
      ref.current = isDirty;
    },
  };
};
```

seems to work ok 🤷‍♂️  Agree with his point about global state, so worth changing, I think

CarlosNZ:
> seems to work ok 🤷‍♂️ Agree with his point about global state, so worth changing, I think

Yeah, that works fine. Thanks.

Just wondered -- is there any problem simplifying it even further and just have a single `useState` value in there? It won't have global state as every component will have its own instance:

```
import { useState } from 'react';

type DirtyState = {
  isDirty: boolean;
  setIsDirty: (on: boolean) => void;
};

export const useDirtyCheck = (): DirtyState => {
  const [isDirty, setIsDirty] = useState(false);
  return { isDirty, setIsDirty };
};
```
Any problems with this? I guess I'm not fully getting what `useRef` offers that we can't achieve just with `useState`

mark-prins:
> Just wondered -- is there any problem simplifying it even further and just have a single useState value in there? 

`useState` is probably fine; the key difference is that `useState` will cause component re-renders whereas `useRef` doesn't; that's why I'd use a ref here, we just want to keep a hold of the value and so the ref is a neater fit with the use case. I'm sure there are finer details that I'm unaware of.. but 🤷

CarlosNZ:
> `useState` is probably fine; the key difference is that `useState` will cause component re-renders whereas `useRef` doesn't; that's why I'd use a ref here, we just want to keep a hold of the value and so the ref is a neater fit with the use case. I'm sure there are finer details that I'm unaware of.. but 🤷

Okay, thanks for the clarification. Have left it with the `useRef`.

PR #1164, opened 2022-04-06
#1163 permission errors

### LINKED ISSUE #1163: Handle permission errors
With this PR https://github.com/openmsupply/remote-server/pull/1010 there is now an error when permission is denied; the example is a requisition mutation, which will give this if the user does not have permission to create/edit requisitions

```
{
    "data": null,
    "errors": [
        {
            "message": "Forbidden",
            "locations": [
                {
                    "line": 2,
                    "column": 3
                }
            ],
            "path": [
                "insertRequestRequisition"
            ],
            "extensions": {
                "details": "InsufficientPermission((\"\", UserPermissions { api: [Admin], stores: {} }))"
            }
        }
    ]
}
```

mark-prins:
Fixes #1163 

I've removed the 'create/edit requisition' permission from the user `Kopu` : if you login as them and try to create or edit a requisition you'll get a modal, as well as the toast notification:

Screen Shot 2022-04-07 at 9 57 01 AM

### DIFF
--- a/packages/common/src/api/GqlContext.tsx
+++ b/packages/common/src/api/GqlContext.tsx
@@ -9,6 +9,33 @@ import {
 import { AuthError } from '../authentication/AuthContext';
 import { LocalStorage } from '../localStorage';
 
+interface ResponseError {
+  message?: string;
+  path?: string[];
+  extensions?: { details?: string };
+}
+
+function hasError(errors: ResponseError[], error: AuthError) {
+  return errors.some(({ message }: { message?: string }) => message === error);
+}
+
+const handleResponseError = (errors: ResponseError[]) => {
+  if (hasError(errors, AuthError.Unauthenticated)) {
+    LocalStorage.setItem('/auth/error', AuthError.Unauthenticated);
+    return;
+  }
+
+  if (hasError(errors, AuthError.PermissionDenied)) {
+    LocalStorage.setItem('/auth/error', AuthError.PermissionDenied);
+    return;
+  }
+
+  const error = errors[0];
+  const { extensions } = error || {};
+  const { details } = extensions || {};
+  throw new Error(details || error?.message || 'Unknown error');
+};
+
 class GQLClient extends GraphQLClient {
   private client: GraphQLClient;
   private emptyData: object;
@@ -44,21 +71,10 @@ class GQLClient extends GraphQLClient {
       data => data ?? this.emptyData,
       ({ response }) => {
         if (response && response.errors) {
-          if (
-            response.errors.some(
-              ({ message }: { message?: string }) =>
-                message === AuthError.Unauthenticated
-            )
-          ) {
-            LocalStorage.setItem('/auth/error', AuthError.Unauthenticated);
-          } else {
-            const error = response.errors[0];
-            const { extensions } = error;
-            const { details } = extensions || {};
-            throw new Error(details || error.message);
-          }
+          handleResponseError(response.errors);
+        } else {
+          throw new Error('Unknown error');
         }
-        throw new Error('Unknown error');
       }
     );
   }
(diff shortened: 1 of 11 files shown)


packages/common/src/api/GqlContext.tsx
```diff
@@ -9,6 +9,33 @@ import {
...
+  return errors.some(({ message }: { message?: string }) => message === error);
+}
+
+const handleResponseError = (errors: ResponseError[]) => {
+  if (hasError(errors, AuthError.Unauthenticated)) {
+    LocalStorage.setItem('/auth/error', AuthError.Unauthenticated);
+    return;
+  }
+
+  if (hasError(errors, AuthError.PermissionDenied)) {
```
mark-prins, line 28: add the new error type

packages/common/src/authentication/AuthContext.tsx
```diff
@@ -15,7 +15,7 @@ const TOKEN_CHECK_INTERVAL = 60 * 1000;
 
 export enum AuthError {
   NoStoreAssigned = 'NoStoreAssigned',
-  PermissionDenied = 'PermissionDenied',
+  PermissionDenied = 'Forbidden',
```
mark-prins, line 18: it's just what the API is returning as the error 🤷

packages/common/src/authentication/api/api.ts
```diff
@@ -74,13 +74,15 @@ export const getAuthQueries = (sdk: Sdk) => ({
...
-      return result.me;
+      try {
+        const result = await sdk.me(
+          {},
+          {
+            Authorization: `Bearer ${token}`,
+          }
+        );
+        return result.me;
+      } catch {}
```
mark-prins, line 85: the login page was returning the 'server connection' error.. and then this query was throwing another error which hid the message. we don't need this one to toast, as auth issues are handled elsewhere

packages/common/src/authentication/api/hooks/useUserDetails.ts
```diff
@@ -4,9 +4,12 @@ import { useAuthApi } from './useAuthApi';
 
 export const useUserDetails = () => {
   const api = useAuthApi();
-  return useMutation<UserNode, unknown, string | undefined, unknown>(
-    api.get.me
-  );
+  return useMutation<
+    UserNode | undefined,
```
mark-prins, line 8: required because of the error suppression which isn't returning an object when there's a problem

packages/host/src/QueryErrorHandler.tsx
```diff
@@ -16,7 +16,11 @@ export const QueryErrorHandler = () => {
     if (!!errorMessage) {
       error(errorMessage)();
     }
-  }, [errorMessage, location.pathname]);
+  }, [errorMessage]);
+
+  useEffect(() => {
```
mark-prins, line 21: offroad.. when there was an error it was getting stuck and re-toasting when you navigated

packages/inventory/src/Stocktake/api/api.ts
```diff
@@ -134,23 +134,24 @@ export const getStocktakeQueries = (sdk: Sdk, storeId: string) => ({
     patch: RecordPatch<StocktakeFragment>
   ): Promise<UpdateStocktakeInput> => {
     const input = stocktakeParser.toUpdate(patch);
-    const result = await sdk.updateStocktake({ input, storeId });
+    const result = (await sdk.updateStocktake({ input, storeId })) || {};
```
mark-prins, line 137: and then there's a bunch of these.. this is because if we get `Forbidden` then there's no data object returned, so this part blows up. We don't want the destructuring error to be toasted, just the 'Unable to create requisition' etc. So.. just being careful with the mutations, to not throw an error and this allows a better error state to be shown.

CarlosNZ (approved):
LGTM.

Thanks Mark, working nicely. Looks like a lot of the work here was just accommodating the type changes as a result of the endpoint not returning an object when Forbidden. 👍

PR #1166, opened 2022-04-07
Open on click nav link

### LINKED ISSUE #1165: Left nav on mobile
Noticed that the app drawer isn't expanding on click any more, making the nav fairly useless on mobile.

mark-prins:
Fixes #1165 

Opens the drawer (if closed) on click, and if the drawer was opened this way, closes on clicking a menu item.
It does mean that you can open the drawer by clicking on a category menu item and then leave the drawer open by clicking on something in-page, like navigating to a detail page. Still.. this is better than it was.

### DIFF
--- a/packages/common/src/hooks/useDrawer/useDrawer.ts
+++ b/packages/common/src/hooks/useDrawer/useDrawer.ts
@@ -9,6 +9,7 @@ type DrawerController = {
   open: () => void;
   close: () => void;
   toggle: () => void;
+  onClick: () => void;
   setHoverOpen: (open: boolean) => void;
   setClickedNavPath: (clicked?: string) => void;
 };
@@ -26,6 +27,13 @@ export const useDrawer = create<DrawerController>(set => {
     close: () => set(state => ({ ...state, isOpen: false, hoverOpen: false })),
     toggle: () =>
       set(state => ({ ...state, isOpen: !state.isOpen, hasUserSet: true })),
+    onClick: () =>
+      set(state => {
+        if (state.isOpen && state.hoverOpen) {
+          return { ...state, isOpen: false, hoverOpen: false };
+        }
+        return state;
+      }),
   };
 });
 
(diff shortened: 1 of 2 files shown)


packages/common/src/hooks/useDrawer/useDrawer.ts
```diff
@@ -26,6 +27,13 @@ export const useDrawer = create<DrawerController>(set => {
     close: () => set(state => ({ ...state, isOpen: false, hoverOpen: false })),
     toggle: () =>
       set(state => ({ ...state, isOpen: !state.isOpen, hasUserSet: true })),
+    onClick: () =>
```
mark-prins, line 30: this was because the AppNavLink only knows the current state if it re-renders, and I didn't see the need to increase the number of renders just to get the state.

CarlosNZ (commented):
Slightly confused, cos the changes seem to make it less useable as far as I can tell. Not sure what the original intended behaviour was.

Before this change it's working as follows:
- Clicking the spinny man icon toggles between auto-collapse(opens on hover) and always open. 

But after these changes, it's almost the same but when going to a new menu section it switches back to auto-collapse. So there doesn't seem to be a way to keep the drawer permanently open any more. 🤷‍♂️

mark-prins:
oh - that's not the intention. The collapse should only happen if the nav was opened using the hover/click.

The issue being fixed is that when you use the site on mobile, once you collapse the navigation, clicking on the nav items does not open the navigation. making the nav a little annoying. you have to explicitly open the navigation again. The change here is to make the click open the nav, which then closes again on navigate.

On desktop this is handled with hoverOver / hoverOut.

CarlosNZ:
> The collapse should only happen if the nav was opened using the hover/click.

So on Desktop, how do I keep the drawer permanently open? I can click the spinning man, and it'll lock, but only until I change to another page, then it collapses again :(

mark-prins:
> So on Desktop, how do I keep the drawer permanently open?

sounds like a bug in this change.. will check it out. It should stay open

mark-prins:
I see - if you open the drawer ( spinny man ) and then click one of the section headings, then click a menu item, the drawer was closing. Fixed with fd9597a6

CarlosNZ (approved):
Cool, working nicely now.

Tested on Desktop and mobile (well, simulated mobile in browser).

PR #1167, opened 2022-04-07
Tidy report usage

### LINKED ISSUE #1159: Printing refinements
Some user feedback:
- Rename `Report` to `Form` or `Template` or something, in the 'select a report' dialog
- Maybe do something if there is only one report. Is there a need to select from a list?

mark-prins:
Fixes #1159 

Renames the selection heading and prints without prompting if there is only one report.
To aid in testing, I've created a new 'invoices' report.. so outbound and inbound will still show the selection popover.

### DIFF
--- a/packages/common/src/intl/locales/en/app.json
+++ b/packages/common/src/intl/locales/en/app.json
@@ -57,5 +57,5 @@
   "button.login": "Log in",
   "error.login": "Invalid username or password",
   "select-store": "Select store",
-  "select-report": "Select Report"
+  "select-report": "Select Template"
 }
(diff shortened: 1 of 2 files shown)


CarlosNZ (approved):
LGTM, but I can't actually find a case where there is more than one report to choose from. 

Can you let me know what combination of store/user/page I should try?

mark-prins:
> Can you let me know what combination of store/user/page I should try?

Inbound or Outbound shipments - the invoices have two reports on the demo server:

Screen Shot 2022-04-11 at 9 43 25 AM

PR #1168, opened 2022-04-07
Change column source

### LINKED ISSUE #1036: response column already issued should use the requested quantity
currently using 

```
accessor: ({ rowData }) =>
          rowData.supplyQuantity - rowData.remainingQuantityToSupply,
```

This means that when you change the supply quantity, the 'already issued' changes which seems odd to me. The amount that has been issued should be the quantity the user requested minus the quantity remaining to supply and not change while editing the requisition itself

mark-prins:
Fixes #1036 
Seemed reasonable, so I've changed. Raised an issue for the API which appears incorrect: https://github.com/openmsupply/remote-server/issues/1017


CarlosNZ (approved):
LGTM :)

PR #1170, opened 2022-04-07
Retain placeholder for new invoices

### LINKED ISSUE #1106: Bug: stock quantity error when adding additional placeholder lines
Steps to reproduce:
...

mark-prins:
Fixes #1106 

Subtle error, this one. The placeholder row was being added to `draftOutboundLines`.. but then on re-render the existing lines on the invoice are checked for `UNALLOCATED` rows, and since there isn't one, a new placeholder is created.

When updating, the first placeholder line is updated, but that no longer exists in state and fails. To resolve, check for lines from the api first and if that fails, check for lines that you've created. only if that doesn't exist then create a new one.

### DIFF
--- a/packages/common/src/hooks/useDirtyCheck/useDirtyCheck.ts
+++ b/packages/common/src/hooks/useDirtyCheck/useDirtyCheck.ts
@@ -1,16 +1,11 @@
-import { useRef } from 'react';
+import create from 'zustand';
 
 type DirtyState = {
   isDirty: boolean;
-  setIsDirty: (value: boolean) => void;
+  setIsDirty: (open: boolean) => void;
 };
 
-export const useDirtyCheck = (): DirtyState => {
-  const ref = useRef<boolean>(false);
-  return {
-    isDirty: ref.current,
-    setIsDirty: isDirty => {
-      ref.current = isDirty;
-    },
-  };
-};
+export const useDirtyCheck = create<DirtyState>(set => ({
+  isDirty: false,
+  setIsDirty: isDirty => set(state => ({ ...state, isDirty })),
+}));
--- a/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEdit.tsx
+++ b/packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEdit.tsx
@@ -20,6 +20,7 @@ import {
   useDraftOutboundLines,
   usePackSizeController,
   useNextItem,
+  PackSizeController,
 } from './hooks';
 import {
   allocateQuantities,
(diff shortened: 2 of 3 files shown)


packages/common/src/hooks/useDirtyCheck/useDirtyCheck.ts
```diff
@@ -1,16 +1,11 @@
-import { useRef } from 'react';
+import create from 'zustand';
```
mark-prins, line 1: aaaand, this didn't work. The hook rerendered and created a new ref, meaning the form is never dirty and new lines are never saved.

Could probably fix that.. but I went with option (a) and just reverted.

packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/OutboundLineEdit.tsx
```diff
@@ -123,41 +126,67 @@ export const OutboundLineEdit: React.FC<ItemDetailsModalProps> = ({
...
-              flex={1}
-              height={400}
-              justifyContent="center"
-              alignItems="center"
-            >
-              <InlineSpinner />
-            </Box>
-          )
-        ) : null}
+        <TableWrapper
```
mark-prins, line 129: three levels of nesting was too much for me to cope with!

packages/invoices/src/OutboundShipment/DetailView/OutboundLineEdit/hooks/useDraftOutboundLines.tsx
```diff
@@ -150,9 +150,14 @@ export const useDraftOutboundLines = (
         .sort(SortUtils.byExpiryAsc);
 
       if (status === InvoiceNodeStatus.New) {
-        const placeholder = lines?.find(
+        let placeholder = lines?.find(
           ({ type }) => type === InvoiceLineNodeType.UnallocatedStock
         );
+        if (!placeholder) {
```
mark-prins, line 156: and here it is.. the actual fix 😁

packages/common/src/hooks/useDirtyCheck/useDirtyCheck.ts
CarlosNZ: I thought we wanted this hook to avoid using global state -- as per last week's discussion? So when you say "reverted" here, you mean put it back to how it was before my changes? 😄

mark-prins: yep. exactly - just put the global state back in 😜  
the ref version wasn't working for me.. when editing the `dirty` flag kept changing `false` 🤷

CarlosNZ (approved):
Just left a comment about the other approach to `useDirtyCheck` that won't make state global. Not sure if better or worse -- see what you think.

mark-prins:
> does it work with that other approached I tried last week

No - I had tried that, and have just retested. It has the same issues. The only option which works reliably is the zustand state, which is why I ended up with it originally. There could well be something causing the re-render which could be isolated and fixed.. perhaps later though. for now, this is working.

PR #1172, opened 2022-04-08
#577 speech bubble comment

### LINKED ISSUE #577: The comment column for supplier requisition lines should be the speech bubble column
Currently the comment column for supplier requisition lines is just a standard text cell.

Since there are so many cells, having the smaller speech bubble column would be good.

Should the comment be editable?

mark-prins:
Fixes #577 

Changed to the speech bubble comment style
Screen Shot 2022-04-08 at 1 58 31 PM

### DIFF
--- a/packages/host/src/CommandK.tsx
+++ b/packages/host/src/CommandK.tsx
@@ -167,13 +167,6 @@ export const CommandK: FC = ({ children }) => {
         ),
     },
     {
-      id: 'navigation:reports',
-      name: `${t('cmdk.goto-reports')} (r)`,
-      shortcut: ['r'],
-      keywords: 'reports',
-      perform: () => navigate(RouteBuilder.create(AppRoute.Reports).build()),
-    },
-    {
       id: 'navigation:suppliers',
       name: `${t('cmdk.goto-suppliers')} (g+s)`,
       keywords: 'suppliers',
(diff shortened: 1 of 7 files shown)


packages/host/src/CommandK.tsx
```diff
@@ -166,13 +166,6 @@ export const CommandK: FC = ({ children }) => {
             .build()
         ),
     },
-    {
```
mark-prins, line 169: offroad: found that while I was testing on the demo server I was accidentally mashing the `r` key and getting a reports placeholder page 🙄

packages/host/src/Site.tsx
```diff
@@ -40,11 +40,12 @@ const Heading: FC = ({ children }) => (
 
 const NotifyOnLogin = () => {
   const { success } = useNotification();
-  const { store } = useAuthContext();
+  const { store, storeId } = useAuthContext();
```
mark-prins, line 43: offroad: when testing the logged out alert, I noticed that when the `refreshToken` happened, the user received two toasts to say 'welcome to xx store' even though they have already been notified. Changing the reference here removes the duplication and removes the notification-on-refresh-token, which should not have been happening

CarlosNZ (commented):
Cool, do you think the ResponseRequisition detail view should also use the popover comment column too, since it's basically the same pattern?

Even the Outbound/Inbound shipment lists too, perhaps?

mark-prins:
> do you think the ResponseRequisition detail view should also use the popover comment column too, since it's basically the same pattern?

cool - have done. 

I've left Inbound / Outbound lists for now, since there's plenty of room and we have the 'colour' column on the LH end already

CarlosNZ (approved):
LGTM 👍

PR #1174, opened 2022-04-10
A whole bunch of bits

### LINKED ISSUE #1169: Usability issues: episode 2
# Outbound Shipment

- [x] O/Shipment: created a line for an item with no stock (placeholder). When I use the "Allocate Placeholder Lines" function, notification says 1 line has been allocated but in fact... no :) Suggestion for notif: 
```
Allocated X line (when line has been fully allocated)
Partially allocated X line (when line has been partially allocated)
X lines not allocated 
```
- [x] when i try to delete a line when status is "shipped", notif says "no you can't unless it's new" but actually you can delete line when status is allocated or picked. Suggestion: You cannot delete a line when status is shipped. 
- [x] O/Shipment: On large screen, OS list do not take the full width of the screen

# Inbound Shipment

- [x] I/Shipment: when deleting, notification says 'Invoice' instead of 'shipment

# Internal Order

## Internal Order's List

- [x] `New Requisition` Button: should be `New Order` ?
- [x] `Search by comment...`: maybe more useful to look by supplier name or status ? (Waiting for a complex search in future releases). 
- [x] Internal order list: add `Created on` 
- ~sorting list with column headers: no way to cancel sorting except if you refresh the page.~
- [x] when deleting, notification says `Deleted X requisition(s)`, should be `Deleted X order(s)`
- ~`Export` button: bulk action ?~

###  Internal Order Headers

- ~`Requested to Suggested` could be a bulk action (`Actions` dropdown)~
- [x] `Supplier Ref` should be `Reference` or `Order Reference`
- [x] No need to skip status for Internal Orders
Screen Shot 2022-04-08 at 2 25 20 PM

###  Internal Order Information Panel

- [x] Missing additional info: `Created on`, `Last modified on`, `Last modified by`

### Order line modal

- [x] Internal Orders > Consumption History chart: orange datapoint should be "AMC", not "Req. Qty." (an error from Manu's preview)

Screen Shot 2022-04-07 at 4 36 48 PM

- [x] Remove stocks details & Order title. Keep the one for Comment (and if not too long, put the comment on the right)
- [x] Internal Orders > Charts: Could we add SOH under AMC ?
Screen Shot 2022-04-07 at 4 38 48 PM
- [x] Internal Orders > Charts: Stock Distribution is not a correct title. Maybe no title at all ?  
Screen Shot 2022-04-07 at 4 41 13 PM

mark-prins:
Fixes #1169 

There's a number of UI refinements, as requested by Richard


packages/common/src/ui/layout/tables/utils/ColumnDefinitionSetBuilder.ts
```diff
@@ -82,7 +82,6 @@ const getColumnLookup = <T extends RecordWithId>(): Record<
   itemName: {
     key: 'itemName',
     label: 'label.name',
-    width: 300,
```
mark-prins, line 85: this may cause layout issues in other pages - but if so, I couldn't find them. All looked ok to me

packages/invoices/src/InboundShipment/api/hooks.ts
```diff
@@ -296,8 +296,8 @@ export const useDeleteSelectedInbounds = () => {
   );
 
   const deleteAction = () => {
-    const number = selectedRows?.length;
-    if (selectedRows && number > 0) {
+    const count = selectedRows?.length;
```
mark-prins, line 299: renamed to `count` so that I could use plurals

packages/invoices/src/OutboundShipment/DetailView/ContentArea.tsx
```diff
@@ -107,7 +107,7 @@ export const ContentAreaComponent: FC<ContentAreaProps> = ({ onRowClick }) => {
   if (!rows) return null;
 
   return (
-    <Box flexDirection="column">
+    <Box flexDirection="column" style={{ width: '100%' }}>
```
mark-prins, line 110: was not stretching full width on larger screens

packages/invoices/src/OutboundShipment/DetailView/columns.ts
```diff
@@ -82,7 +82,6 @@ export const useOutboundColumns = ({
       [
         'itemName',
         {
-          maxWidth: 400,
```
mark-prins, line 85: as far as I could see, all columns were fixed, and we want the table to stretch

packages/invoices/src/OutboundShipment/api/hooks/document/useOutboundDelete.ts
```diff
@@ -38,7 +38,7 @@ export const useOutboundDelete = () => {
           onSuccess: () => queryClient.invalidateQueries(api.keys.base()),
         });
         const deletedMessage = t('messages.deleted-invoices', {
-          number: numberSelected,
+          count: numberSelected,
```
mark-prins, line 41: plurals again

packages/invoices/src/OutboundShipment/api/hooks/line/useOutboundAllocateSelectedLines.ts
```diff
@@ -67,12 +67,45 @@ export const useOutboundAllocateSelectedLines = (): {
...
+    if (batchResponse?.__typename === 'BatchOutboundShipmentResponse') {
+      const { allocateOutboundShipmentUnallocatedLines } = batchResponse;
+      const count = {
+        success: 0,
+        partial: 0,
+        failed: 0,
+      };
+
+      allocateOutboundShipmentUnallocatedLines?.forEach(line => {
+        const { id, response } = line;
```
mark-prins, line 81: I'm not entirely enjoying the very specific, and nested, shape of returned objects. Have tried to be as safe as possible, though I wonder now if we may end up with no notifications when the api return shape is changed again. I _think_ that's better than notifying incorrectly?

packages/invoices/src/OutboundShipment/api/hooks/line/useOutboundAllocateSelectedLines.ts
```diff
@@ -67,12 +67,45 @@ export const useOutboundAllocateSelectedLines = (): {
...
+          }
+          if (response.inserts.totalCount > 0) {
+            count.partial++;
+            return;
+          }
+          count.failed++;
+        }
+      });
 
+      if (count.success > 0) {
```
mark-prins, line 97: The request was to show full, partial and failed - but we don't have much space in a snackbar, so I went with this approach. Most likely users will have only one type, so I thought this would work. We could go all out and have a summary modal, but I prefer the less invasive snackbar. you don't have to click to clear the snack, like you do a modal.

packages/requisitions/src/RequestRequisition/DetailView/Footer/StatusChangeButton.tsx
```diff
@@ -98,7 +98,7 @@ const useStatusChangeButton = () => {
     onConfirm: onConfirmStatusChange,
   });
 
-  // When the status of the invoice changes (after an update), set the selected option to the next status.
+  // When the status changes (after an update), set the selected option to the next status.
```
mark-prins, line 101: Yes, all of this cold just be one button. I've kept the component in case the status list changes again, or if finalising becomes possible ( currently it isn't, we aren't allowed to modify a 'sent' requisition )

packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/ItemCharts/StockDistribution.tsx
```diff
@@ -106,13 +112,20 @@ const MonthlyBar = ({
         borderColor: 'gray.dark',
         borderStyle: 'solid',
         height: '20px',
+        overflow: 'hidden',
```
mark-prins, line 115: on the items where I've added far too much stock, the monthly summary is showing so very small we can't see anything! I've added the overflow to make it look ok, and a tooltip to show the details.

packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEditForm.tsx
```diff
@@ -61,7 +61,13 @@ export const RequestLineEditFormLayout = ({
       <Grid item xs={4}>
         {Left}
       </Grid>
-      <Grid item xs={4}>
+      <Grid
```
mark-prins, line 64: to push the input fields down the bottom. I thought it looked better there, now that the heading has been removed.

packages/invoices/src/OutboundShipment/DetailView/ContentArea.tsx
CarlosNZ: How wide are we talking? I ran in full-screen on 1680x1050 with the Drawer closed and the table still went full-width without this style setting. 

Then again, I guess there's no downside either way, so... 🤷‍♂️

packages/invoices/src/OutboundShipment/DetailView/columns.ts
CarlosNZ: Same comment as above. We do want the table to stretch, but if a particular item has a name that fills most of the screen, we probably want to put a limit on it at some point. I think `maxWidth` is fine, as the column will still stretch appropriately, it just won't let it get ridiculous! 😄

Maybe `maxWidth: 650`?

packages/invoices/src/OutboundShipment/api/hooks/line/useOutboundAllocateSelectedLines.ts
CarlosNZ: > I _think_ that's better than notifying incorrectly?

Yeah. Is it possible to have a general warning notification that always pops up whenever the return shape is different to what's expected?

Maybe only when in dev environment, if that's possible?  Just thinking that we'd want to notice this ASAP.

packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEditForm.tsx
CarlosNZ: Or vertically centered? I dunno, it's a pretty trivial different tbh.

Screen Shot 2022-04-12 at 11 21 02 AM

CarlosNZ (approved):
Thanks! :)

packages/common/src/ui/layout/tables/utils/ColumnDefinitionSetBuilder.ts
mark-prins: I thought that adding a `maxWidth` would constrain the table on really wide screens.. but what it does is expand the width of the other columns such as location and unit. That's going to look a bit strange when the long item names are truncated and there's heaps of space around other columns.

I've added in a max width - but am still not sure we have the column sizing good in all scenarios.

packages/invoices/src/OutboundShipment/DetailView/ContentArea.tsx
mark-prins: my monitor is 1920 and doesn't feel like super-wide, but does this:

Screen Shot 2022-04-12 at 2 16 04 PM

The problem was original noticed by Richard on his screen in the office. probably similar?

packages/invoices/src/OutboundShipment/api/hooks/line/useOutboundAllocateSelectedLines.ts
mark-prins: perhaps bugsnag could catch it too?

packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEditForm.tsx
mark-prins: yeah, I though that vertically centred would look good.. I thought bottom aligned was the best of the options tried though. The mis-alignment of those input fields isn't great

packages/common/src/ui/layout/tables/utils/ColumnDefinitionSetBuilder.ts
CarlosNZ: >  but am still not sure we have the column sizing good in all scenarios.

Yeah, I agree. I think putting a small-ish max-width on most columns would help, especially the ones with numbers, and then letting the free text fields grow, but only up to a point. But yeah, will take a bit of trial and error, and is quite time consuming to check all scenarios.

I'll make a note to notice when a column is looking weird and I'll take a screenshot. Then once we've got a decent number of these we can make some overall adjustments.

PR #1177, opened 2022-04-11
Log bugs

### LINKED ISSUE #53: Add logging
I like logging.

I would love to have a lot of logging everywhere which can be toggled on/off. Buuut.. I think in a browser environment this is a little harder? There is no file to write too. So we need to keep logs in memory and periodically hit a server api?

I like the idea of that. But maybe we just add bugsnag and use breadcrumbs, for now?

mark-prins:
Fixes #53 

Added in bugsnag with a minimal implementation. 
Added version information to the `package.json` - which is then shown on the settings page:

Screen Shot 2022-04-12 at 9 05 12 AM

as well as the api version.
Added github actions to bump versions, and create a tag when doing so


packages/common/src/types/schema.ts
```diff
@@ -1734,7 +1734,7 @@ export type ItemNode = {
   atcCategory: Scalars['String'];
   availableBatches: StockLineConnector;
   code: Scalars['String'];
-  ddd: Scalars['Float'];
+  ddd: Scalars['String'];
```
mark-prins, line 1737: this one was a surprise; have updated as required.

packages/host/src/Host.tsx
```diff
@@ -42,6 +44,11 @@ const queryClient = new QueryClient({
   },
 });
 
+Bugsnag.start({
+  apiKey: 'a09ce9e95c27ac1b70ecf3c311e684ab',
```
mark-prins, line 48: thought to encode in here, rather than put into a config file - it's required to be in client side code, and being in here is at least more obfuscated than in a `config.js` file. There's little risk in this going out anyway - and is unlikely to change.

packages/common/src/types/schema.ts
clemens-msupply: what is "ddd"?

clemens-msupply:
If we just want to have a reference for testers/devs for which exact version is running we could also use the git short hash. You could extract this info when doing the bundle and you don't have to bump versions manually... E.g. have `App version`, `Build id` and `Api version`

packages/host/src/Admin/Settings.tsx
```diff
@@ -98,6 +101,32 @@ export const Settings: React.FC = () => {
         onToggle={onToggleCustomLogo}
         title={t('heading.custom-logo')}
       />
+      <Grid style={{ position: 'absolute', right: 0, bottom: 30 }}>
+        <Grid container padding={1} flexDirection="column">
+          <Grid item display="flex" flex={1} gap={1}>
+            <Grid item justifyContent="flex-end" flex={1} display="flex">
+              <Typography fontWeight={700} whiteSpace="nowrap">
+                App version:
```
CarlosNZ, line 109: Localise these?

CarlosNZ (approved):
Great, looks good.

Just wondering -- what counts as a "bug" that gets reported? Any uncaught exception? I tried forcing a few errors and I couldn't see anything show up in our [bugsnag page](https://app.bugsnag.com/sustainable-solutions-nz-ltd/openmsupply-client/overview?release_stage=development).

packages/common/src/types/schema.ts
mark-prins: Defined Daily Dose - which, according to WHO is "The assumed average maintenance dose per day for a drug used for its main indication in adults"

mark-prins:
thanks @clemens-msupply - that's a good idea!

PR #1180, opened 2022-04-12
Show name.isOnHold

### LINKED ISSUE #451: Display if a name is on hold before trying to make an invoice
mSupply:

image

should we be displaying if the name is on hold? Additionally if the name is for a stock transfer (red highlight, I think?)

Have added to the api changes if so: https://docs.google.com/spreadsheets/d/1H0Rv1DEOgW0cE07wpfc_KivLf8TcNhYDqlwRp92x-To/edit#gid=1127702983&range=A109

mark-prins:
Fixes #451 
Shows if a name is on hold, and prevents using the name if they are on hold.
A lot of file changes for a simple thing..

Result is:

Screen Shot 2022-04-12 at 4 52 53 PM

Screen Shot 2022-04-12 at 4 53 01 PM

this example in the Opua store - where I've added Waikato as a customer and supplier, and put it on hold


clemens-msupply (approved):
Looks good to me!

For graphql, you edit the `*.graphql` file and then ran `yarn run generate`?

Just some observations/feedback from looking at the screenshots. As a "new" user its not quite clear to me what "on holds" actually means. What is the reason it is on hold/disabled? (we don't track this in, e.g. the `note`, do we?) Also who set it? the other store or did we do it. And how can I "fix" it?

These are probably beginners questions and not sure if the should be addressed. Also not sure how to address it, e.g. a hover help pop up on the menu item might clutter the UI...

mark-prins:
> For graphql, you edit the *.graphql file and then ran yarn run generate?

yep. or if you want to be even lazier, `yarn generate` will do 😉 
It is running generate off the demo server, which is running the current `develop` branch.

Good points about 'oh hold' and what that means. and how to correct it! I did initially just add a tooltip to the option.. but.. with the tooltip component we are using, it doesn't allow tooltips on disabled components 🙄  so I went with the text `(On Hold)` instead. In current mSupply there's a checkbox against every item in the list, and if you click a store which is on hold it pops up a dialog telling you that you cannot use this store. Displaying a modal is a great way to show extra information, such as how to enable the store.. but is also an annoying thing to pop up and have to click to get rid of!

actually, now that I think about it, 'on hold' is a strange thing to apply to a store. should be 'disabled' - we use 'on hold' for invoices. and 'locked' for stocktakes.

PR #1182, opened 2022-04-13
Manually created inbound status values

### LINKED ISSUE #995: Statuses for manually created inbounds are a little odd, you'll never guess why
Manually created don't utilise these statuses:

image

Looks a bit odd! Maybe remove them for manually created ones?

mark-prins:
Fixes #995 
A manually created inbound shipment now has status values and disabled state.

To test: using the demo API - Kawakawa store has invoice number 2 which is manual and number 1 which is created automatically.

For automatic invoices, you cannot edit until the delivery is confirmed as 'Delivered'. 
You can change status though! 😁 

Couple of offroads: 
- added a hover to the breadcrumb on the request of @clemens-msupply ( though I kept the font bold. thoughts? the request was to keep it at a normal font weight, so that when you navigate it doesn't change so much. I thought it looked nice when bold, so left it )
- Fixed an issue when changing stores.. because I was using the store selector when testing this issue!


packages/common/src/api/hooks/useFieldsSelector/useFieldsSelector.ts
```diff
@@ -33,19 +33,16 @@ export const useFieldsSelector = <Entity, KeyOfEntity extends keyof Entity>(
   timeout = 1000
 ): FieldSelectorControl<Entity, KeyOfEntity> => {
   const queryClient = useQueryClient();
-  const select = useCallback((requisition: Entity) => {
+  const select = useCallback((entity: Entity) => {
```
mark-prins, line 36: offroad - this function is used by more than just requisitions!

packages/common/src/authentication/AuthContext.tsx
```diff
@@ -175,8 +175,7 @@ export const AuthProvider: FC = ({ children }) => {
       username: mostRecentlyUsedCredentials?.username ?? '',
       store,
     });
-    const authCookie = getAuthCookie();
-    const newCookie = { ...authCookie, store };
+    const newCookie = { ...cookie, store };
```
mark-prins, line 178: ah. yeah - the issue is that when you change stores it was reading from the cookie. If you have recently logged in, the cookie might not be set yet, and that would wipe out the token, cause an auth error and annoy Mark.

packages/common/src/authentication/api/hooks/useUserDetails.ts
```diff
@@ -14,5 +14,8 @@ export const useUserDetails = () => {
 
 export const useUserStores = (token: string) => {
   const api = useAuthApi();
-  return useQuery(api.keys.me(token), api.get.stores(), { cacheTime: 0 });
+  return useQuery(api.keys.me(token), api.get.stores(), {
```
mark-prins, line 17: Just because it is daft to ask for `me` when you have no token 🤷  
Actually not really required now that the underlying issue is fixed

packages/common/src/ui/components/navigation/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -10,6 +10,7 @@ const Breadcrumb = styled(Link)({
   color: 'inherit',
   fontWeight: 'bold',
   textDecoration: 'none',
+  '&:hover': { textDecoration: 'underline' },
```
mark-prins, line 13: the breadcrumb change

clemens-msupply (approved):
Thanks looks good!

Some personal user feedback:
- Think the breadcrumb link highlighting can be more prominent, e.g. make it look like a button.
- When having no item selected the root bread crumb isn't clickable. If a user discovers this early on the user might assume the bread crumb is never clickable and might not try to click it when it becomes clickable. IMHO better to always have it clickable.
- The change between bold and and plain font in the bread crumb bar makes it "jump", suggesting that the whole page updated and makes it harder to recognize the bread crumb bar as a bread crumb bar, i.e. bread crumb bar doesn't appear as a single widget.

Manual shipments
- Without the knowledge about manual shipment I think it would have been hard to see why the shipments are different. Maybe some indication that it has been created manually.
- Similar to previous comment, it was not clear to me why I can't change the status or can't change the hold flag. Maybe some comment that shipment needs to be received first.
- For non-manual shipments, maybe an indication which part of the status is on the remote store and which is on the local store might be helpful, e.g. a line between Shipped and Delivered with labels "sending store" and "local store" respectively.
- Having the order history and the horizontal status bar is a bit confusing. Maybe just a single status field with the order history popup?

Sorry for inlining the user feedback here, please let me know if there is a better channel to do this.

mark-prins:
Thanks for the feedback @clemens-msupply that's really helpful!

as a start - what about these?

Screen Shot 2022-04-19 at 3 58 21 PM

Screen Shot 2022-04-19 at 3 54 01 PM

clemens-msupply:
> Thanks for the feedback @clemens-msupply that's really helpful!
> 
> as a start - what about these?
> 
> Screen Shot 2022-04-19 at 3 58 21 PM
> 
> Screen Shot 2022-04-19 at 3 54 01 PM

Yes having these bubbles would be really helpful!

PR #1183, opened 2022-04-14
Usability: 4

### LINKED ISSUE #1181: Usability Issues Episode 4
### Delete a location

- [ ] It seem logical that you should be able to delete a location from another store, but why do you see it in the 1st place ? :)
Screen Shot 2022-04-12 at 4 58 30 PM

- [x] Select > Delete lines: should be delete selected lines ?

## Stock

- [x] Rename the menu => `View Stock` 

## Stocktake

- [ ] Stocktake or Stock take ;)

### Stock take: Edit Item

- [x] Similar to Inbound Shipment, tab shortcut won't work for french keyboards (AZERTY). Shift + 1 on the french KB will give you `1`. So if somebody is typing 1, 2 or 3, system would change tabs instead of writing 1
- [x] Item code appears twice:
Screen Shot 2022-04-12 at 6 42 30 PM

- [ ] Add batch button:  + icon appears twice: 

Screen Shot 2022-04-12 at 6 45 31 PM

### Stock take: Add item
- [x] item is not added when using `Add Item` button

mark-prins:
Fixes #1181 

Some more usability items raised by Richard

### DIFF
--- a/packages/common/src/intl/locales/en/inventory.json
+++ b/packages/common/src/intl/locales/en/inventory.json
@@ -1,5 +1,5 @@
 {
-  "button.delete-lines": "Delete lines",
+  "button.delete-lines": "Delete selected lines",
   "button.save-and-confirm-status": "Confirm {{status}}",
   "error.cant-save": "There was a problem saving.",
   "error.stocktake-not-found": "Stocktake not found",
--- a/packages/common/src/ui/components/navigation/Tabs/TabKeybindings.tsx
+++ b/packages/common/src/ui/components/navigation/Tabs/TabKeybindings.tsx
@@ -1,0 +1,37 @@
+import React, { useEffect } from 'react';
+
+interface TabKeybindingsProps<T> {
+  tabs: T[];
+  onAdd?: () => void;
+  setCurrentTab: React.Dispatch<React.SetStateAction<T>>;
+}
+
+export function TabKeybindings<T>({
+  tabs,
+  onAdd,
+  setCurrentTab,
+}: TabKeybindingsProps<T>) {
+  const keybindings = (e: KeyboardEvent) => {
+    if (e.ctrlKey) {
+      tabs.forEach((tab, n) => {
+        if (e.key === `${n + 1}`) {
+          e.preventDefault();
+          setCurrentTab(tab);
+        }
+      });
+    }
+    if (onAdd) {
+      if (e.key === '+') {
+        e.preventDefault();
+        onAdd();
+      }
+    }
+  };
+
+  useEffect(() => {
+    window.addEventListener('keydown', keybindings);
+    return () => window.removeEventListener('keydown', keybindings);
+  }, []);
+
+  return <></>;
+}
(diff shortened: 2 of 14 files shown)


packages/common/src/intl/locales/en/inventory.json
```diff
@@ -1,5 +1,5 @@
 {
-  "button.delete-lines": "Delete lines",
+  "button.delete-lines": "Delete selected lines",
```
mark-prins, line 2: I see that this was inconsistent, so have updated them

packages/common/src/ui/components/navigation/Tabs/TabKeybindings.tsx
```diff
@@ -0,0 +1,37 @@
+import React, { useEffect } from 'react';
```
mark-prins, line 1: lifted this out to a component.. because it was fun to do 🤷 
It does make it easier to keep them consistent, and is slightly easier to add new tabs in to the key bindings

packages/common/src/ui/components/navigation/Tabs/Tabs.tsx
```diff
@@ -19,7 +19,6 @@ export const TabList = styled(MuiTabList)({
 export const Tab = styled(MuiTab)({
   color: '#8f90a6',
   fontSize: '12px',
-  width: '100px',
```
mark-prins, line 22: not sure what this will break 🤷 
The wider tabs in inbound shipments `Quantities (Ctrl+1)` were not fitting in!

packages/inventory/src/Stocktake/DetailView/modal/StocktakeLineEdit/StocktakeLineEdit.tsx
```diff
@@ -65,6 +65,10 @@ export const StocktakeLineEdit: FC<StocktakeLineEditProps> = ({
     }
   };
 
+  const hasValidBatches = draftLines.some(
```
mark-prins, line 68: prevent saving when there are no valid batches.. which I _think_ was the problem raised

mark-prins:
Found a couple of extra issues
- Unable to enter values on the internal order edit modal for suggested quantity greater than 999
- When suggested quantity is non-zero, the stock evolution chart is adding the requested quantity, but the values already have suggested quantity added, with a hard-coded delivery lead time

packages/common/src/ui/components/navigation/Tabs/TabKeybindings.tsx
```diff
@@ -0,0 +1,37 @@
...
+  tabs,
+  onAdd,
+  setCurrentTab,
+}: TabKeybindingsProps<T>) {
+  const keybindings = (e: KeyboardEvent) => {
+    if (e.ctrlKey) {
+      tabs.forEach((tab, n) => {
+        if (e.key === `${n + 1}`) {
+          e.preventDefault();
+          setCurrentTab(tab);
```
clemens-msupply, line 19: Maybe add a comment with what the key bindings are and what they suppose to do (for slow people like me who take longer than needed to parse the code ;-))

clemens-msupply (approved):
Looks good! thanks

packages/common/src/ui/components/navigation/Tabs/TabKeybindings.tsx
mark-prins: Good point - thanks. have done 😁

PR #1186, opened 2022-04-19
1185 permission issues

### LINKED ISSUE #1185: Permissions Issues
With the addition of permission checking across all endpoints there are some issues to sort out

- [x] While on the login screen, if logged out, you get a number of toasts. I think it should only show the logged out modal

https://user-images.githubusercontent.com/9192912/163907737-e6fa8023-579a-40bd-9e7d-3b55be2cd87e.mp4

- [x] The dashboard queries throw an error if you don't have view invoice permission. Would be nice if it handled those quietly
- [x] The _reports_ endpoint gives an error for the admin user, which results in the modal for all things, when it should only be affecting reports 🙄 

Screen Shot 2022-04-19 at 9 30 55 AM

mark-prins:
Fixes #1185 
Checking for undefined responses mainly - also not toasting when the user is logged out, since that is handled by the Auth Modal

### DIFF
--- a/packages/common/src/api/GqlContext.tsx
+++ b/packages/common/src/api/GqlContext.tsx
@@ -9,7 +9,7 @@ import {
 import { AuthError } from '../authentication/AuthContext';
 import { LocalStorage } from '../localStorage';
 
-const permissionExceptions = ['reports'];
+const permissionExceptions = ['reports', 'stockCounts', 'invoiceCounts'];
 interface ResponseError {
   message?: string;
   path?: string[];
--- a/packages/common/src/api/GqlContext.tsx
+++ b/packages/common/src/api/GqlContext.tsx
@@ -9,15 +9,20 @@ import {
 import { AuthError } from '../authentication/AuthContext';
 import { LocalStorage } from '../localStorage';
 
+const permissionExceptions = ['reports'];
 interface ResponseError {
   message?: string;
   path?: string[];
   extensions?: { details?: string };
 }
 
-function hasError(errors: ResponseError[], error: AuthError) {
-  return errors.some(({ message }: { message?: string }) => message === error);
-}
+const hasError = (errors: ResponseError[], error: AuthError) =>
+  errors.some(({ message }: { message?: string }) => message === error);
+
+const hasPermissionException = (errors: ResponseError[]) =>
+  errors.every(({ path }: { path?: string[] }) =>
+    (path || []).every(p => permissionExceptions.includes(p))
+  );
 
 const handleResponseError = (errors: ResponseError[]) => {
   if (hasError(errors, AuthError.Unauthenticated)) {
@@ -25,7 +30,10 @@ const handleResponseError = (errors: ResponseError[]) => {
     return;
   }
 
-  if (hasError(errors, AuthError.PermissionDenied)) {
+  if (
+    hasError(errors, AuthError.PermissionDenied) &&
+    !hasPermissionException(errors)
+  ) {
     LocalStorage.setItem('/auth/error', AuthError.PermissionDenied);
     return;
   }
--- a/packages/common/src/authentication/api/api.ts
+++ b/packages/common/src/authentication/api/api.ts
@@ -86,7 +86,7 @@ export const getAuthQueries = (sdk: Sdk) => ({
     },
     stores: () => async () => {
       const result = await sdk.me();
-      return result.me?.stores?.nodes;
+      return result?.me?.stores?.nodes;
     },
   },
 });
--- a/packages/common/src/intl/locales/en/common.json
+++ b/packages/common/src/intl/locales/en/common.json
@@ -35,6 +35,7 @@
   "error.no-data": "No data available",
   "error.no-items": "No items",
   "error.no-items-filter-on": "No items to display. Try changing the filter criteria.",
+  "error.no-report-permission": "No permission to view reports",
   "error.no-reports-available": "No reports available",
   "error.no-results": "No records to display",
   "error.something-wrong": "Oops! Something's gone wrong.",
--- a/packages/common/src/ui/components/popover/PaperPopover/PaperPopoverSection.tsx
+++ b/packages/common/src/ui/components/popover/PaperPopover/PaperPopoverSection.tsx
@@ -4,14 +4,26 @@ import Typography from '@mui/material/Typography';
 
 export interface PaperPopoverSectionProps {
   label?: string;
+  labelStyle?: React.CSSProperties;
+  alignItems?: 'center' | 'flex-start' | 'flex-end' | 'stretch';
 }
 
 export const PaperPopoverSection: FC<PaperPopoverSectionProps> = ({
   children,
   label,
+  labelStyle,
+  alignItems,
 }) => (
-  <Box gap={2} p={3} flexDirection="column" display="flex">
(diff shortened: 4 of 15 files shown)


packages/common/src/api/GqlContext.tsx
```diff
@@ -9,23 +9,31 @@ import {
...
   extensions?: { details?: string };
 }
 
-function hasError(errors: ResponseError[], error: AuthError) {
-  return errors.some(({ message }: { message?: string }) => message === error);
-}
+const hasError = (errors: ResponseError[], error: AuthError) =>
+  errors.some(({ message }: { message?: string }) => message === error);
+
+const hasPermissionException = (errors: ResponseError[]) =>
```
mark-prins, line 22: didn't especially like adding in the exceptions - though if the check fails, it will default to showing errors.

packages/common/src/authentication/api/api.ts
```diff
@@ -86,7 +86,7 @@ export const getAuthQueries = (sdk: Sdk) => ({
     },
     stores: () => async () => {
       const result = await sdk.me();
-      return result.me?.stores?.nodes;
+      return result?.me?.stores?.nodes;
```
mark-prins, line 89: there's a few instances of this sort of thing. If the query throws an error and the data returned is undefined, we want to handle that state rather than throwing an unhelpful error at this level.

packages/common/src/intl/locales/en/common.json
```diff
@@ -35,6 +35,7 @@
   "error.no-data": "No data available",
   "error.no-items": "No items",
   "error.no-items-filter-on": "No items to display. Try changing the filter criteria.",
+  "error.no-report-permission": "No permission to view reports",
```
mark-prins, line 38: there's separate handling for the lack of print permission - that one will show the permission modal, which is good. This one is different, and shows that you can't even view a list of reports

packages/common/src/ui/components/popover/PaperPopover/PaperPopoverSection.tsx
```diff
@@ -4,14 +4,26 @@ import Typography from '@mui/material/Typography';
...
 
 export const PaperPopoverSection: FC<PaperPopoverSectionProps> = ({
   children,
   label,
+  labelStyle,
+  alignItems,
 }) => (
-  <Box gap={2} p={3} flexDirection="column" display="flex">
-    <Typography fontWeight="700">{label}</Typography>
+  <Box
```
mark-prins, line 17: had to extend this component slightly, in order to center the loading spinner.

packages/dashboard/src/widgets/StockWidget.tsx
```diff
@@ -21,6 +21,8 @@ export const StockWidget: React.FC = () => {
   const formatNumber = useFormatNumber();
   const { data: expiryData, isLoading: isExpiryLoading } = useStockCounts();
   const { data: itemStatsData, isLoading: isItemStatsLoading } = useItemStats();
+  const [hasExpiryError, setHasExpiryError] = React.useState(false);
```
mark-prins, line 24: set a local state error and don't show stats if you don't have permission to view them.
Could display text, but I thought in this case, better to just quietly hide the stats.

packages/host/src/QueryErrorHandler.tsx
```diff
@@ -11,9 +15,10 @@ export const QueryErrorHandler = () => {
   const [errorMessage, setErrorMessage] = useState<string | null>(null);
   const location = useLocation();
   const generalError = t('error.general-query-error');
+  const [authError] = useLocalStorage('/auth/error');
 
   useEffect(() => {
-    if (!!errorMessage) {
+    if (!!errorMessage && authError !== AuthError.Unauthenticated) {
```
mark-prins, line 21: The `Unauthenticated` error is handled with a Modal, so there's no need to toast

packages/system/src/Report/api/api.ts
```diff
@@ -31,11 +31,11 @@ export const getReportQueries = (sdk: Sdk, storeId: string) => ({
         reportId,
         storeId,
       });
-      if (result.printReport.__typename === 'PrintReportNode') {
+      if (result?.printReport.__typename === 'PrintReportNode') {
         return result.printReport.fileId;
       }
 
-      throw new Error('Report not found');
+      throw new Error('Unable to print report');
```
mark-prins, line 38: ok for now. ideally this should be translated.. but that's a bit hard in the api. fairly unlikely to be shown

packages/system/src/Report/api/hooks.ts
```diff
@@ -40,8 +40,15 @@ export const useReports = (category?: ReportCategory) => {
     initialListParameters
   );
 
-  return useQuery(api.keys.paramList(queryParams), async () =>
-    api.get.list({ filterBy, sortBy, offset })
+  return useQuery(
+    api.keys.paramList(queryParams),
+    async () => api.get.list({ filterBy, sortBy, offset }),
+    {
+      onError: (e: Error) => {
```
mark-prins, line 47: yeah. not ideal. Just wanted to be able to inform the user that they don't have permission

packages/common/src/authentication/api/api.ts
clemens-msupply: Why is it null? isn't the me return type Promise<MeQuery>?, i.e. never null? or do I miss something?

packages/system/src/Report/api/hooks.ts
clemens-msupply: Yes have to think about how to manage/define the report permissions...

clemens-msupply (approved):
Could you explain again why the `result?.` checks are needed? from the TS type it looks they can't be null/undefined?

packages/common/src/authentication/api/api.ts
mark-prins: We're using [react-query](https://react-query.tanstack.com/reference/useQuery) for the handling of return objects at the react component level and that is imposing it's own values.

The query method does indeed return a promise, in this case, `Promise<MeQuery>` - but only on success is the data updated. When there is an error, the default `data` value of `undefined` remains unchanged.

packages/common/src/intl/locales/en/common.json
mark-prins: oh.. that's interesting! I was as one point able to query for reports.. but was getting a `Forbidden` response when printing! no idea how that is possible.. as you say, it's just the `Report` permission. If that ever does happen, it is handled nicely 😁  mostly though, users might not have permission to view reports, in which case they'll get the print button and a message saying that they don't have permission to view reports.

packages/common/src/authentication/api/api.ts
clemens-msupply: Ok thanks, so the correct type is `Promise<MeQuery | undefined>`? think we can update the type, can't we?

i.e.
```
export type SdkFunctionWrapper = <T>(action: (requestHeaders?:Record<string, string>) => Promise<T>, operationName: string) => Promise<T | undefined>;
```

mark-prins: that export is in the generated file though, and I haven't found an option to modify the result to be potentially undefined. I could do `const result = await sdk.me() as MeQuery | undefined;` and that will pick up the type errors if the undefined isn't handled. I don't like the retyping, but maybe that's better than the alternative

PR #1190, opened 2022-04-21
1189 fix tests

### LINKED ISSUE #1189: Update tests for React v18
Following on from #1188 - the tests now require some attention

mark-prins:
Fixes #1189 

Tests are now working!
Have left storybook - react v18 support is only in the alpha release, and since that only gives warnings at this stage, have just left it. We can update that later.


jest.config.js
```diff
@@ -12,10 +12,6 @@ module.exports = {
   setupFilesAfterEnv: ['<rootDir>/jest-setup.ts'],
   moduleNameMapper: {
     kbar: '<rootDir>/__mocks__/kbar.ts',
-    // ref: https://react-hooks-testing-library.com/installation#being-specific
```
mark-prins, line 15: no longer required! the `@testing-library/react` now has a `renderHook` which supports v18. there's no plans to update `@testing-library/react-hooks` to v18 so there's little choice really

packages/common/src/hooks/useAppBarRect/useAppBarRect.test.tsx
```diff
@@ -39,9 +40,18 @@ describe('useAppBarRect', () => {
...
+
+    return <span>{renderCounter.current}</span>;
+  };
+
   it('Triggers a new state update when the window dimensions change', () => {
-    const { result } = renderHook(useAppBarRect);
+    const { getByText } = render(<Component />);
 
-    expect(result.all.length).toBe(2);
+    const hasRenderedTwice = getByText(/2/);
```
mark-prins, line 54: one thing missing from the new `renderHook` is a history of renders. I've had to improvise.

packages/common/src/intl/context/IntlContext.tsx
```diff
@@ -37,7 +40,7 @@ export const IntlProvider: FC<PropsWithChildrenOnly> = ({ children }) => {
             },
           ],
         },
-        debug: true,
+        debug: isDevelopment,
```
mark-prins, line 43: thought that this was a cute idea 🤷

packages/common/src/intl/context/IntlContext.tsx
```diff
@@ -10,12 +10,15 @@ import { PropsWithChildrenOnly } from '@common/types';
 const defaultNS = 'common';
 export const IntlProvider: FC<PropsWithChildrenOnly> = ({ children }) => {
   React.useEffect(() => {
+    if (i18next.isInitialized) return;
```
mark-prins, line 13: this isn't necessary now - but left it in. I had an issue because I was running `init` in the testing provider which then screwed up the site 🙄

packages/common/src/utils/testing/testing.tsx
```diff
@@ -54,20 +50,24 @@ export const IntlTestProvider: FC<PropsWithChildren<IntlTestProviderProps>> = ({
   locale,
 }) => {
   useEffect(() => {
+    i18next.changeLanguage(locale);
```
mark-prins, line 53: reworked slightly - the new `Suspense` is suspending due to `i18next` and the `useEffect` only fires when not suspended. It means that with the original version, i18next would never `init`

This way, it can init on load, and then when the props come down, the language can be set correctly.

packages/host/src/components/AppDrawer/AppDrawer.test.tsx
```diff
@@ -83,18 +77,16 @@ describe('AppDrawer', () => {
...
 
-    waitFor(() => {
+    await waitFor(() => {
       expect(getByText(/dashboard/i)).not.toBeVisible();
       expect(getByText(/distribution/i)).not.toBeVisible();
       expect(getByText(/suppliers/i)).not.toBeVisible();
-      expect(getByText(/stock/i)).not.toBeVisible();
-      expect(getByText(/tools/i)).not.toBeVisible();
+      expect(getByText(/stock$/i)).not.toBeVisible();
+      // expect(getByText(/tools/i)).not.toBeVisible();
```
mark-prins, line 87: these were failing - but not noticed because we weren't waiting for them. Adding the `await` fixes an issue with the test runner, and highlighted the failing tests

packages/inventory/src/Stocktake/ListView/Toolbar.tsx
```diff
@@ -5,19 +5,33 @@ import {
   useTranslation,
   DeleteIcon,
   AppBarContentPortal,
-  SearchBar,
   FilterController,
+  StocktakeNodeStatus,
```
mark-prins, line 9: offroad! 😁 

The filter for stocktakes has changed, and no longer supports filtering by comment..
I've popped in a quick filter by status instead (nothing else available looked useful) - though the API only supports `equalTo` so I couldn't implement a multiple option version. not a problem here, since there's only two status values 😆

packages/host/src/components/AppDrawer/AppDrawer.test.tsx
clemens-msupply: There is a linter rule "no-floating-promises" which looks for these kind of bugs :)

packages/common/src/utils/testing/testing.tsx
clemens-msupply: Just skimmed through React suspend docs, its a bit magically isn't it? i.e. it reads like it interrupts/suspends a non-async method? probably it does some other trickery though, do you have some deeper understanding on how it is implemented?

clemens-msupply (approved):
Thanks looks good (skimmed through most changes)

packages/common/src/utils/testing/testing.tsx
mark-prins: It is a bit tricksy. We were already using it (pre release version was in react v17), so I was surprised to have i18next handle it so differently

packages/host/src/components/AppDrawer/AppDrawer.test.tsx
mark-prins: thanks! wasn't aware of that one.

We are using `plugin:@typescript-eslint/recommended` which includes the [recommended rules](https://github.com/typescript-eslint/typescript-eslint/tree/main/packages/eslint-plugin/docs/rules) which in turn means we are already using that rule.. and it didn't pick up this particular problem. The issue with the test is that it is checking for an element which won't exist, because we removed that menu item!

PR #1195, opened 2022-04-27
Server sync settings

### LINKED ISSUE #1193: Server sync config
This PR (https://github.com/openmsupply/remote-server/pull/1044
) contains all required changes to do the initial sync config, e.g. config sync settings and restart the server:

The new api can also be used to figure out if the server is running normally or if it is in "stage0" config mode. In stage0 mode no bearer token is needed.

The server config and the restart endpoints are the same in both modes, e.g. same schema could be used...

For example,  when starting the client:
- Check if server is in stage0 mode and show a config screen to update sync settings and restart the server
- In normal mode the same config screen could be used to update the settings and restart the server.

mark-prins:
Fixes #1193 

Adds a 'synchronisation' section to the admin page which allows updating of all the sync settings. On save the server is restarted. 

Screen Shot 2022-04-28 at 9 01 48 AM

Thought for a while about polling for server status on restart, thinking that I'll implement that at a higher level and integrate a server offline message. 

Have included basic permission handling - if you do not have the server settings permission then you are shown the permission denied modal and the sync settings form is disabled. Also took the opportunity to refactor the hooks in this area, along the lines that Josh has done with Outbounds. Intending to change all hooks to that pattern when time permits.


packages/host/src/Admin/SyncSettings.tsx
```diff
@@ -0,0 +1,194 @@
...
+export const SyncSettings = ({}) => {
+  const { data, isLoading, isError } = useHost.utils.settings();
+  const { mutate, isLoading: isSaving } = useHost.sync.update();
+  const { mutateAsync: serverRestart } = useHost.utils.restart();
+  const t = useTranslation('common');
+  const [syncSettings, setSyncSettings] = useState<SyncSettings | null>(null);
+  const { success, info } = useNotification();
+
+  const currentSettings = {
+    centralServerSiteId: data?.syncSettings?.centralServerSiteId || 1,
```
mark-prins, line 149: yes, all this to exclude the `__typename`

packages/host/src/Admin/SyncSettings.tsx
```diff
@@ -0,0 +1,194 @@
...
+    ? { ...currentSettings, ...syncSettings }
+    : undefined;
+
+  const onSave = () => {
+    if (!syncSettings) return;
+    const successSnack = success(t('success.sync-settings'));
+    const restart = async () => {
+      successSnack();
+      await serverRestart(); // returns 'Restarting'
+      const infoSnack = info(t('info.server-restarting'));
```
mark-prins, line 167: discarding the 'Restarting' message in order to provide a translatable message. Should check for the response but didn't want to be limiting ourselves to a hard coded string at this stage (I'm not confident that the API response won't change at some point!) so went with the slightly looser approach and will 'info' when the request is sent

packages/host/src/Admin/SyncSettings.tsx
```diff
@@ -0,0 +1,194 @@
...
+      <Setting
+        title={t('label.settings-username')}
+        component={<StringSyncSetting property="username" />}
+      />
+      <Setting
+        title={t('label.settings-password')}
+        component={
+          <StringSyncSetting
+            property="password"
+            type="password"
```
mark-prins, line 107: type="password" to obscure the input chars, and autocomplete to keep chrome happy ( raises a console warning otherwise )

packages/host/src/Admin/SyncSettings.tsx
```diff
@@ -0,0 +1,194 @@
...
+      <NumericTextInput
+        value={value}
+        onChange={e => onChange(e.target.value)}
+        disabled={isDisabled}
+      />
+    );
+  };
+
+  return (
+    <form style={{ width: '100%' }}>
```
mark-prins, line 93: the form is just to prevent a console warning in chrome, due to the password input

packages/host/src/api/hooks/sync/useSyncSettingsUpdate.ts
```diff
@@ -0,0 +1,22 @@
...
+  UpdateSyncSettingsInput,
+  useMutation,
+  useQueryClient,
+} from '@openmsupply-client/common';
+import { useHostApi } from '../utils/useHostApi';
+
+export const useSyncSettingsUpdate = () => {
+  const queryClient = useQueryClient();
+  const api = useHostApi();
+  return useMutation<
```
mark-prins, line 12: explicitly setting the typing on `useMutation` calms typescript down a bit. Usually it's just variables and return type that I'm concerned with

packages/inventory/src/Stocktake/api/operations.generated.ts
```diff
@@ -4,11 +4,11 @@ import { GraphQLClient } from 'graphql-request';
 import * as Dom from 'graphql-request/dist/types.dom';
 import gql from 'graphql-tag';
 import { graphql, ResponseResolver, GraphQLRequest, GraphQLContext } from 'msw'
-export type StocktakeRowFragment = { __typename: 'StocktakeNode', id: string, comment?: string | null, description?: string | null, createdDatetime: any, finalisedDatetime?: any | null, stocktakeDate?: string | null, stocktakeNumber: number, status: Types.StocktakeNodeStatus, isLocked: boolean };
+export type StocktakeRowFragment = { __typename: 'StocktakeNode', id: string, comment?: string | null, description?: string | null, createdDatetime: string, finalisedDatetime?: string | null, stocktakeDate?: string | null, stocktakeNumber: number, status: Types.StocktakeNodeStatus, isLocked: boolean };
```
mark-prins, line 7: a number of changes picked up with the codegen - which is why I thought to do this part of the sync settings first. will do the first use page next once all this plumbing is in place

packages/host/src/Admin/SyncSettings.tsx
```diff
@@ -0,0 +1,194 @@
...
+import { useHost } from '../api/hooks';
+import { Setting } from './Setting';
+
+type SyncSettings = Omit<UpdateSyncSettingsInput, '__typename'>;
+
+const isValid = (syncSettings: SyncSettings | null) => {
+  if (!syncSettings) return false;
+  return (
+    !!syncSettings.url &&
+    !!syncSettings.username &&
```
clemens-msupply, line 23: The api actually checks if the url is a valid url and if the username is not empty... Otherwise I don't do any validation

packages/host/src/api/api.ts
```diff
@@ -1,10 +1,27 @@
...
 
 export const getHostQueries = (sdk: Sdk) => ({
   get: {
-    version: () => async () => {
+    settings: async () => {
+      const result = await sdk.serverSettings();
+      return result.serverSettings;
+    },
+    version: async () => {
       const result = await sdk.apiVersion();
```
clemens-msupply, line 11: Think I mentioned it already, probably a good idea to merge the server status, sync progress, api version into a `/status` endpoint...

clemens-msupply (approved):
Thanks very exciting. One thing I noticed is that the text fields lose focus (Firefox) and key strokes are interpreted as keyboard bindings, e.g. typing `m[o]biltest` jumps to "Outbound shipments".

Issues on the server:
- the token bucket with existing tokens should be kept across restarts otherwise users have to re-authenticate
- when starting with settings from the config these settings are always preferred over the DB settings and restart doesn't use the new settings -> always use DB settings if present and it was a restart for now?

packages/host/src/Admin/SyncSettings.tsx
mark-prins: that's right.. I remember seeing that in the PR and was going to make use of it here. then forgot. Should perhaps add URL validation here.. but I rather like the server validating and having that in one place. I've just managed to save an invalid URL and created chaos with the server falling back to bootstrap mode! 🤣

packages/host/src/api/api.ts
mark-prins: would be neat, yeah. I can see that being very useful

mark-prins:
> One thing I noticed is that the text fields lose focus

Just noticed that! I think if the server settings is undefined then the unstable reference is causing re-renders. Will tidy that up.

> typing `m[o]biltest` jumps to "Outbound shipments".

Oh no! 😱  Hadn't seen that.. might just be firefox, will check

packages/host/src/Admin/SyncSettings.tsx
clemens-msupply: :) it just checks if the URL is well formatted, I don't try to connect to it...

PR #1196, opened 2022-04-28
Initialisation

### LINKED ISSUE #1194: Bootstrap mode
This issue should be tackled after #1193 is in place - as the api methods would then be available.
Create a bootstrap / first use page which allows sync configuration for first use

From the login screen:
- Check the server mode
- If bootstrap, redirect to a first use screen
- Similar form that required by #1193, so maybe we can reuse some of that?
- on save, restart the server then redirect to `/login` again

mark-prins:
Fixes #1194 

Adds an initialisation screen and handles the sync config settings.
There's a few things which I'd like to improve - this will at least get things working


packages/common/src/authentication/AuthContext.tsx
```diff
@@ -218,8 +221,12 @@ export const AuthProvider: FC<PropsWithChildrenOnly> = ({ children }) => {
     const timer = window.setInterval(() => {
       const authCookie = getAuthCookie();
       const { token } = authCookie;
+      const isInitScreen = matchPath(
```
mark-prins, line 224: added this in because the only way I could see of determining that the sync settings had been applied and the server was back up and running was to poll for the `Unauthenticated` response to the settings query. 

Without this little addition, the cookie is checked, the auth error set and the polling returns too early 🙄

packages/config/src/routes.ts
```diff
@@ -1,4 +1,5 @@
 export enum AppRoute {
+  Initialise = 'initialise',
```
mark-prins, line 2: added a route for this - there's one strangeness involved.. you can navigate to that URL manually at any time.
This won't work because the queries would then require authentication if the server isn't in bootstrap mode - thought to handle that later after we've refined the status queries.

packages/host/src/api/api.ts
```diff
@@ -5,7 +5,7 @@ export const getHostQueries = (sdk: Sdk) => ({
   get: {
     settings: async () => {
       const result = await sdk.serverSettings();
-      return result.serverSettings;
+      return result?.serverSettings;
```
mark-prins, line 8: so that I could get the original error if there was one, rather than throwing an error here due to `result` being undefined. Not actually used now, but still a good check to have in.

packages/host/src/components/Initialise/Initialise.tsx
```diff
@@ -0,0 +1,117 @@
...
+      UrlInput={
+        <LoginTextInput
+          fullWidth
+          label={t('label.settings-url')}
+          value={url}
+          disabled={isLoading}
+          onChange={e => setUrl(e.target.value)}
+        />
+      }
+      SiteIdInput={
```
mark-prins, line 75: hoping that we could remove this at some point - so haven't gone very fancy with the numeric input handling.

packages/host/src/components/Login/Login.tsx
```diff
@@ -31,6 +34,16 @@ export const Login: React.FC = ({}) => {
     error,
   } = useLoginForm(passwordRef);
 
+  useEffect(() => {
+    setPageTitle(`${t('app.login')} | ${t('app')} `);
+  }, []);
+
+  useEffect(() => {
```
mark-prins, line 41: redirect from login to init if the server is in bootstrap mode

packages/host/src/components/hooks.ts
```diff
@@ -0,0 +1,129 @@
...
+    error,
+    setError,
+    url,
+    siteId,
+  } = state;
+  const [isPolling, setIsPolling] = useState(false);
+  const [isBootstrap, setIsBootstrap] = useState(false);
+  const { mutateAsync: update } = useHost.sync.update();
+  const { data } = useHost.utils.settings({
+    refetchInterval: 5000,
```
mark-prins, line 63: the `refetchInterval` will have the query requery every 5s. Originally had it as 1s but worried that the hits to the server might have slowed the restart and it was taking over a minute anyway, so a few extra seconds didn't matter. Wonder if it should be reduced again though.

packages/host/src/components/hooks.ts
```diff
@@ -0,0 +1,129 @@
...
+    setPassword('');
+
+    await restart().catch(e => {
+      console.error(e);
+      setError({ message: 'Unable to restart the server' });
+      setIsLoading(false);
+      return;
+    });
+
+    setIsPolling(true);
```
mark-prins, line 96: enables the polling query - this then queries the server status regularly and returns valid responses for a while, then errors out for some while the server is unavailable, then returns a response with an error while the server syncs, and finally gives an authentication error when the server is configured and running.

auth errors are handled by the GqlContext which sets the local storage error - so am listening for that and redirecting when that comes in.

packages/host/src/components/hooks.ts
```diff
@@ -0,0 +1,129 @@
...
+      console.log('Timed out', data);
+      const message = isBootstrap
+        ? 'Unable to sync! Please check your settings.'
+        : 'Server restart has timed out';
+      setError({ message });
+    }, SERVER_RESTART_TIMEOUT);
+  };
+
+  const isValid = !!username && !!password && !!url;
+  useEffect(() => {
```
mark-prins, line 120: the authentication listener might come in while the data is being fetched, so I wanted to track the state of the most recent status value. I think we can tidy up all of the status / polling aspects

packages/host/src/components/hooks.ts
```diff
@@ -0,0 +1,129 @@
...
+      if (key === '/auth/error' && value === AuthError.Unauthenticated) {
+        // Server is up! and rejecting our request!
+        setIsLoading(false);
+        setIsPolling(false);
+
+        navigate(`/${AppRoute.Login}`, { replace: true });
+      }
+    });
+
+    setTimeout(() => {
```
mark-prins, line 108: to prevent it polling infinitely. With the sample datafile and running locally with sqlite, the server reset and sync takes more than a minute, so I've had to stretch out the timeout value to 90s to cater for it. might still timeout.. and if so, should probably give users the option of refreshing the page

packages/host/src/components/Initialise/hooks.ts
```diff
@@ -0,0 +1,130 @@
...
+    error,
+    setError,
+    url,
+    siteId,
+  } = state;
+  const [isPolling, setIsPolling] = useState(false);
+  const [isBootstrap, setIsBootstrap] = useState(false);
+  const { mutateAsync: update } = useHost.sync.update();
+  const { data } = useHost.utils.settings({
+    refetchInterval: POLLING_INTERVAL,
```
mark-prins, line 65: the `refetchInterval` sets the query to re-query regularly. This way I can check to see if the server is responding with a configured and running response or not.
The `enabled` flag does what it sounds like. Have set that to a state value, so that it can be enabled after the server is updated and restarted.

packages/host/src/components/Initialise/hooks.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    await restart().catch(e => {
+      console.error(e);
+      setError({ message: 'Unable to restart the server' });
+      setIsLoading(false);
+      return;
+    });
+
+    setIsPolling(true);
+    LocalStorage.setItem('/auth/error', undefined);
```
mark-prins, line 99: the query originally returns a valid response while shutting down; after a while it doesn't return anything, then returns an error and finally will return an authentication error, as the user isn't logged in.

The auth error is handled in GqlContext which sets an auth error on local storage - have set a listener for that and will wait for that response before redirecting. at this point the user can login.

With a demo datafile and running sqlite locally, it takes more than a minute for the whole process - so have set 90s as the timeout

packages/host/src/components/Initialise/hooks.ts
```diff
@@ -0,0 +1,130 @@
...
+        setIsPolling(false);
+
+        navigate(`/${AppRoute.Login}`, { replace: true });
+      }
+    });
+
+    setTimeout(() => {
+      setIsLoading(false);
+      setIsPolling(false);
+      const message = isBootstrap
```
mark-prins, line 113: the data could be undefined when the server is ready ( pending response ) and so I want to store the most recent status value for use in choosing which error to show the user. If the sync values are incorrect and the server has reset to bootstrap mode, we want to tell the user to check the settings before trying again.

packages/host/src/components/Initialise/hooks.ts
```diff
@@ -0,0 +1,130 @@
...
+  AuthenticationError,
+  AuthError,
+  LocalStorage,
+  ServerStatus,
+  useNavigate,
+} from '@openmsupply-client/common';
+import { useHost } from '../../api/hooks';
+
+const SERVER_RESTART_TIMEOUT = 90000;
+const POLLING_INTERVAL = 3000;
```
mark-prins, line 14: polling every 3s : originally was 1s but was concerned that the queries were delaying the server restart process! Given how long the whole process takes, thought that a few seconds extra doesn't matter

packages/host/src/components/Initialise/Initialise.tsx
```diff
@@ -0,0 +1,117 @@
...
+          inputProps={{
+            autoComplete: 'username',
+          }}
+          autoFocus
+        />
+      }
+      PasswordInput={
+        <LoginTextInput
+          fullWidth
+          label={t('label.settings-password')}
```
clemens-msupply, line 56: Got confused here and entered the user credentials and not the central server account. Maybe better to include this into the label (maybe also helpful for the URL).

Maybe just me, but I also ignored the left part of the screen with the description of the init page, probably assuming that it doesn't contain information about the screen but just general mSupply info. Maybe move the description into the form to make it clearer that this is the initial setup screen? Maybe also a big header like "Initial server setup" and/or different colour for the init screen....

packages/host/src/components/Initialise/hooks.ts
clemens-msupply: I just played with it and increasing the value makes it faster, the query seems to slow down the server from shutting down gracefully...

Think best would be to wait some time before start polling. Think without ongoing connections the server restart is actually quite fast. I tried:

```js
    setTimeout(() => {
      setIsPolling(true);
    }, 6000);
```
and this was ok

packages/host/src/components/Initialise/hooks.ts
clemens-msupply: Ok maybe we should have public /status endpoint containing general information, e.g. api version and server status so that it can be polled without permissions?

I guess for sync status permissions would be required?

clemens-msupply (approved):
Very cool! as already discussed some more work required on the server side to make it really smooth, e.g. if there are settings errors or report sync progress...

I inlined some UI suggestions

packages/host/src/components/Initialise/hooks.ts
mark-prins: an open endpoint which returns status & version would be great - we could remove some of the hackyness here 😄 
and I think is safe enough to have publicly available. Should be no need to have sync status public that I can see, though some aspect of it as part of the status would be needed - the client needs to know if the user can log in, which requires sync to have completed.

So either

```
{
  version: String
  status: ServerStatus
  sync: Active | Idle
}
```

or

```
export enum ServerStatus {
  Running = 'RUNNING',
  /** Server misses configuration to start up fully */
  Stage_0 = 'STAGE_0',
  Synchronising = 'SYNCHRONISING'
}
```

or 🤷

packages/host/src/components/Initialise/hooks.ts
mark-prins: 😁 ah, thanks. I wondered about doing that and didn't actually test it. Have added it in

mark-prins:
I've renamed the fields and changed the colour of the page - but couldn't find a way to make a title on the form look good. The page is also too long with all those fields.. so had to drop that idea. had made the logo smaller to accomodate the title and have left that change in. hopefully enough?

PR #1198, opened 2022-04-29
Stabilise translation hook

### LINKED ISSUE #884: The reference to `t` from `useTranslation is unstable
The following code will run in an infinite loop

```
const [ state, setState ] = useState({})
const t = useTranslation();

useEffect(() => {
    setState({})
}, [t])
```

which is not ideal! Ideally the reference for `t` would remain stable. This will likely cause some subtle and hard to find bugs and performance implications in the future

mark-prins:
fixes #884 

well, that was simpler than expected.

## Testing

Simple test, from the issue itself. Just add

```
  const [state, setState] = React.useState(0);
  const t = useTranslation();

  useEffect(() => {
    console.log(`** RENDER ${state} **`);
    setState(state + 1);
  }, [t]);
```

to a page and watch the fun.

### DIFF
--- a/packages/common/src/intl/strings/IntlStrings.ts
+++ b/packages/common/src/intl/strings/IntlStrings.ts
@@ -1,3 +1,4 @@
+import { useCallback } from 'react';
 import { Namespace, useTranslation as useTranslationNext } from 'react-i18next';
 import { TOptions } from 'i18next';
 import { LocaleKey } from '../locales';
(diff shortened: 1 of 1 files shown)


clemens-msupply (approved):
Looks good

Reason is that useCallback doesn't change until t changes but the returned arrow function changes on ever call, right?

mark-prins:
yes, exactly right. turns out that the t function from i18next is stable if you destructure like we have but our hook was recreating a function on every render 🙄

PR #1199, opened 2022-04-29
Handle undefined data

### LINKED ISSUE #1197: Authentication Error !!!
On a Central server, Intentionally did not kept the sync user details correct inside `configuration ->> base.yaml`.
Started remote server using `cargo run` although the `omsupply-database` was introduced which then failed to push the records onto the remote server, end result was something like that:

` Caused by:│@openmsupply-client/host:   ../config/src/index.ts 2.54 KiB [built] [code generated] HTTP status client error (401 Unauthorized) for url (http://localhost:2048/sync/│@openmsupply-client/host:   ../config/src/routes.ts 2.41 KiB [built] [code generated] v5/queued_records?limit=500)`

The 4d Primary server was running, with the frontend, Authenticated with the valid username and password, that got authenticated correctly takes inside the frontend with saying `Authentication Error !!!`

Hope this make sense to describe the issue i am seeing.

[Authentication Error Video](https://www.dropbox.com/s/m5rd3xuvzafsj7x/openmSupply%20issue%20authentication.mp4?dl=0)

mark-prins:
Fixes #1197 

The client part anyway - there's an issue with the server allowing authentication when I believe it shouldn't. See https://github.com/openmsupply/remote-server/issues/1061 for status on that one.

In this PR, have removed all of the additional errors, and therefore error toasts, that Bijay was observing. Now, from a client point of view, things are behaving rationally - the only extra step is preventing the login.

### DIFF
--- a/packages/inventory/src/Stocktake/api/api.ts
+++ b/packages/inventory/src/Stocktake/api/api.ts
@@ -84,7 +84,7 @@ export const getStocktakeQueries = (sdk: Sdk, storeId: string) => ({
           },
           filter: filterBy,
         });
-        return result.stocktakes;
+        return result?.stocktakes;
       },
     listAll:
       ({ sortBy }: { sortBy: SortBy<StocktakeRowFragment> }) =>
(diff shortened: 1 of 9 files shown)


clemens-msupply (approved):
still a bit annoying that we can't make TypeScript complain about it automatically...

PR #1202, opened 2022-04-29
Force integers for quantities

### LINKED ISSUE #1201: Outbound Shipment: Invalid value for argument
Creating a new outbound shipment, issuing with a small quantity instance 0.5, It complained saying, `Invalid value for argument`.

### Steps to reproduce:

- Setup a remote server, Web client and Central server
- Once setup is completed, Login
- Click to create a new `Outbound shipment`
- Select the item having SOH > 0
- On a issued batch line, enter quantity with 0.5
- Click OK
- See the message

Screen Shot 2022-04-29 at 11 00 35 AM

mark-prins:
Fixes #1201 
There's a few places where specifying a non-integer value will result in an error. This is an attempt to avoid the errors. Hopefully the rounding isn't too annoying for users 🤷 

## Testing

Create or edit each of these, and specify a real value. The value should round (`<0.5` => `0` and `0.5+` => `1`) when the input field loses focus
- Outbound shipments
- Response requisitions
- Inbound shipments
- Internal orders
- stocktakes

### DIFF
--- a/packages/common/src/ui/layout/tables/components/Cells/NumberInputCell/NonNegativeIntegerCell.tsx
+++ b/packages/common/src/ui/layout/tables/components/Cells/NumberInputCell/NonNegativeIntegerCell.tsx
@@ -5,7 +5,7 @@ import { RecordWithId } from '@common/types';
 import { useBufferState, useDebounceCallback } from '@common/hooks';
 
 // where NonNegative is n >=0
-export const NonNegativeNumberInputCell = <T extends RecordWithId>({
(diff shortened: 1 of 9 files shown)


clemens-msupply (approved):
Can you have an input field that doesn't allow to enter floats? e.g. catch the issues on key press level...

mark-prins:
yes - and we do! I had added one for the number input cell.. have just added a `NonNegativeIntegerInput` as well and used that in place of the separate `.round` and string parsing. thanks!

PR #1213, opened 2022-05-02
Add default stocktake description

### LINKED ISSUE #1208: Stock take can be saved without description
### Steps to reproduce:

- Setup, remote-server, web-client and central server
- Once setup finished, Login using web client
- Create a new stock take
- Fill in the necessary details
- Don't put any description
- Verify it 
- Stock take created without the description
   - Also wait so that it can make to central server
   - Once reached view that stock take and click OK
      - You will see a alert

It does not matter much to the open mSupply, I guess but this will create an issue on Central server if double click that Stock take click OK always will give an alert

mark-prins:
Fixes #1208 

Not very happy with the concept - went with it because
- it makes viewing stocktake in mSupply happier
- allows for localisation of the default description
- makes the list view look nicer

but.. 🤷 

Ideally we'd create the default on the server, but wasn't sure that could have translated / localised versions of the text if we did that


clemens-msupply (approved):
LGTM

PR #1215, opened 2022-05-03
Render expando only when expanded

### LINKED ISSUE #776: Expand components should mount when expanded
An example `Expando`.  component: https://github.com/openmsupply/openmsupply-client/blob/9231ae6ad1bd79154f826d5b9610a8fd219cce74/packages/inventory/src/Stocktake/DetailView/ContentArea.tsx#L15-L26

Which is used here:  https://github.com/openmsupply/openmsupply-client/blob/9231ae6ad1bd79154f826d5b9610a8fd219cce74/packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx#L96-L110

It would be better if the expando component was not rendered when not expanded.

I suspect just passing `isExpanded` to `Expando` which will be able to conditionally render might do the trick

mark-prins:
Fixes #776 

Well, that took longer than expected! 

The root of the issue was that we are rendering a `DataTable` to show the expando content.. and that wasn't instantiating a new table context. This meant that the expanded state was being reset - by playing with the rendering, the table row had to be rendered and with the reset table context, it wasn't expanded anymore! 🙄

### DIFF
--- a/packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
+++ b/packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
@@ -9,7 +9,7 @@ import {
   useIsFocused,
   useRowStyle,
 } from '../../context';
-import { Collapse, Fade } from '@mui/material';
+import { Fade } from '@mui/material';
 
 interface DataRowProps<T extends RecordWithId> {
   columns: Column<T>[];
@@ -17,7 +17,7 @@ interface DataRowProps<T extends RecordWithId> {
   onClick?: (rowData: T) => void;
   rowData: T;
   rowKey: string;
-  ExpandContent?: FC<{ rowData: T }>;
+  ExpandContent?: FC<{ rowData: T; isExpanded: boolean }>;
   dense?: boolean;
   rowIndex: number;
   keyboardActivated?: boolean;
@@ -108,21 +108,13 @@ export const DataRow = <T extends RecordWithId>({
           })}
         </TableRow>
       </Fade>
-      <tr>
-        <td colSpan={columns.length}>
-          <Collapse
-            sx={{
-              flex: 1,
-              '& .MuiCollapse-wrapperInner': {
-                display: 'flex',
-              },
-            }}
-            in={isExpanded}
-          >
-            {ExpandContent ? <ExpandContent rowData={rowData} /> : null}
-          </Collapse>
-        </td>
-      </tr>
+      {isExpanded && !!ExpandContent ? (
+        <tr>
+          <td colSpan={columns.length}>
+            <ExpandContent rowData={rowData} isExpanded={isExpanded} />
+          </td>
+        </tr>
+      ) : null}
     </>
   );
 };
--- a/packages/common/src/ui/layout/tables/components/Expand/MiniTable.tsx
+++ b/packages/common/src/ui/layout/tables/components/Expand/MiniTable.tsx
@@ -3,6 +3,7 @@ import { Column } from './../../columns/types';
 import { Box, alpha } from '@mui/material';
 import { RecordWithId } from '@common/types';
 import { DataTable } from '../../DataTable';
+import { createTableStore, TableProvider } from '../../context/TableContext';
 
 interface MiniTableProps<T extends RecordWithId> {
   rows: T[];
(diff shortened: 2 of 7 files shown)


packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
```diff
@@ -9,15 +9,15 @@ import {
   useIsFocused,
   useRowStyle,
 } from '../../context';
-import { Collapse, Fade } from '@mui/material';
+import { Fade } from '@mui/material';
```
mark-prins, line 12: 😢 

sad so lose the collapse transition.. but it isn't possible without rendering all expando sections which is what we are trying to avoid.

packages/common/src/ui/layout/tables/components/Expand/MiniTable.tsx
```diff
@@ -30,7 +31,9 @@ export const MiniTable = <T extends RecordWithId>({
           overflow: 'hidden',
         }}
       >
-        <DataTable dense columns={columns} data={rows} />
+        <TableProvider createStore={createTableStore}>
```
mark-prins, line 34: This was the problem. creating a new store here isolates the context to each table and fixes the issue.

packages/common/src/ui/layout/tables/context/TableContext.ts
```diff
@@ -47,7 +47,7 @@ const getRowState = (
   ...updates,
 });
 
-export const createTableStore = (): UseStore<TableStore> =>
+export const createTableStore = (): UseBoundStore<TableStore> =>
```
mark-prins, line 50: `useStore` is deprecated

packages/inventory/src/Stocktake/DetailView/SidePanel.tsx
```diff
@@ -24,7 +24,7 @@ const AdditionalInfoSection: FC = () => {
   const t = useTranslation('common');
 
   const { comment, user, update } = useStocktakeFields(['comment', 'user']);
-  const [bufferedComment, setBufferedComment] = useBufferState(comment);
+  const [bufferedComment, setBufferedComment] = useBufferState(comment ?? '');
```
mark-prins, line 27: was throwing a warning in console - changing from uncontrolled to controlled.

packages/inventory/src/Stocktake/DetailView/Toolbar.tsx
```diff
@@ -72,7 +72,7 @@ export const Toolbar: FC = () => {
         <Grid item>
           <DropdownMenu disabled={isDisabled} label={t('label.actions')}>
             <DropdownMenuItem IconComponent={DeleteIcon} onClick={onDelete}>
-              {t('button.delete-lines', { ns: 'distribution' })}
+              {t('button.delete-lines')}
```
mark-prins, line 75: throwing a warning in console - the `distribution` ns does not have this key

packages/common/src/ui/layout/tables/context/TableContext.ts
```diff
@@ -172,6 +172,7 @@ export const createTableStore = (): UseStore<TableStore> =>
 
         // Get currently focused row, if any
         const [currentFocusId, currentFocusObj] =
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
clemens-msupply, line 175: what is unused here?

packages/common/src/ui/layout/tables/components/DataRow/DataRow.tsx
clemens-msupply: on the plus side it feel snappier :)

clemens-msupply (approved):
looks good to me!

packages/common/src/ui/layout/tables/context/TableContext.ts
mark-prins: it doesn't like the first parameter in `rows.find(([_, { isFocused }])...` not being used. 
have realised belatedly that I can just use `rows.find(([, { isFocused }])...` so have removed the eslint hint. 🙄

clemens-msupply: strange since '_' usually makes the linter happy

PR #1217, opened 2022-05-04
Limit internal order quantity

### LINKED ISSUE #970: Ensure Int inputs are less than 4294967295
Got this error when updating a response line :

`"Failed to parse \"Int\": Only integers from 0 to 4294967295 are accepted.`

Seems big enough.. we should prevent this from occurring by having this as the max for all quantities 🤷

mark-prins:
Fixes #970 

Only found one instance left, after the recent updates to add the `NonNegativeIntegerInput` component.

### DIFF
--- a/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEditForm.tsx
+++ b/packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEditForm.tsx
@@ -88,6 +88,7 @@ export const RequestLineEditForm = ({
   const t = useTranslation('replenishment');
   const formatNumber = useFormatNumber();
   const { lines } = useRequestLines();
+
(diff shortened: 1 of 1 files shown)


packages/requisitions/src/RequestRequisition/DetailView/RequestLineEdit/RequestLineEditForm.tsx
```diff
@@ -147,7 +148,7 @@ export const RequestLineEditForm = ({
             Input={
               <NonNegativeIntegerInput
                 autoFocus
-                value={draftLine?.requestedQuantity}
+                value={draftLine?.requestedQuantity ?? 0}
```
mark-prins, line 151: was uncontrolled and the updates weren't applied to the value

clemens-msupply (approved):

PR #1218, opened 2022-05-05
Refactor userQueryParams for Outbounds

### LINKED ISSUE #1204: Cannot delete the Outbound shipment
### Step to reproduce

- Setup remote-server, Central server, Web-client
- Login to Web-client
- Create a dummy Outpatient invoice, add a line then delete the line save it
- View the Outpatient list
- Find that Outpatient shipment
- Select it and click `Delete selected lines`  from the option DDL
- It asked to `Select the line / lines`, but is already selected

Screen Shot 2022-04-29 at 1 45 54 PM

mark-prins:
Fixes #1204 

Am creating the PR at this stage.. there is more refactoring to do, but I want to reduce the number of files in the PR 😉 
This one is the essence of the change - the next steps are:
- replicate these changes across all usages of `useQueryParams`
- remove the hooks `useFilterBy` `useSortBy` & `usePagination`
- rework existing tests

The core of the problem is that the current hooks aren't sharing state.. so a list of items is retrieved by two separate calls, and each has different query params, though the code is expecting to see the same list of results.

I've lifted the queryParams state out, so it is created in the page context. Almost always, this is within the table context, so have created a query params context which is instantiated by the table context. For the cases where the queryParams are used outside of the table context (Name search is one) then the query params context can be instantiated separately.

Will also replace the `useSharedSort` implementation required by this PR: https://github.com/openmsupply/openmsupply-client/pull/1017 as I expect that won't be needed now.

### DIFF
--- a/packages/common/src/hooks/useFilterBy/useFilterBy.tsx
+++ b/packages/common/src/hooks/useFilterBy/useFilterBy.tsx
@@ -1,6 +1,6 @@
 import { useState } from 'react';
 
-interface FilterByConditionByType {
+export interface FilterByConditionByType {
   string: 'equalTo' | 'like';
   date: 'beforeOrEqualTo' | 'afterOrEqualTo' | 'equalTo';
 }
--- a/packages/common/src/hooks/useQueryParams/useQueryParams.ts
+++ b/packages/common/src/hooks/useQueryParams/useQueryParams.ts
@@ -1,7 +1,26 @@
-import { RecordWithId } from './../../types/index';
-import { usePagination, PaginationState } from '../usePagination';
-import { useSortBy, SortState, SortRule } from '../useSortBy';
-import { useFilterBy, FilterState, FilterBy } from '../useFilterBy';
+import create, { SetState, UseBoundStore } from 'zustand';
+import createContext from 'zustand/context';
+import { RecordWithId } from '@common/types';
+import {
+  usePagination,
+  PaginationState,
+  PaginationController,
+} from '../usePagination';
+import {
+  useSortBy,
+  SortState,
+  SortRule,
+  SortController,
+  SortBy,
+} from '../useSortBy';
+import {
+  useFilterBy,
+  FilterState,
+  FilterBy,
+  FilterController,
+  FilterByConditionByType,
+} from '../useFilterBy';
+import { Column } from '../../ui';
 
 export interface QueryParams<T extends RecordWithId>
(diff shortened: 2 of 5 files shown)


packages/common/src/hooks/useFilterBy/useFilterBy.tsx
```diff
@@ -1,6 +1,6 @@
 import { useState } from 'react';
 
-interface FilterByConditionByType {
+export interface FilterByConditionByType {
```
mark-prins, line 3: only the types will be used from the existing hooks, so will just lift those out

packages/common/src/hooks/useQueryParams/useQueryParams.ts
```diff
@@ -48,3 +67,140 @@ export const useQueryParams = <T extends RecordWithId>({
     queryParams,
   };
 };
+
+export interface QueryParamsStateNew<T extends RecordWithId> {
```
mark-prins, line 71: temporary, so I don't have to refactor everything in one big PR

packages/common/src/ui/layout/tables/context/TableContext.tsx
```diff
@@ -30,8 +37,30 @@ export interface TableStore {
...
+const { Provider, useStore: useTableStore } = createContext<TableStore>();
+
+const TableProvider = <T extends RecordWithId>({
+  children,
+  queryParamsStore,
+  ...props
+}: PropsWithChildren<{
+  initialStore?: UseBoundStore<TableStore>;
+  createStore: () => UseBoundStore<TableStore>;
+  queryParamsStore?: UseBoundStore<QueryParamsStateNew<T>>;
```
mark-prins, line 49: intending to make this required when all usages are updated

packages/invoices/src/OutboundShipment/ListView/ListView.tsx
```diff
@@ -85,7 +79,7 @@ export const OutboundShipmentListViewComponent: FC = () => {
 
       <DataTable
         pagination={{ ...pagination, total: data?.totalCount }}
-        onChangePage={onChangePage}
+        onChangePage={pagination.onChangePage}
```
mark-prins, line 82: changed the shape of the object a little too.. I haven't found a good reason to retain the nesting & duplication

clemens-msupply (approved):
looks good to me

PR #1219, opened 2022-05-06
1024 refactor query params 2

mark-prins:
Fixes #1024 

part two - updates the `ListView` components to use the query params store.

### DIFF
--- a/packages/common/src/intl/locales/en/distribution.json
+++ b/packages/common/src/intl/locales/en/distribution.json
@@ -58,7 +58,6 @@
   "label.requisition": "Requisition",
   "label.requested-quantity": "Requested quantity",
   "label.select-rows-to-allocate-them": "Select rows to allocate them",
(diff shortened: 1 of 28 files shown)


packages/common/src/intl/locales/en/distribution.json
```diff
@@ -98,7 +97,7 @@
   "messages.deleted-requisitions": "Deleted {{number}} requisitions",
   "messages.error-saving-shipment": "Error saving shipment 🥺",
   "messages.must-allocate-all-lines": "Cannot change the status until all lines have been allocated.",
-  "messages.select-rows-to-delete": "Select rows to delete",
+  "messages.select-rows-to-delete-them": "Select rows to delete them",
```
clemens-msupply, line 100: First I though its reads not very well, e.g. compared to "Select rows for deletion" or "Select rows to be deleted" but your version is actually very clear and unambiguous :-)

PR #1221, opened 2022-05-06
1204 refactor params 3

### LINKED ISSUE #1204: Cannot delete the Outbound shipment

mark-prins:
Fixes #1204 

yes, I know, that issue has been fixed twice already 🙄 

This PR removes the old hooks, and updates DetailViews.
One final PR coming.. I'd like to
- review the existing tests (update or remove)
- review the existing stories (update or remove)
- Try to type the context of the generic in `useQueryParams::createContext`


packages/invoices/src/InboundShipment/api/hooks.ts
```diff
@@ -125,17 +125,23 @@ export const useInboundLines = (itemId?: string) => {
...
+  const { sortBy, onChangeSortBy } = sort;
+  const selectItems = (invoice: InboundFragment) =>
+    inboundLinesToSummaryItems(
       invoice.lines.nodes.filter(line => isA.stockInLine(line))
-    ).sort(
-      SortUtils.getDataSorter(sortBy.key as keyof InboundItem, !!sortBy.isDesc)
-    );
-  }, []);
+    )
+      .map(item => ({
```
mark-prins, line 134: a neat little bug here: the hook was using the rows, mapping them to `SummaryItems` which do not have the properties which are sorted by. 

So the `getDataSorter` is unable to sort the rows, which means that the `OK & Next` button was using the rows in the wrong order. I've popped in a prop, of the sort key, which allows correct sorting.

packages/requisitions/src/RequestRequisition/DetailView/columns.ts
```diff
@@ -1,46 +1,21 @@
...
-  }, [sortBy]);
-  return { sortBy, onChangeSortBy };
-};
-
 export const useRequestColumns = () => {
   const t = useTranslation('common');
   const { maxMonthsOfStock } = useRequestFields('maxMonthsOfStock');
-  const { sortBy, onChangeSortBy } = useSharedSortBy();
+  const { sort } = useQueryParamsStore();
+  const { sortBy, onChangeSortBy } = sort;
```
mark-prins, line 18: the whole `sharedSortBy` is no longer needed, as the QueryParamsStore handles it

packages/system/src/Report/api/hooks.ts
```diff
@@ -29,20 +28,15 @@ const useReportApi = () => {
...
-  const initialFilterBy = category
-    ? { category: { equalTo: category } }
-    : undefined;
-  const initialListParameters = {
-    initialSortBy: { key: 'name' },
-    initialFilterBy,
+  const filterBy = category ? { category: { equalTo: category } } : null;
+  const queryParams = {
+    filterBy,
+    sortBy: { key: 'name', isDesc: false, direction: 'asc' as 'asc' | 'desc' },
```
mark-prins, line 34: thought to just hard code these, as the params are not changed by the UI / user. we are just returning a lookup list. If we add in controls, we can support params again.

clemens-msupply (approved):
Looks good to me

PR #1233, opened 2022-05-10
1115 Warning modals for Delete actions

### LINKED ISSUE #1115: Warning modals for all "Delete" actions
For safety, we probably want to have a warning to the user whenever deleting something.

When doing this, we can also change the "Action" drop-down to clickable links when there is only one action available, as per [this discussion](https://github.com/openmsupply/openmsupply-client/pull/1112#discussion_r834864810) (and how to implement it).

(Styling for action when in "button" form: https://github.com/openmsupply/openmsupply-client/pull/1112#discussion_r834903976)

CarlosNZ:
Fix #1115 

~~Just leaving this as a Draft PR initially -- there are quite a few of these delete actions around, so just want to make sure we're happy with this pattern before copying it to all the rest.~~

Implemented now using a common `useDeleteConfirmation` hook for all `useDelete...` hooks.


packages/common/src/ui/components/menus/DropdownMenu/DropdownMenu.tsx
```diff
@@ -94,9 +95,11 @@ export const DropdownMenu: FC<PropsWithChildren<DropdownMenuProps>> = ({
         disabled={disabled}
         value=""
         size="small"
+        open={open}
         labelId={`action-drop-down-label-${label}`}
         variant="outlined"
         IconComponent={ChevronDownIcon}
+        onClick={() => setOpen(curr => !curr)}
```
CarlosNZ, line 102: The drop-down was staying open after clicking it. This is the only way I could find to change that -- perhaps there is a better way?

packages/invoices/src/OutboundShipment/api/hooks/document/useOutboundDelete.ts
```diff
@@ -26,6 +27,23 @@ export const useOutboundDelete = () => {
...
+    const successSnack = success(deletedMessage);
+    successSnack();
+  };
+
+  const confirmDelete = useConfirmationModal({
+    onConfirm: deleteData,
+    message: t('messages.delete-shipment-warning'),
+    title: t('heading.are-you-sure'),
+  });
+
```
CarlosNZ, line 46: Initially tried putting the confirmation call in the Toolbar component, which seems more appropriate. However then it would show the confirmation regardless of the additional delete checks (canDelete, none selected), which was not good.

Putting in here means splitting "deleteAction" into two functions, one pre-confirmation, and called on confirmation. That okay?

packages/common/src/intl/locales/en/replenishment.json
```diff
@@ -81,5 +81,6 @@
   "messages.deleted-requisitions_one": "Deleted {{count}} order",
   "messages.deleted-requisitions_other": "Deleted {{count}} orders",
   "messages.saved": "Saved",
-  "messages.select-rows-to-delete-them": "Select rows to delete them"
+  "messages.select-rows-to-delete-them": "Select rows to delete them",
+  "messages.delete-shipment-warning": "This will permanently delete shipment(s)"
```
mark-prins, line 85: generally, I prefer using the plural function e.g. `label.number-months_one` and `label.number-months_other` in this file. The grammar of the singular version stands out for me.. 'This will permanently delete shipment' doesn't sound good.

Given how common this action / phrase is, maybe this should be in `common` and just be `This will delete permanently` or something suitably generic.

mark-prins: Also - it is being used by Outbounds, so should be in `distribution`?

packages/common/src/ui/components/menus/DropdownMenu/DropdownMenu.tsx
mark-prins: will have to investigate. I'd hope that the component would provide something better

packages/invoices/src/OutboundShipment/ListView/Toolbar.tsx
```diff
@@ -13,7 +13,7 @@ import { useOutbound, OutboundRowFragment } from '../api';
 export const Toolbar: FC<{
   filter: FilterController;
 }> = ({ filter }) => {
-  const t = useTranslation('distribution');
+  const t = useTranslation('common');
```
mark-prins, line 16: the `button.delete-lines` used in this component is in `distribution` and the `placeholder.search-by-name` which is in `common` is handled since common is a fallback

packages/invoices/src/OutboundShipment/api/hooks/document/useOutboundDelete.ts
mark-prins: yeah, I agree. Would be nice in the Toolbar. return a promise? handle resolve and reject in Toolbar?
pass in methods for `onSuccess` etc?

mark-prins (commented):
will need to think more carefully about it, just putting in the comment so we have something to chat about!

packages/common/src/ui/components/menus/DropdownMenu/DropdownMenu.tsx
CarlosNZ: Hmm, there's something about the `<DropdownMenu/>` component that is making the `<Select>`component behave in a non-standard way. When I replace the menu with regular Mui `<MenuItem>` components it works as expected. Will investigate further.

mark-prins: Objection withdrawn.. I've played around and realised that we're using a `Select` component and not setting the value, so not using it as a Select. Should probably be using a [Menu](https://mui.com/material-ui/react-menu/) which requires managing the open state manually anyway. So.. back to where we started with what you've already done!

packages/invoices/src/InboundShipment/api/hooks/document/useInboundDelete.ts
```diff
@@ -0,0 +1,47 @@
+import {
```
CarlosNZ, line 1: Do you know how to get this to show up in the diff as a renamed (moved) file rather than a deleted file and a whole new file?

packages/invoices/src/InboundShipment/api/hooks/line/useDeleteSelectedLines.ts
```diff
@@ -1,48 +1,50 @@
...
-          .flat()
-      : lines?.filter(({ id }) => state.rowState[id]?.isSelected);
-  });
+      return isGrouped
+        ? items
+            ?.filter(({ id }) => state.rowState[id]?.isSelected)
+            .map(({ lines }) => lines.flat())
+            .flat()
+        : lines?.filter(({ id }) => state.rowState[id]?.isSelected);
+    }) || [];
```
CarlosNZ, line 26: Thought it was better to have empty array fallback here rather than allowing it to be possibly `undefined` in delete confirmation.

packages/system/src/Location/ListView/Toolbar.tsx
```diff
@@ -63,6 +64,12 @@ export const Toolbar: FC<{
     }
   };
 
+  const showDeleteConfirmation = useConfirmationModal({
+    onConfirm: deleteAction,
+    message: t('messages.confirm-delete-locations'),
+    title: t('heading.are-you-sure'),
+  });
+
```
CarlosNZ, line 72: Had to break the pattern for this one -- the deleteAction determines if it can delete by just trying the mutation and catching the error, so can't verify if `canDelete` before running the actual delete. 🤷‍♂️

mark-prins (approved):
👍  That's fantastic thanks! much nicer having a confirmation modal for the deletes and the hook makes it super easy.

I was going to comment about updating the translations to use plurals.. and that was getting a bit wordy, so I've just committed the changes in 37e38a23 Got a little carried away.. but you can revert if you like 😁  I just thought that the messaging was nicer with the count in it..

packages/common/src/intl/locales/en/distribution.json
```diff
@@ -91,10 +91,14 @@
...
-  "messages.deleted-invoices_one": "Deleted {{count}} shipment",
-  "messages.deleted-invoices_other": "Deleted {{count}} shipments",
-  "messages.deleted-lines": "Deleted {{number}} lines",
-  "messages.deleted-requisitions": "Deleted {{number}} requisitions",
+  "messages.deleted-shipments_one": "Deleted {{count}} shipment",
+  "messages.deleted-shipments_other": "Deleted {{count}} shipments",
+  "messages.confirm-delete-shipments": "This will permanently remove shipments",
+  "messages.confirm-delete-lines": "This will permanently remove lines from invoice",
+  "messages.deleted-lines_one": "Deleted {{number}} line",
+  "messages.deleted-lines_other": "Deleted {{number}} lines",
```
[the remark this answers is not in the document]
CarlosNZ: Oh, so it has to specifically be called `count`, it can't just match whatever variable name was passed in! 😲

Thanks for fixing :)

packages/inventory/src/Stocktake/DetailView/Toolbar.tsx
```diff
@@ -21,7 +21,7 @@ export const Toolbar: FC = () => {
     'description',
     'stocktakeDate',
   ]);
-  const { onDelete } = useStocktake.line.deleteSelected();
+  const onDelete = useStocktake.line.deleteSelected();
```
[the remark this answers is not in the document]
CarlosNZ: No, it was actually inconsistent already -- the ones that called the delete method `onDelete` tended to return an object, the ones that called their delete method `deleteAction` were returning the method directly. 

For example, in OutboundShipment / ListView / Toolbar.tsx we already had:
```
  const onDelete = useOutbound.document.delete();
```
which was renaming the directly exported "deleteAction" from `useOutboundDelete.ts`. So they were all getting renamed to `onDelete` where they were used in the `onClick={onDelete}` but it was about 50/50 were importing a direct "deleteAction" method or an object with an "onDelete" method. 

So I just made them all consistent in terms of hook usage in the Toolbars, maybe should have gone the other way?

(Personally, I like returning something directly from a hook when there's only one thing returned. If I see `const {somevalue} = useSomeHook()`, I assume there are multiple properties returned and we're only using one of them, whereas with `const somevalue = useSomeHook() I can tell immediately that that hook only returns one thing)

packages/requisitions/src/RequestRequisition/api/hooks/line/useDeleteRequestLines.ts
```diff
@@ -28,19 +27,25 @@ export const useDeleteRequestLines = () => {
...
-    if (isDisabled) {
-      info(t('label.cant-delete-disabled-requisition'))();
-      return;
-    }
-    const number = selectedRows?.length;
-    if (selectedRows && number) {
-      mutate(selectedRows, {
-        onSuccess: success(t('messages.deleted-lines', { number: number })),
+    mutateAsync(selectedRows)
+      // .then(() => queryClient.invalidateQueries(api.keys.base()))
```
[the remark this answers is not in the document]
CarlosNZ: Ah yes, all gone now 65a28bcb

packages/system/src/Location/ListView/Toolbar.tsx
```diff
@@ -21,7 +22,7 @@ export const Toolbar: FC<{
   data: LocationRowFragment[];
   filter: FilterController;
 }> = ({ data }) => {
-  const t = useTranslation('inventory');
+  const t = useTranslation(['inventory', 'common']);
```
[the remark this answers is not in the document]
CarlosNZ: Ah, okay 👍🏼  e7a8c276

packages/inventory/src/Stocktake/DetailView/Toolbar.tsx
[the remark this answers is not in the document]
mark-prins: > (Personally, I like returning something directly from a hook when there's only one thing returned. If I see const {somevalue} = useSomeHook(), I assume there are multiple properties returned and we're only using one of them, whereas with const somevalue = useSomeHook() I can tell immediately that that hook only returns one thing)

Ditto

PR #1234, opened 2022-05-10
Custom language detection

### LINKED ISSUE #1232: Problem with language detection in Firefox
Seems to be affecting Firefox browser only currently (confirmed okay in Chrome and Safari).

After removing all local storage and cookies, Firefox gives the following error when reloading:
Screen Shot 2022-05-10 at 4 06 26 PM

Probably the locale string retrieved from the browser is fishy -- need to investigate, and maybe provide a fallback if there's a problem?

mark-prins:
Fixes #1232 

Have taken the i18next navigator detection and tweaked to remove the region. Fixes the problem for me in firefox


CarlosNZ (approved):
Works for me! Thanks

PR #1236, opened 2022-05-11
Remove add item when disabled

### LINKED ISSUE #1235: Small bug: Add item link still active in blank table when it shouldn't be
Screen Shot 2022-05-11 at 4 21 13 PM
Notice the "Add item" button at the top is greyed out, because the Shipment is completed, yet the "Add item" link in the middle of the page is still active.

mark-prins:
Fixes #1235 

Remove the link on the detail view when the record is disabled for the following:
- Inbound shipment
- Outbound shipment
- Internal order

Added bonus: noticed that the stocktake detail wasn't using the <NothingHere/> component, so have popped that in too


CarlosNZ (approved):
LGTM!

PR #1239, opened 2022-05-12
Auth error state set & reset

### LINKED ISSUE #1237: Auth error update on login
The localStorage setting of `/auth/error` is being set to "Unauthenticated" when the cookie expires.
When logging in, this is unset by the line `setError(store ? undefined : AuthError.NoStoreAssigned);` 

The logic could be improved - if the login was unsuccessful then the error should be `AuthError.Unauthenticated` rather than `AuthError.NoStoreAssigned`

mark-prins:
Fixes #1237 

This is overboard, possibly? I've wandered off a little here, just to show this:

Screen Shot 2022-05-12 at 4 28 48 PM

which was not being displayed any longer, since the permission denied error was being set after this error was ( the dashboard queries were triggering them), and therefore the permission denied modal was shown instead of this one. To get things working, I've had to add a request check function into GqlContext which allows the skipping of requests.

The _actual_ issue was not very complex. It came from the health supply hub, which is using a copy of `common` and having auth issues, because the line which resets auth state on login was removed. To make things clearer, I've lifted that out to a helper method ( thanks @jmbrunskill for the heads up on that one! )

### DIFF
--- a/packages/common/src/authentication/AuthContext.tsx
+++ b/packages/common/src/authentication/AuthContext.tsx
@@ -112,7 +112,7 @@ export const AuthProvider: FC<PropsWithChildrenOnly> = ({ children }) => {
   const { mutateAsync, isLoading: isLoggingIn } = useGetAuthToken();
   const authCookie = getAuthCookie();
   const [cookie, setCookie] = useState<AuthCookie | undefined>(authCookie);
-  const [error, setError] = useLocalStorage('/auth/error');
+  const [error, setError, removeError] = useLocalStorage('/auth/error');
   const storeId = cookie?.store?.id ?? '';
   const { mutateAsync: getStores } = useUserDetails();
   const { setHeader, setSkipRequest } = useGql();
@@ -151,7 +151,7 @@ export const AuthProvider: FC<PropsWithChildrenOnly> = ({ children }) => {
   const setLoginError = (isLoggedIn: boolean, hasValidStore: boolean) => {
     switch (true) {
       case isLoggedIn && hasValidStore: {
-        setError(undefined);
+        removeError();
         break;
       }
       case !isLoggedIn: {
--- a/packages/common/src/authentication/AuthContext.tsx
+++ b/packages/common/src/authentication/AuthContext.tsx
@@ -156,7 +156,6 @@ export const AuthProvider: FC<PropsWithChildrenOnly> = ({ children }) => {
       }
       case !hasValidStore: {
         setError(AuthError.NoStoreAssigned);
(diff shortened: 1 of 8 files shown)


packages/common/src/authentication/AuthContext.tsx
```diff
@@ -147,7 +148,41 @@ export const AuthProvider: FC<PropsWithChildrenOnly> = ({ children }) => {
     return !!stores && stores?.length > 0 ? stores?.[0] : undefined;
   };
 
+  const setLoginError = (isLoggedIn: boolean, hasValidStore: boolean) => {
+    switch (true) {
+      case isLoggedIn && hasValidStore: {
```
mark-prins, line 153: a longhand version of what was in place - to make it all clearer.

packages/common/src/authentication/AuthContext.tsx
```diff
@@ -147,7 +148,41 @@ export const AuthProvider: FC<PropsWithChildrenOnly> = ({ children }) => {
...
+        break;
+      }
+      default: {
+        setError(AuthError.Unauthenticated);
+      }
+    }
+  };
+
+  const authNameQueries = ['authToken', 'me'];
+  const isAuthRequest = (definitionNode: DefinitionNode) => {
```
mark-prins, line 168: this bit is app specific and shouldn't be in common.. but that's where the auth context currently is. I've lifted out of GqlContext.. ideally this would all be in some app-specific common package. that will come soon..

CarlosNZ (commented):
From your original issue:

> The logic could be improved - if the login was unsuccessful then the error should be AuthError.Unauthenticated rather than AuthError.NoStoreAssigned

However, when attempting to log in with incorrect password, the "NoStoreAssigned" error is still being set in localStorage rather than "Unauthenticated"

Screen Shot 2022-05-13 at 12 24 08 PM

mark-prins:
ah yes. of course. have updated in a47ac0ca
thanks

CarlosNZ (approved):
Looking good, thanks.

There's one little thing that bothers me -- and only because it's a subtle thing that caused a problem in Conforma that took a long time to figure out:

When the error field is `undefined`, it's actually storing the *string* "undefined" rather than an empty value, presumably because a true `undefined` is not a valid value for localStorage.

Screen Shot 2022-05-16 at 3 51 39 PM

Any chance we could set this to `null` perhaps, or delete the key altogether when it's not being used? Not a biggie, but just to avoid in future someone wondering why `if localStorage.get(key) === undefined` isn't working and tearing their hair out for several days.

mark-prins:
thanks Carl - 55db443f

PR #1245, opened 2022-05-16
Language detection order

### LINKED ISSUE #1241: Changing language in Admin view doesn't work
☝️
Needed for Madagascar demo on Thursday 🙏

mark-prins:
Fixes #1241 
Oops. Update the language detection order, so that localStorage is before the new custom navigator detection 🙄

### DIFF
--- a/packages/common/src/intl/context/IntlContext.tsx
+++ b/packages/common/src/intl/context/IntlContext.tsx
@@ -47,12 +47,11 @@ export const IntlProvider: FC<PropsWithChildrenOnly> = ({ children }) => {
         defaultNS,
         detection: {
           order: [
-            'omsBrowserLanguageDetector',
             'querystring',
             'cookie',
             'localStorage',
             'sessionStorage',
-            'navigator',
+            'omsBrowserLanguageDetector',
             'htmlTag',
           ],
         },


clemens-msupply:
Maybe its just my tablet but now the normal login name gets converted to lower case, e.g. Admin -> admin...

CarlosNZ (approved):
Ah -- that looks more relevant to the issue! 😆

packages/common/src/intl/context/IntlContext.tsx
```diff
@@ -47,12 +47,11 @@ export const IntlProvider: FC<PropsWithChildrenOnly> = ({ children }) => {
         defaultNS,
         detection: {
           order: [
-            'omsBrowserLanguageDetector',
             'querystring',
             'cookie',
             'localStorage',
             'sessionStorage',
-            'navigator',
```
clemens-msupply, line 55: Is removing the navigator correct?

mark-prins: the custom detector is a theft of the navigator code, so I thought that we didn't need to duplicate 😉

mark-prins:
> Maybe its just my tablet but now the normal login name gets converted to lower case, e.g. Admin -> admin...

Argh! sorry 🤷  can't see why that would happen. I've only added the autocapitalize hint to the intialise screen. The login is a separate component. 
Maybe it's time to do this issue? https://github.com/openmsupply/remote-server/issues/745

Issue #1244, opened 2022-05-16
Outbound Shipment: Add item text on body does not work


bijaySussol:
### Steps to reproduce

- Setup remote server, frontend with the central server
-  Make sure the connection is OK
- Login from frontend
- Click on `Outbound Shipment` from the sidebar
- Click `New Shipment` choose the Customer take to the input details of Outbound
- On the body section of page there is a text saying `Add item`
- click on that
- You will see its restricted

mark-prins:
oh, oops. I created a new issue: #1249 sorry. Didn't read this in time! you were quicker than I was at spotting it. 
Should be fixed with #1250

PR #1250, opened 2022-05-17
Outbound: add item

### LINKED ISSUE #1249: Outbound: add item - no stock available
Adding an outbound line, in a store which has stock, and I see `There is no stock available`.
Why?

mark-prins:
Fixes #1249, #1244

The problem is partly a user error. Clicking the add item in the NothingHere component gave the error, using the app bar button is fine. Just had to update the usage of `addItem` for outbounds


CarlosNZ (approved):
LGTM!

PR #1253, opened 2022-05-18
1252 use Url Query parameters

CarlosNZ:
Initial attempt at openmsupply/open-msupply#2.

Draft PR while we work out if this is a good idea. Currently only implemented for Outbound Shipments.

Created 2 hooks:
- `useUrlQuery` a general wrapper for ReactRouter `useSearchParams` to auto-convert between types and partial updating of query. 
- `useHandleQueryParams` -- a "go-between" between `useUrlQuery` and `useQueryParams` (zustand store). `useQueryParams` was passing the change methods (`onChangeSortBy` etc.) down to the DataTable which were activated onClicks, etc. Instead I've passed down URLQuery update methods from new hook, and this hook responds to changes in URLQuery by calling the `useQueryParams` methods in turn.

Was intending to put the "handle query params" logic inside `useQueryParams`, but the store create function wouldn't allow calling additional hooks (such as `useSearchParams`).

I like the functionality, but there could be additional problems that make this a no-go -- hence just implementing on one table for starters.

See additional comments inline.


packages/common/src/ui/layout/tables/DataTable.tsx
```diff
@@ -31,6 +31,7 @@ export const DataTableComponent = <T extends RecordWithId>({
   noDataMessage,
   pagination,
   onChangePage,
+  onChangeSortBy,
```
CarlosNZ, line 34: Passed the alternative `onChangeSortBy` down through these props, as passing it through "columns" had problems with stale closure (was capturing the state of searchParams when first instantiated and never updating)

packages/common/src/ui/layout/tables/components/Header/Header.tsx
```diff
@@ -21,11 +21,13 @@ export const HeaderRow: FC<PropsWithChildren<{ dense?: boolean }>> = ({
 interface HeaderCellProps<T extends RecordWithId> {
   column: Column<T>;
   dense?: boolean;
+  onChangeSortBy?: (column: Column<T>) => void;
 }
 
 export const HeaderCell = <T extends RecordWithId>({
   column,
   dense = false,
+  onChangeSortBy: onChangeSortAlt,
```
CarlosNZ, line 30: Bad naming. Suggestions?

packages/invoices/src/OutboundShipment/ListView/ListView.tsx
```diff
@@ -35,9 +36,20 @@ export const OutboundShipmentListViewComponent: FC = () => {
...
+    updateFilterQuery,
+    onChangeUrlQuery,
+  } = useHandleQueryParams({ sort, pagination, filterKey: 'otherPartyName' });
+  const { sortBy } = sort;
   useDisableOutboundRows(data?.nodes);
 
+  useEffect(() => {
+    onChangeUrlQuery(columns, filter);
+  }, [urlQuery]);
+
```
CarlosNZ, line 52: I don't like having this `useEffect` in here. Would prefer it was all handled inside `useHandleQueryParams`. Unfortunately`onChangeUrlQuery` needs access to the *current* columns state, so has to be passed in when run, can't use the value of columns when the hook is first instantiated.

Any suggestions?

josh-griffin:
😁 👋 

I dropped by and this looked fun. What you've done is really clever!

I've made some suggestions here https://github.com/openmsupply/openmsupply-client/pull/1254 really I wanted to validate what I was thinking worked.. seemed to, so was easier to just show it

Basically the url state is global and I don't think we would need to sync it with the useQueryParams store - we could just kill it. However that doesn't work for any secondary tables - so if that's needed, two hooks? One backed by the URL state and another by a store?

CarlosNZ:
> I've made some suggestions here #1254 really I wanted to validate what I was thinking worked.. seemed to, so was easier to just show it

Well hello, stranger! Thanks for that -- I'll check out your suggestions soon 👍

packages/common/src/hooks/useUrlQuery/useHandleQueryParams.ts
```diff
@@ -0,0 +1,74 @@
...
+  const updateFilterQuery = (key: string, value: string) => {
+    updateQuery({ [key]: value });
+  };
+
+  const onChangeUrlQuery = (
+    columns: Column<any>[],
+    filter: FilterController
+  ) => {
+    // Update Sort
+    if (urlQuery?.['sort'] || urlQuery?.['dir']) {
```
CarlosNZ, line 43: I do not like this syntax, but Typescript is requiring it. Do you have any issue with setting `"noPropertyAccessFromIndexSignature": false` in tsconfig.json?
Then it can just be standard dot notation: `if (urlQuery?.sort || urlQuery?.dir)...`

mark-prins: I would rather not. The complaint that ts is raising is that the `sort` prop might not exist and given that the shape is

```
interface UrlQueryObject {
  [key: string]: string | number | boolean;
}
```

I agree with ts on this one. I can't see how the `urlQuery` can be undefined though, so you don't need the `?` in which case `urlQuery['sort']` looks fine to me.

An alternative would be to define the expected, but optional, props for the query

```
interface UrlQueryObject
  extends Record<string, string | number | boolean | undefined> {
  sort?: string;
  dir?: 'asc' | 'desc';
  page?: number;
  ...
}
```

which allows `    if (urlQuery.sort || urlQuery.dir) {` while forcing unexpected params into the index format.. `urlQuery['tomato']`

packages/invoices/src/OutboundShipment/ListView/ListView.tsx
```diff
@@ -68,18 +80,19 @@ export const OutboundShipmentListViewComponent: FC = () => {
...
     ],
-    { onChangeSortBy, sortBy },
+    { onChangeSortBy: () => {}, sortBy },
     [sortBy]
   );
 
   return (
     <>
-      <Toolbar filter={filter} />
+      <Toolbar filter={filter} updateFilter={updateFilterQuery} />
```
mark-prins, line 89: the filter is being retained - if I filter, then clear the filter then sort the column, the filter reappears in the URL and the list is filtered

packages/invoices/src/OutboundShipment/ListView/Toolbar.tsx
```diff
@@ -12,7 +12,8 @@ import { useOutbound, OutboundRowFragment } from '../api';
 
 export const Toolbar: FC<{
   filter: FilterController;
-}> = ({ filter }) => {
+  updateFilter: (key: string, value: string) => void;
```
mark-prins, line 15: it should be possible to use the method on the `filter` object - I'd look at removing the params from the list hook; they were only put in there to fix the issue of having multiple hooks call different params instances. Now that we have a global state, that can be undone and the filter can be the `useUrlQuery` version only, which is then called by `Toolbar` and `columns`

mark-prins (commented):
😁  awesome - and nice work with the rendering, it's all looking pretty sweet.

I'm unsure if I should review in too much depth, or whether you're integrating Josh's ideas?
There's a few things which aren't behaving and in general, I agree with the idea of putting the global state in the url only, and just having a local state for the lower level table instances. we can simplify the process nicely that way.

When navigating back to the list view, I most often use the breadcrumb which shows me the list using the default options. I think that's fine - it seems logical to me that the navigation links would give you a default view, whereas the back navigation returns you back to the state you had.

CarlosNZ:
> I'm unsure if I should review in too much depth, or whether you're integrating Josh's ideas?

Yeah, I will do that. So hold off for the time being.

> I agree with the idea of putting the global state in the url only, and just having a local state for the lower level table instances. we can simplify the process nicely that way.

I think we're all on the same page about what the best approach is, so I'll make that happen when I get a chance.

CarlosNZ:
Closing for now -- will integrate suggestions and make a new PR on the [new repo](https://github.com/openmsupply/open-msupply/issues/2)

PR #1254, opened 2022-05-18
Add suggestions

josh-griffin:
😁


josh-griffin:
Sweet 😁  I'll close this mess I made aye. Might find that a global store for any additional tables on a page are not required as it might just be a smaller & simpler component

mark-prins:
hm, probably right - that the additional tables don't need global state. so we just need an option whether to save the state in the url or not

